PR #6, opened 2026-03-21
Sync Bluesky replies, likes, and reposts as WordPress comments

### LINKED ISSUE #5: Sync Bluesky replies back as WordPress comments
## Summary

Pull replies to published Bluesky posts back into WordPress as comments, enabling two-way interaction between the WordPress site and the AT Protocol network.

## Background

ATmosphere currently publishes posts to Bluesky (`app.bsky.feed.post`) and standard.site (`site.standard.document`). Replies on Bluesky are not synced back. standard.site does not define a reply/comment mechanism, so this is Bluesky-only for now.

### Primary: Poll `listNotifications`

- Register a WP-Cron event (e.g. every 15 minutes) calling `app.bsky.notification.listNotifications` with `reasons=["reply"]`.
- Track a cursor in `wp_options` to only process new notifications.
- For each reply:
  - Match `record.reply.root.uri` or `record.reply.parent.uri` to a local post via `_atmosphere_bsky_uri` post meta.
  - Resolve the author via `app.bsky.actor.getProfile` (display name, handle, avatar).
  - Insert as a WordPress comment using `wp_new_comment()`.
  - Store `comment_meta`: `protocol`, `source_uri`, `source_did` for deduplication and attribution.
- Handle nested replies by setting `comment_parent` when the parent URI maps to an existing comment.

### Secondary: `getPostThread` backfill

- On manual trigger or less-frequent cron, call `app.bsky.feed.getPostThread` for each published post's AT-URI.
- Walk the `replies[]` tree and insert any missing comments.
- Useful for catching replies from before the plugin was installed.

## Why not Jetstream/firehose?

Requires a persistent WebSocket connection, which is impractical in a standard WordPress/PHP environment.

## Reference

- `app.bsky.notification.listNotifications` — https://docs.bsky.app/docs/api/app-bsky-notification-list-notifications
- `app.bsky.feed.getPostThread` — https://docs.bsky.app/docs/api/app-bsky-feed-get-post-thread
- The WordPress ActivityPub plugin handles incoming replies similarly via `Interactions::add_comment()`

pfefferle:
Fixes #5.

Supersedes the reply-only design that was in the earlier version of this PR. The single poller now handles all three notification types in one pass and stores them in a shape that is compatible with `wordpress-activitypub`, so theme code, admin filters, and future moderation tooling can query reactions from both protocols uniformly.

## What's in here

- `Reaction_Sync` class polls `app.bsky.notification.listNotifications` for `reply`, `like`, and `repost` reasons, paginates via cursor (`atmosphere_reactions_cursor` option), dispatches by reason through a shared `insert_reaction` helper.
- Replies become `comment_type = 'comment'` with nested-thread walking (direct → parent comment → root fallback).
- Likes become `comment_type = 'like'`, reposts become `comment_type = 'repost'`, both with empty `comment_content` and top-level parent — same shape `wordpress-activitypub` uses.
- Every reaction comment gets the same meta keys the AP plugin writes:
  - `protocol = 'atproto'`
  - `source_id` = the reaction's AT-URI (dedup key)
  - `source_url` = the human-visitable `https://bsky.app/…/post/…` for replies, empty for likes/reposts (no web page exists for `app.bsky.feed.like` / `repost` records)
- Atproto-specific extras preserved as their own keys: `_atmosphere_bsky_cid`, `_atmosphere_author_did`.
- Author profiles cached via transients (no CPT in this PR — deferred).
- Hourly WP-Cron (`atmosphere_sync_reactions`), registered on activation when connected, cleared on deactivation. The legacy `atmosphere_sync_comments` hook name is also cleared defensively.
- 14 PHPUnit tests covering reply/like/repost success paths, dedup on `source_id`, unknown-target skip, and nested reply threading.

## Testing instructions

1. Publish a post to Bluesky via ATmosphere.
2. On Bluesky, have someone reply to, like, and repost the post.
3. Trigger the cron manually: `wp cron event run atmosphere_sync_reactions` (or wait for the hourly schedule).
4. Verify:
   - The reply appears as a WordPress comment with `comment_type = 'comment'` and a populated `source_id` / `source_url` meta.
   - The like and repost appear as comments with `comment_type = 'like'` / `'repost'`, empty body, and populated `source_id` meta.
   - Running the cron again does not duplicate any of them.
5. Nested case: have someone reply to the reply on Bluesky; after the next cron run, the new comment should be attached as a child of the previous one (same `comment_post_ID`, non-zero `comment_parent`).

## Known follow-up (not blocking this PR)

`insert_reaction` calls `get_gmt_from_date( $record['createdAt'] )` to derive `comment_date_gmt`. Bluesky's `createdAt` is already UTC ISO-8601, but `get_gmt_from_date` applies a local-to-UTC conversion, so on sites whose timezone is not UTC the stored GMT column is off by the site's offset. This bug existed before this PR (inherited from the original reply-only implementation) and is tolerable — the dedup key and all behaviour driven by comment date are unaffected. Will be fixed in a follow-up PR alongside switching to `gmdate( 'Y-m-d H:i:s', strtotime( … ) )`.

### Changelog entry

<details>

#### Significance

- [ ] Patch
- [x] Minor
- [ ] Major

#### Type

- [x] Added - for new features

#### Message

Sync Bluesky replies, likes, and reposts back as WordPress comments.

</details>

### DIFF
--- a/includes/class-reaction-sync.php
+++ b/includes/class-reaction-sync.php
@@ -288,6 +288,16 @@ class Reaction_Sync {
 	}
 
 	/**
+	 * Process a repost notification into a WordPress repost comment.
+	 *
+	 * @param array $notification Notification from listNotifications.
+	 * @return int|false Comment ID or false.
+	 */
+	private static function process_repost( array $notification ): int|false {
+		return self::process_subject_reaction( $notification, 'repost' );
+	}
+
+	/**
 	 * Process a reply notification into a WordPress comment.
 	 *
 	 * @param array $notification Notification from listNotifications.
--- a/includes/class-reaction-sync.php
+++ b/includes/class-reaction-sync.php
@@ -99,11 +99,7 @@ class Reaction_Sync {
 			$notifications = $response['notifications'] ?? array();
 
 			foreach ( $notifications as $notification ) {
-				if ( 'reply' !== ( $notification['reason'] ?? '' ) ) {
-					continue;
-				}
-
-				self::process_reply( $notification );
+				self::process_notification( $notification );
 			}
 
 			$cursor = $response['cursor'] ?? null;
@@ -117,14 +113,38 @@ class Reaction_Sync {
 	}
 
 	/**
-	 * Fetch a page of reply notifications from the PDS.
+	 * Dispatch a single notification to its reason-specific handler.
+	 *
+	 * Unknown reasons are silently skipped.
+	 *
+	 * @param array $notification Notification from listNotifications.
+	 * @return int|false Comment ID if inserted, false if skipped.
+	 */
+	private static function process_notification( array $notification ): int|false {
+		switch ( $notification['reason'] ?? '' ) {
+			case 'reply':
+				return self::process_reply( $notification );
+			case 'like':
+				return self::process_like( $notification );
+			case 'repost':
+				return self::process_repost( $notification );
+			default:
+				return false;
+		}
+	}
+
+	/**
+	 * Fetch a page of reaction notifications from the PDS.
+	 *
+	 * Asks for reply, like, and repost reasons. Dispatch happens in
+	 * process_notification based on the individual notification's reason.
 	 *
 	 * @param string|null $cursor Pagination cursor.
 	 * @return array|\WP_Error
 	 */
 	private static function fetch_notifications( ?string $cursor = null ): array|\WP_Error {
 		$params = array(
-			'reasons' => 'reply',
+			'reasons' => array( 'reply', 'like', 'repost' ),
 			'limit'   => 50,
 		);
 
@@ -136,14 +156,88 @@ class Reaction_Sync {
 	}
 
 	/**
-	 * Process a single reply notification into a WordPress comment.
+	 * Insert a reaction notification as a WordPress comment and persist its meta.
+	 *
+	 * Caller is responsible for target-post resolution, dedup, and comment_type.
+	 * This method handles the wp_insert_comment call and writes the standard
+	 * reaction meta (protocol, source_id, source_url, CID, author DID).
+	 *
+	 * @param int    $post_id        WP post the reaction attaches to.
+	 * @param string $comment_type   One of 'comment', 'like', 'repost'.
+	 * @param string $content        Comment body (reply text, or '' for like/repost).
+	 * @param int    $comment_parent Parent comment ID, 0 for top-level.
+	 * @param array  $notification   Raw notification from listNotifications.
+	 * @param array  $profile        Resolved author profile (name, handle, avatar).
+	 * @return int|false Comment ID or false.
+	 */
+	private static function insert_reaction(
+		int $post_id,
+		string $comment_type,
+		string $content,
+		int $comment_parent,
+		array $notification,
+		array $profile
+	): int|false {
+		$uri    = $notification['uri'] ?? '';
+		$cid    = $notification['cid'] ?? '';
+		$author = $notification['author'] ?? array();
+		$record = $notification['record'] ?? array();
+
+		$author_handle = $profile['handle'] ?? ( $author['handle'] ?? '' );
+		$author_name   = $profile['name'] ?? $author_handle;
+
+		$comment_data = array(
+			'comment_post_ID'      => $post_id,
+			'comment_parent'       => $comment_parent,
+			'comment_author'       => $author_name,
+			'comment_author_url'   => 'https://bsky.app/profile/' . $author_handle,
+			'comment_author_email' => '',
+			'comment_content'      => \wp_kses_post( $content ),
+			'comment_date_gmt'     => \get_gmt_from_date( $record['createdAt'] ?? '' ),
+			'comment_type'         => $comment_type,
+			'comment_approved'     => 1,
+			'comment_agent'        => 'ATmosphere/' . ATMOSPHERE_VERSION,
+		);
+
+		\remove_action( 'check_comment_flood', 'check_comment_flood_db' );
+		$comment_id = \wp_insert_comment( $comment_data );
+		\add_action( 'check_comment_flood', 'check_comment_flood_db', 10, 4 );
+
+		if ( ! $comment_id ) {
+			return false;
+		}
+
+		\update_comment_meta( $comment_id, self::META_PROTOCOL, 'atproto' );
+		\update_comment_meta( $comment_id, self::META_SOURCE_ID, $uri );
+		\update_comment_meta(
+			$comment_id,
+			self::META_SOURCE_URL,
+			self::build_bsky_web_url( $uri, $author_handle )
+		);
+		\update_comment_meta( $comment_id, self::META_BSKY_CID, $cid );
+		\update_comment_meta( $comment_id, self::META_AUTHOR_DID, $author['did'] ?? '' );
+
+		/**
+		 * Fires after a Bluesky reaction is synced as a WordPress comment.
+		 *
+		 * @param int    $comment_id   The new comment ID.
+		 * @param array  $notification The raw notification data.
+		 * @param int    $post_id      The WordPress post ID.
+		 * @param string $comment_type One of 'comment', 'like', 'repost'.
+		 */
+		\do_action( 'atmosphere_reaction_synced', $comment_id, $notification, $post_id, $comment_type );
+
+		return $comment_id;
+	}
+
+	/**
+	 * Process a reply notification into a WordPress comment.
 	 *
 	 * @param array $notification Notification from listNotifications.
-	 * @return int|false Comment ID or false if skipped.
+	 * @return int|false Comment ID or false.
 	 */
 	private static function process_reply( array $notification ): int|false {
 		$reply_uri = $notification['uri'] ?? '';
-		$reply_cid = $notification['cid'] ?? '';
 		$record    = $notification['record'] ?? array();
 		$author    = $notification['author'] ?? array();
 
@@ -151,7 +245,6 @@ class Reaction_Sync {
 			return false;
 		}
 
-		// Dedup: skip if already imported.
 		if ( self::find_comment_by_source_id( $reply_uri ) ) {
 			return false;
 		}
@@ -163,7 +256,6 @@ class Reaction_Sync {
 			return false;
 		}
 
-		// Try to match the parent to a local post or comment.
 		$post_id        = 0;
 		$comment_parent = 0;
 
@@ -190,63 +282,22 @@ class Reaction_Sync {
 			return false;
 		}
 
-		// Resolve author profile.
+		$text = $record['text'] ?? '';
+
+		if ( '' === $text ) {
+			return false;
+		}
+
 		$profile = self::resolve_author( $author['did'] ?? '' );
 
-		$comment_text = $record['text'] ?? '';
-
-		if ( empty( $comment_text ) ) {
-			return false;
-		}
-
-		$author_handle = $profile['handle'] ?? ( $author['handle'] ?? '' );
-		$author_name   = $profile['name'] ?? $author_handle;
-
-		$comment_data = array(
-			'comment_post_ID'      => $post_id,
-			'comment_parent'       => $comment_parent,
-			'comment_author'       => $author_name,
-			'comment_author_url'   => 'https://bsky.app/profile/' . $author_handle,
-			'comment_author_email' => '',
-			'comment_content'      => \wp_kses_post( $comment_text ),
-			'comment_date_gmt'     => \get_gmt_from_date( $record['createdAt'] ?? '' ),
-			'comment_type'         => 'comment',
-			'comment_approved'     => 1,
-			'comment_agent'        => 'ATmosphere/' . ATMOSPHERE_VERSION,
+		return self::insert_reaction(
+			$post_id,
+			'comment',
+			$text,
+			$comment_parent,
+			$notification,
+			$profile
 		);
-
-		// Disable flood control for programmatic insertion.
-		\remove_action( 'check_comment_flood', 'check_comment_flood_db' );
-
-		$comment_id = \wp_insert_comment( $comment_data );
-
-		\add_action( 'check_comment_flood', 'check_comment_flood_db', 10, 4 );
-
-		if ( ! $comment_id ) {
-			return false;
-		}
-
-		// Store AT Protocol metadata.
-		\update_comment_meta( $comment_id, self::META_PROTOCOL, 'atproto' );
-		\update_comment_meta( $comment_id, self::META_SOURCE_ID, $reply_uri );
-		\update_comment_meta(
-			$comment_id,
-			self::META_SOURCE_URL,
-			self::build_bsky_web_url( $reply_uri, $author_handle )
-		);
-		\update_comment_meta( $comment_id, self::META_BSKY_CID, $reply_cid );
-		\update_comment_meta( $comment_id, self::META_AUTHOR_DID, $author['did'] ?? '' );
-
-		/**
-		 * Fires after a Bluesky reply is synced as a WordPress comment.
-		 *
-		 * @param int   $comment_id   The new comment ID.
-		 * @param array $notification The raw notification data.
-		 * @param int   $post_id      The WordPress post ID.
-		 */
-		\do_action( 'atmosphere_reaction_synced', $comment_id, $notification, $post_id );
-
-		return $comment_id;
 	}
 
 	/**
--- a/includes/class-reaction-sync.php
+++ b/includes/class-reaction-sync.php
@@ -65,6 +65,62 @@ class Reaction_Sync {
 	public const META_AUTHOR_DID = '_atmosphere_author_did';
 
 	/**
+	 * Comment meta key for the reaction author's avatar URL.
+	 *
+	 * Atproto-specific. Populated at insert time from the Bluesky
+	 * profile so get_avatar() does not fall through to gravatar.
+	 *
+	 * @var string
+	 */
+	public const META_AUTHOR_AVATAR = '_atmosphere_author_avatar';
+
+	/**
+	 * Register display-side hooks.
+	 */
+	public static function register(): void {
+		\add_filter( 'get_avatar_comment_types', array( self::class, 'avatar_comment_types' ) );
+		\add_filter( 'pre_get_avatar_data', array( self::class, 'filter_avatar_data' ), 10, 2 );
+	}
+
+	/**
+	 * Tell WordPress that like and repost comments are avatar-eligible.
+	 *
+	 * @param array $types Registered avatar-eligible comment types.
+	 * @return array
+	 */
+	public static function avatar_comment_types( array $types ): array {
+		return \array_values( \array_unique( \array_merge( $types, array( 'comment', 'like', 'repost' ) ) ) );
+	}
+
+	/**
+	 * Short-circuit get_avatar_data for atproto-sourced comments.
+	 *
+	 * @param array $args        Avatar args.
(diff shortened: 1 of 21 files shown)


kraftbj:
One thing I want to flag before this merges: the watermark in `paginate()` advances to the first URI of the first page regardless of whether those items were actually processed. If a like or repost notification arrives before its target post has `_atmosphere_bsky_uri` meta written (network lag, a partially-published post, etc.), `process_subject_reaction` returns `false` and the reaction is silently dropped. The watermark has already advanced past it, so the next run doesn't retry — that notification is lost.

Two options:

- Only advance the watermark to the newest URI we actually processed or intentionally skipped as a non-reaction. Items dropped for missing local state shouldn't count toward the watermark.
- Or don't advance at all and rely on `find_comment_by_source_id` for dedup. Cost is up to `MAX_PAGES * 50` comment-meta lookups per run per stream.

Not a blocker if the answer is "document as known-miss and add an admin re-scan button later," but I'd want that path captured before shipping — silent data loss on a background sync is hard to debug after the fact.

kraftbj (commented):
Inline notes on a few lower-priority items from my review.

includes/class-reaction-sync.php
```diff
@@ -0,0 +1,579 @@
...
+	}
+
+	/**
+	 * Tell WordPress that like and repost comments are avatar-eligible.
+	 *
+	 * @param array $types Registered avatar-eligible comment types.
+	 * @return array
+	 */
+	public static function avatar_comment_types( array $types ): array {
+		return \array_values( \array_unique( \array_merge( $types, array( 'comment', 'like', 'repost' ) ) ) );
```
kraftbj, line 122: Minor: this runs on every `get_avatar_comment_types` call with a fixed added set ('comment', 'like', 'repost'). A static cache keyed on the input — or an early-return when those three are already present — avoids the repeated `array_unique`/`array_values` pass. Not a hot path, so only worth it if you care about the allocation.

includes/class-reaction-sync.php
```diff
@@ -0,0 +1,579 @@
...
+			static fn( ?string $cursor ) => self::fetch_notifications( $cursor ),
+			'notifications',
+			self::OPTION_LAST_SEEN_NOTIFICATION,
+			static fn( array $item ) => self::process_notification( $item )
+		);
+
+		// Self-reactions on our own posts (listNotifications omits these).
+		self::sync_own_collection( 'app.bsky.feed.like', 'like' );
+		self::sync_own_collection( 'app.bsky.feed.repost', 'repost' );
+		self::sync_own_collection( 'app.bsky.feed.post', 'comment' );
```
kraftbj, line 172: Back-of-envelope on the sync cost: `listNotifications` + three `listRecords` streams, up to 5 pages × 50 items each, so ~1,000 items worst case per run. Each one hits `find_post_by_bsky_uri` (one post-meta lookup; profile cache absorbs the `getProfile` duplicates). On a busy account, the hourly cron could get slow. Worth tracking as a follow-up — possibly batching the post lookups or adding an index on `_atmosphere_bsky_uri` — not blocking.

includes/class-reaction-sync.php
```diff
@@ -0,0 +1,579 @@
...
+		$author_handle = $profile['handle'] ?? ( $author['handle'] ?? '' );
+		$author_name   = $profile['name'] ?? $author_handle;
+
+		$gm_date = \get_gmt_from_date( $record['createdAt'] ?? '' );
+
+		$comment_data = array(
+			'comment_post_ID'      => $post_id,
+			'comment_parent'       => $comment_parent,
+			'comment_author'       => $author_name,
+			'comment_author_url'   => \esc_url_raw( 'https://bsky.app/profile/' . $author_handle ),
```
kraftbj: Two small things on this line:

1. `$author_handle` isn't URL-encoded before concat. AT Protocol handles are DNS-like so this is almost certainly safe in practice, but `rawurlencode( $author_handle )` costs nothing and is more defensive. Same at line 530 in `build_bsky_web_url`.

2. For likes and reposts, `source_url` is empty (no bsky.app page for those rkeys) and `comment_author_url` points at the profile. A reader clicking the comment author's name lands on the Bluesky profile, which is fine — but the meaning of each field then differs across comment types. Either a code comment documenting the intent, or swap so `source_url` holds the profile URL for likes/reposts, keeping both fields consistent.

includes/class-reaction-sync.php
```diff
@@ -0,0 +1,579 @@
...
+		}
+
+		$parts = \explode( '/', $at_uri );
+		$rkey  = \end( $parts );
+
+		if ( empty( $rkey ) || 'app.bsky.feed.post' !== \prev( $parts ) ) {
+			return '';
+		}
+
+		return \esc_url_raw( 'https://bsky.app/profile/' . $handle . '/post/' . $rkey );
```
kraftbj: Same handle-encoding note as line 435 — `rawurlencode( $handle )` before concat is the defensive move.

includes/class-reaction-sync.php
```diff
@@ -0,0 +1,579 @@
...
+	): int|false {
+		$uri    = $notification['uri'] ?? '';
+		$cid    = $notification['cid'] ?? '';
+		$author = $notification['author'] ?? array();
+		$record = $notification['record'] ?? array();
+
+		$author_handle = $profile['handle'] ?? ( $author['handle'] ?? '' );
+		$author_name   = $profile['name'] ?? $author_handle;
+
+		$gm_date = \get_gmt_from_date( $record['createdAt'] ?? '' );
```
kraftbj: Copying a finding from Copilot's review on the fosse bundled copy ([Automattic/fosse#15](https://github.com/Automattic/fosse/pull/15/files#r3118171010)) — you already flag this as a known follow-up in the PR body, but worth anchoring inline so it doesn't get lost:

> `$record['createdAt']` from Bluesky is already UTC (ISO-8601 with a timezone). Passing it through `get_gmt_from_date()` applies a local→UTC conversion again on non-UTC sites, which will store an incorrect `comment_date_gmt` (and derived `comment_date`). Prefer parsing `createdAt` as UTC and formatting `comment_date_gmt` directly from that UTC timestamp instead of using `get_gmt_from_date()` here.

`gmdate( 'Y-m-d H:i:s', strtotime( $record['createdAt'] ?? '' ) )` is the usual one-liner.

pfefferle:
Thanks for the thorough review! Addressed in 12963b4:

**Watermark data-loss concern (main thread)** — went with option 1. `paginate()` now re-walks a 10-item grace window past the prior watermark on each run. Dedup via `source_id` meta keeps already-processed items from re-inserting; transient drops (e.g. a reaction landing in the sub-second window between `applyWrites` success and the `_atmosphere_bsky_uri` meta write) get a second chance. Tests cover the grace-window re-walk, its page-spanning variant, and stream-end inside the window. Not a hard guarantee on very busy accounts that can exceed 10 reactions/run, but the realistic race window is sub-second and the cost is ~10 indexed dedup lookups per stream per run. I considered the tri-state `int|false|null` return contract as a follow-on, but the "authority == ours" heuristic for transient-vs-permanent has permanent-failure modes (deleted posts, drafts, missing meta) that would quietly reintroduce the same class of silent-failure bug in a different form, so I'd rather not until there's a concrete reason.

**L429 createdAt UTC** — switched to `\gmdate( 'Y-m-d H:i:s', \strtotime( ... ) )` with a regression test on a non-UTC timezone site. For the record, our WP version's `get_gmt_from_date` does respect embedded TZ via `date_create`, so the fosse finding doesn't strictly reproduce here — but the explicit form is simpler, independent of `wp_timezone()`, and tolerant of out-of-spec records without a TZ suffix.

**L435 + L530 rawurlencode** — applied to both the `insert_reaction` profile-URL build and `build_bsky_web_url`.

**L435 source_url semantics** — added a short comment in `insert_reaction` documenting that `source_url` is intentionally empty for likes/reposts (no bsky.app landing page) and `comment_author_url` is what links out for those.

**L108 avatar types caching** — skipping for now. An early-return doesn't fire in practice (WP's default input is `['comment']`, so our types are never pre-present). A one-slot memo on `$types` would work but felt like micro-optimization for a non-hot path; happy to add it if profiling ever flags it.

**L158 ~1000 items/run** — acknowledged as a follow-up; will track separately.

includes/class-reaction-sync.php
```diff
@@ -0,0 +1,620 @@
...
+			++$pages;
+
+			if ( ! $cursor ) {
+				break;
+			}
+		}
+
+		if ( $newest ) {
+			\update_option( $option_key, $newest, false );
+		}
```
kraftbj, line 280: Watermark advances unconditionally whenever `$newest` is set, including when the loop exited because we hit `MAX_PAGES` (5 pages × 50 items = 250). On a very quiet site this never matters. On a busy account that accumulates >250 reactions between runs — or on a first-run backfill after connecting — items past page 5 fall off the back and never get retried.

The grace-window fix addresses the sub-second publish race, which is a different failure mode; it doesn't help here because the watermark has advanced cleanly past items we never saw.

A defensive fix is to only persist the watermark when the loop exits via sentinel-reached or cursor-exhausted, not when it exits via MAX_PAGES. On a MAX_PAGES exit, either leave the option alone (next run re-walks from the top and dedups via `source_id`) or bump `MAX_PAGES` generously.

That said — I can see the counter-argument that this is overkill for the realistic workload (single-author WP blog, hourly cron, ~1000 items/hour is already an extremely active Bluesky account), and the cost of the defensive fix is extra dedup lookups on an unbounded window. What's your read? Happy to defer if you think the realistic worst case stays under the cap and we'd rather add a "rescan from epoch" admin button later. But the silent-advance behaviour feels like the same class of bug the grace window was designed to prevent, just at a different timescale.

includes/class-reaction-sync.php
```diff
@@ -0,0 +1,620 @@
...
+
+			foreach ( $items as $item ) {
+				$uri = $item['uri'] ?? '';
+
+				if ( null === $newest && $uri ) {
+					$newest = $uri;
+				}
+
+				if ( null === $rewalk && $last_seen && $uri === $last_seen ) {
+					$rewalk = self::WATERMARK_GRACE;
```
kraftbj: Small doc/behaviour mismatch: the docstring at `paginate()` says "keeps walking for up to `WATERMARK_GRACE` more items," but because the watermark URI itself is processed (and dedup-skipped) before the counter decrements, you actually re-walk `WATERMARK_GRACE - 1` items past the watermark — 9 when the constant is 10.

Not a correctness issue, but either the doc or the counter init should be nudged so they agree. Either:

```suggestion
				if ( null === $rewalk && $last_seen && $uri === $last_seen ) {
					$rewalk = self::WATERMARK_GRACE + 1;
				}
```

…or update the docstring to say "up to `WATERMARK_GRACE - 1` more items" / tune the constant to 11. Your call — just want code and comment aligned.

includes/class-reaction-sync.php
```diff
@@ -0,0 +1,620 @@
...
+			'comment_post_ID'      => $post_id,
+			'comment_parent'       => $comment_parent,
+			'comment_author'       => $author_name,
+			'comment_author_url'   => \esc_url_raw( 'https://bsky.app/profile/' . \rawurlencode( $author_handle ) ),
+			'comment_author_email' => '',
+			'comment_content'      => \wp_kses_post( $content ),
+			'comment_date'         => \get_date_from_gmt( $gm_date ),
+			'comment_date_gmt'     => $gm_date,
+			'comment_type'         => $comment_type,
+			'comment_approved'     => 1,
```
kraftbj, line 497: Question (not a blocker, but I'd like to surface it before this ships): `comment_approved => 1` is written unconditionally, so reply text from arbitrary Bluesky accounts lands published without going through `wp_allow_comment()` / Akismet / the moderation queue / `pre_comment_approved`.

For likes and reposts that's clearly fine (no body, no spam vector). For `comment_type = 'comment'` replies, what's the intended policy?

A few sub-questions:

- Do we want to rely on upstream (Bluesky) moderation exclusively, or let Akismet / `comment_moderation` also apply locally?
- If a reply is deleted on Bluesky (user deletes, labeler takedown, account suspension), do we plan a future sync pass that removes or unpublishes the mirrored comment? Or do we consider the mirrored copy independent once inserted?
- If an admin moderates the local copy (trashes, marks spam), does that round-trip anywhere?

I don't have strong opinions on the answer — could go either way and both are defensible. But the current code implicitly picks "upstream-only, one-way mirror, auto-approve" and I'd rather that be an explicit product decision than an implementation default. If the call is "ship as-is for v1, revisit," happy with that, just want it in the PR record.

includes/class-reaction-sync.php
```diff
@@ -0,0 +1,620 @@
...
+			'comment_agent'        => 'ATmosphere/' . ATMOSPHERE_VERSION,
+		);
+
+		\remove_action( 'check_comment_flood', 'check_comment_flood_db' );
+		$comment_id = \wp_insert_comment( $comment_data );
+		\add_action( 'check_comment_flood', 'check_comment_flood_db', 10, 4 );
```
kraftbj: Minor: I don't think this remove/add pair does anything. `check_comment_flood_db` is hooked to the `check_comment_flood` action, and that action is only fired by `wp_new_comment()` / `check_comment_flood_db()` — not by `wp_insert_comment()`, which is what we're calling on line 480. So we're suppressing a callback that was never going to run, then re-adding it.

Not harmful, but it's dead code and it implies we're taking precautions we aren't. Either drop the two lines, or switch to `wp_new_comment()` for the reply path (which would also address the moderation question above) and keep them meaningful.

includes/class-reaction-sync.php
```diff
@@ -0,0 +1,620 @@
...
+		$cursor    = null;
+		$pages     = 0;
+		$rewalk    = null;
+
+		while ( $pages < self::MAX_PAGES ) {
+			$response = $fetch( $cursor );
+
+			if ( \is_wp_error( $response ) ) {
+				return;
+			}
```
kraftbj, line 234: Note for future hardening (not blocking): `WP_Error` from the fetch just returns silently. Rate limits, OAuth/DPoP refresh failures, transient 5xx from the PDS — all look identical from the outside (sync quietly stops making progress). On a background cron with no UI surface, that's the kind of thing that goes unnoticed until someone asks "why haven't any reactions appeared in three weeks?"

A minimum `error_log()` of the `WP_Error` code + message on this path would at least give operators a breadcrumb. Longer term, a transient with last-error + last-success timestamp surfaced on the settings screen would be nice. Fine as a follow-up.

includes/class-reaction-sync.php
```diff
@@ -0,0 +1,620 @@
...
+		$comment_parent = 0;
+		$post_id        = self::find_post_by_bsky_uri( $parent_uri );
+
+		if ( ! $post_id ) {
+			// Nested reply: parent is an existing synced comment.
+			$parent_comment_id = self::find_comment_by_source_id( $parent_uri );
+
+			if ( $parent_comment_id ) {
+				$parent_comment = \get_comment( $parent_comment_id );
+				$post_id        = (int) $parent_comment->comment_post_ID;
```
kraftbj: Required before merge: `\get_comment()` can return `null` (comment deleted between the `find_comment_by_source_id` lookup and this call, cache miss, etc.). On PHP 8.2 the `->comment_post_ID` access on null is a fatal error, not a warning — which would take down the whole cron run for every reaction after it.

```suggestion
			if ( $parent_comment_id ) {
				$parent_comment = \get_comment( $parent_comment_id );

				if ( $parent_comment ) {
					$post_id        = (int) $parent_comment->comment_post_ID;
					$comment_parent = $parent_comment_id;
				}
			}
```

If `$parent_comment` is null we fall through to the root-post fallback on line 382, which is the right recovery.

kraftbj (changes requested):
Nice iteration — the grace-window + watermark design is a real improvement over the earlier version. A few questions about intent below, and at minimum the docs should match what the code actually does, plus one fatal-guard I'd like in before merge.

The big one is the MAX_PAGES interaction with the watermark: I think I understand the tradeoff, but I'd like your thoughts before it ships. The null-deref guard in `process_reply` is a hard request. The rest are small notes and one product question about moderation I'd rather surface now than in a follow-up.

pfefferle:
Thanks for the careful second pass! Addressed in cedda3b:

**L375 `get_comment()` null fatal (required)** — guarded. If `get_comment()` returns null we fall through to the root-post fallback, which matches your suggested patch. Added a regression test that uses the core `get_comment` filter to return null for the looked-up comment ID — without the fix it fails with `Attempt to read property "comment_post_ID" on null`, with the fix it passes and the reply gets attached at the root post.

**L237 grace-window off-by-one** — took the `+1` option: `$rewalk = self::WATERMARK_GRACE + 1` on watermark hit, so `WATERMARK_GRACE` items strictly past the watermark are re-walked (10 now means 10, not 9). Updated the docstring to say "strictly past" and adjusted the two paginate tests whose expected counts shifted.

**L481 dead `check_comment_flood` remove/add** — dropped. `wp_insert_comment` doesn't fire the action, so the pair was precautionary theatre. If we later decide replies should run the moderation pipeline we'd switch to `wp_new_comment` and the guard would become meaningful again.

**L221 silent `WP_Error`** — added a minimum `error_log()` breadcrumb with the error code, message, and stream/option key. Not a full observability story, but "why did sync stop?" now has a trace. Transient + settings-screen surface is a reasonable follow-up.

**L261 MAX_PAGES watermark advance on truncated walks** — deferring deliberately. You laid out the tradeoff well: the defensive fix (skip watermark write on MAX_PAGES exit) doesn't actually recover items past page 5 on a sustained high-rate account — the next run hits MAX_PAGES again against the same top 250, dedups them all, and items 251+ are still never reached. It also doesn't compose well with backfill: on first-run connection against a long-lived Bluesky account, we'd re-walk the same top 250 every hour forever. Realistic worst case (single-author blog, hourly cron, >250 reactions/hour) is already an unusually hot account, and the mitigations that actually help — cursor continuation on MAX_PAGES exit, or an admin "rescan from epoch" button — are bigger than this PR wants to absorb. Captured as follow-up; the admin rescan button is the likely shape.

**L475 moderation policy (explicit v1 decision)** — shipping as-is: `comment_approved => 1`, one-way mirror, rely on upstream moderation. Rationale: sync'd records don't look like local comments (no email, generic user agent), so Akismet would false-positive most of them; local moderation would effectively mean "reactions stop appearing with no feedback loop." Deletes-on-Bluesky / local-trash-round-trip are explicit non-goals for v1 — the mirrored copy is independent once inserted. If that changes we can hook a delete-reconcile pass and a `pre_comment_approved` filter entry point later.

kraftbj (approved):
Follow-up pass looks clean. All six items from the prior review landed well:

- **Null-deref guard (required)** — fixed with the filter-based regression test I was hoping for; the test fails on main without the guard and passes with it. Nice.
- **Grace-window +1** — the inline comment explaining *why* `GRACE + 1` makes the init match the docstring is exactly the right breadcrumb for the next person who reads this.
- **error_log breadcrumb** — minimal but sufficient, and the `phpcs:ignore` comment is scoped correctly.
- **Dead flood-check pair** — dropped cleanly.

The two deferrals are both reasoned well enough that I'm happy to ship:

- **MAX_PAGES watermark truncation** — your point that the defensive fix doesn't actually recover items past page 5 on a sustained high-rate account (next run dedups the same top 250, items 251+ still unreachable) is correct, and I hadn't fully thought through the first-run-backfill interaction. Admin "rescan from epoch" button is the right shape when we need it.
- **Moderation policy** — explicit v1 call: one-way mirror, trust upstream, no Akismet. The "sync'd records look unlike local comments so Akismet false-positives them all" argument is convincing. Good to have it in the PR record. If we ever want round-trip reconcile, `pre_comment_approved` and a delete-pass are the hook points.

Approving. Ship it.

PR #9, opened 2026-03-23
Add Markpub content parser

pfefferle:
## Proposed changes:
* Add Markpub parser (`at.markpub.markdown`) that walks Gutenberg blocks via `parse_blocks()` and converts each to GFM markdown.
* Register Markpub as the default content parser via the `atmosphere_content_parser` filter.
* Add tests for all supported block types and filter integration.

Stacked on #8. Merge that first.

### Other information:

## Testing instructions:

* Verify `?atproto` preview on a published post returns valid JSON with a `content` field containing `at.markpub.markdown` data.
* Verify block content converts to markdown correctly (headings, links, images, lists, code blocks).
* Verify classic editor content falls back to inline HTML-to-markdown conversion.
* Verify `atmosphere_content_parser` filter with `__return_null` disables the content field.
* Run `npm run env-test` — all tests pass.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Add rich content support for standard.site documents using the Markpub format.

</details>


includes/content-parser/class-markpub.php
```diff
@@ -0,0 +1,377 @@
...
+		return 'at.markpub.markdown';
+	}
+
+	/**
+	 * {@inheritDoc}
+	 *
+	 * @param string   $content Raw post content.
+	 * @param \WP_Post $post    The WordPress post object.
+	 */
+	public function parse( string $content, \WP_Post $post ): array { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
```
kraftbj: Small nit — the `phpcs:ignore` rule name here doesn't quite match what would actually fire. `VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable` targets variables, but `$post` is an unused *parameter* (required by the `Content_Parser` interface contract), so the matching sniff would be `Generic.CodeAnalysis.UnusedFunctionParameter`. Even nicer might be a short comment explaining that `$post` is required by the interface but this implementation doesn't need it — saves future readers from wondering why it's declared but unused.

includes/content-parser/class-markpub.php
```diff
@@ -0,0 +1,377 @@
...
+		$markdown = \apply_filters( 'atmosphere_html_to_markdown', $markdown, $content );
+
+		return array(
+			'$type'      => 'at.markpub.markdown',
+			'text'       => array(
+				'$type'    => 'at.markpub.text',
+				'markdown' => $markdown,
+			),
+			'flavor'     => 'gfm',
+			'extensions' => array( 'strikethrough', 'table' ),
```
kraftbj: I noticed `'table'` is advertised in `extensions`, but `core/table` isn't in the `match` above — it falls through to `fallback()` and ends up stripped by `wp_strip_all_tags()`. It might be worth either dropping `'table'` from this list until a table handler lands, or adding a minimal table transform so the advertised capability matches what the parser actually emits. Happy either way.

includes/content-parser/class-markpub.php
```diff
@@ -0,0 +1,377 @@
...
+		);
```
kraftbj, line 73: One thing to consider here: when every block returns `null` (e.g. a post of only `core/spacer` blocks) or when the inline conversions produce an empty string, this still returns a fully-formed record with `markdown => ''`. Downstream the Publisher would ship an empty content body to the PDS with no signal to the user. Would it make sense to return `null` (or an equivalent sentinel) when the trimmed markdown is empty but `$content` had non-trivial length, so `Document` can omit the `content` field? The test update at `tests/phpunit/tests/transformer/class-test-document.php` looks like it's already prepared to handle a null parser result.

kraftbj (commented):
Thanks for this — the parser structure is clean and the namespaced-prefix conventions look good throughout. Left a few inline notes on some edge cases I spotted while reading; none of them block the overall approach.

includes/content-parser/class-markpub.php
```diff
@@ -0,0 +1,377 @@
...
+
+		$md = '![' . $alt . '](' . $src . ')';
+
+		// Check for a caption in figcaption.
+		$caption_proc = new \WP_HTML_Tag_Processor( $html );
+		if ( $caption_proc->next_tag( 'FIGCAPTION' ) ) {
+			$caption = \wp_strip_all_tags(
+				\preg_replace( '#.*<figcaption[^>]*>#si', '', $html )
+			);
+			$caption = \trim( \preg_replace( '#</figcaption>.*#si', '', $caption ) );
```
kraftbj: Nice use of `WP_HTML_Tag_Processor` to detect the figcaption. One edge case worth flagging: because `wp_strip_all_tags()` on line 161 runs before the `</figcaption>.*` strip on line 164, the closing-tag marker is removed first — so any sibling content after the figcaption (e.g. a trailing `<p>` inside the same `<figure>`, or text nodes) gets concatenated into the caption text.

Swapping the order should handle it:

```php
$caption = \preg_replace( '#.*<figcaption[^>]*>#si', '', $html );
$caption = \preg_replace( '#</figcaption>.*#si', '', $caption );
$caption = \trim( \wp_strip_all_tags( $caption ) );
```

includes/content-parser/class-markpub.php
```diff
@@ -0,0 +1,377 @@
...
+		$md = \preg_replace( '#<(?:em|i)>(.*?)</(?:em|i)>#si', '*$1*', $md );
+
+		// Strikethrough.
+		$md = \preg_replace( '#<(?:s|del|strike)>(.*?)</(?:s|del|strike)>#si', '~~$1~~', $md );
+
+		// Inline code.
+		$md = \preg_replace( '#<code>(.*?)</code>#si', '`$1`', $md );
+
+		// Line breaks.
+		$md = \preg_replace( '#<br\s*/?\s*>#si', "  \n", $md );
```
kraftbj: Something to keep in mind for this run of `preg_replace()` calls (and the two in `image()` on lines 162/164): `preg_replace` returns `null` on PCRE failure, which the lazy `(.*?)` patterns here can trigger on very large or deeply nested inputs via backtrack/recursion limits. Once `$md` becomes `null`, each subsequent `preg_replace` emits a deprecation on PHP 8.2+ but still continues, and `\trim( null )` / `\wp_strip_all_tags( null )` quietly return `''` — so a single pathological input could erase the whole post body with no signal.

A light-touch guard — capturing the return and falling back to the prior `$md` (optionally with a `wp_trigger_error`) when it's `null` — would keep the failure visible. Happy to help sketch it out if useful.

pfefferle:
Thanks for the review! All five points addressed in 8991043:

**L36 `phpcs:ignore` rule + `$post` comment** — both `Generic.CodeAnalysis.UnusedFunctionParameter` and `VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable` actually fire on this parameter, so the ignore now lists both. Added a docblock note explaining that `$post` is required by the `Content_Parser` contract even though Markpub doesn't use it.

**L65 advertised `table` extension** — dropped from the `extensions` array. No `core/table` handler yet; the advertised capability should match what's actually emitted. Can re-add when a table transform lands.

**L66 empty-markdown null sentinel** — widened `Content_Parser::parse` return type to `?array`. Markpub returns `null` when the trimmed markdown is empty; `Document::get_content` bails early on null and skips the `atmosphere_document_content` filter (keeping callbacks' `array $content` signature honest). New tests cover a post of only `core/spacer` blocks, literal empty post content, and the Document-side null-handling path (asserting the filter is not invoked).

**L164 figcaption sibling-bleed** — took your suggested order swap: `preg_replace` the surrounding figcaption markers first, then `wp_strip_all_tags`. Added a regression test with a trailing `<p>` inside the same `<figure>`; without the fix it fails with `"Real captionShould not appear in caption"` concatenated.

**L367 preg_replace PCRE guard** — added `safe_replace` / `safe_replace_callback` helpers that fall back to the unchanged subject and `wp_trigger_error` when `preg_replace` returns `null`. Applied to all inline conversions and the figcaption chain. Didn't TDD this one directly — reliably tripping PCRE limits in a unit test is awkward — but the new call sites are uniform and the helper is small enough to read at a glance.

kraftbj (dismissed):

kraftbj (dismissed):

pfefferle:
@kraftbj we should maybe find a more generic way, to also transform into the other standard.site content formats.

kraftbj:
Agreed — the single-parser filter and one-NSID-per-implementation shape will get awkward as soon as a second standard.site format lands. Filed #45 to nail down the interface before that happens (registry vs per-NSID filter vs canonical-IR projection — open for discussion).

For this PR, let's move forward with the current shape and iterate. Markpub is the only parser today, so the registry refactor doesn't block anything user-facing, and #45 keeps the design conversation in its own thread.

pfefferle:
Superseding this in favor of #112.

#112 generalizes the content-parser system into a selectable registry and migrates this Markpub parser (`at.markpub.markdown`) onto the new `Parser_Base`. The Markpub work from this branch is carried forward there — the parser, its tests, and the changelog entry — so this PR is fully absorbed. Closing to consolidate review on #112.

PR #32, opened 2026-04-23
Publish approved WordPress comments to Bluesky as replies

pfefferle:
### Scope

- **Bluesky only.** Comments publish as `app.bsky.feed.post` reply records. No `site.standard.document` or `site.standard.comment` is written — standard.site has no comment NSID, so there is no lexicon target. (Posts remain dual-written as bsky + document.)
- **Registered WP users only.** Anonymous commenters (name/email only) never publish; the eligibility gate requires `$comment->user_id > 0`.
- **Only on posts already published to AT Protocol.** The parent post must have both `Post::META_URI` and `Post::META_CID` (needed to build a valid reply.root strongRef); comments on not-yet-synced posts are skipped.

### Feature

- New `Transformer\Comment` turns an approved `\WP_Comment` into an `app.bsky.feed.post` reply record. Root is always the parent post's bsky record; parent resolves to a locally-published sibling comment, a federated (Reaction_Sync) parent, or falls back to the root when either URI or CID is missing (strongRef requires both).
- New `Publisher::publish_comment` / `update_comment` / `delete_comment` methods plus a TID-only `delete_comment_by_tid` for hard-delete after the row is gone. `update_comment` and `delete_comment` key off `META_URI` (set only after a successful API call), not `META_TID` (persisted locally by `get_rkey()`), so a previously-failed publish correctly retries as a create.
- Comment lifecycle hooks on `Atmosphere`: `transition_comment_status`, `comment_post`, `edit_comment`, and `delete_comment` schedule async publish / update / delete jobs via WP-Cron.
- Async cron handlers **re-check eligibility at fire time** so approve → unapprove and unapprove → re-approve races between enqueue and execution don't publish or delete unintended state.
- `Atmosphere::should_publish_comment()` exposes the gate and honors `atmosphere_should_publish_comment`. Filter-forward: `atmosphere_transform_comment` mutates the record before publish.

### Dedup contract (loop prevention)

- On successful publish, the reply's AT-URI is mirrored into `Reaction_Sync::META_SOURCE_ID`. When `Reaction_Sync::process_reply` next sees the same URI come back via `listRecords`, `find_comment_by_source_id` matches the local row and skips it — no duplicate row, no publish loop.
- Defence in depth in `is_comment_eligible`: comments stamped with the plugin's own `ATmosphere/` agent string (as `Reaction_Sync::insert_reaction` does) are rejected even if `META_PROTOCOL` has not yet been written.

### Refactor (bundled)

- `Publisher::publish` / `update` / `delete` renamed to `publish_post` / `update_post` / `delete_post` (and `delete_by_tids` → `delete_post_by_tids`). The generic entry points `publish` / `update` / `delete` now accept `\WP_Post|\WP_Comment` and dispatch by type, so polymorphic callers stay simple while async hooks that know their type call the explicit variant.
- Test files consolidated so each test class maps to a real class: `Test_Publisher` absorbs the comment tests, `Test_Reaction_Sync` absorbs the dedup regression, and `Test_Atmosphere` replaces the old `Test_Status_Change` and absorbs the comment eligibility tests.

### Other information:

## Testing instructions:

* Connect the plugin to a Bluesky account and publish any post (confirm bsky URI + CID are stored in post meta).
* Log in as a registered user (any role) and leave an approved comment on that post.
* In Bluesky, confirm a reply appears threaded on the original post.
* Edit the comment in WordPress → reply text updates on Bluesky.
* Unapprove (or trash) the comment → the Bluesky record is deleted.
* Approve → unapprove a comment before `wp cron event run atmosphere_publish_comment` fires: nothing publishes (eligibility is re-checked at fire time).
* Run the Reaction_Sync cron (`wp cron event run atmosphere_sync_reactions`) and confirm your outbound comment is **not** re-ingested as a new WP comment.
* For federated parent threading: reply (from WordPress) to a comment that was itself synced inbound from Bluesky, and confirm the AT record uses the federated parent's AT-URI as `reply.parent`.

### Changelog entry

<details>

#### Significance

- [x] Minor

#### Type

- [x] Added - for new features

#### Message

Publish replies from registered WordPress users to Bluesky as native replies, with edit and unapprove/delete synced back to the AT Protocol record.

</details>

### DIFF
--- a/atmosphere.php
+++ b/atmosphere.php
@@ -68,6 +68,10 @@ function deactivate() {
 	\wp_clear_scheduled_hook( 'atmosphere_sync_reactions' );
 	\wp_clear_scheduled_hook( 'atmosphere_sync_publication' );
 	\wp_clear_scheduled_hook( 'atmosphere_delete_records' );
+	\wp_clear_scheduled_hook( 'atmosphere_publish_comment' );
+	\wp_clear_scheduled_hook( 'atmosphere_update_comment' );
+	\wp_clear_scheduled_hook( 'atmosphere_delete_comment' );
+	\wp_clear_scheduled_hook( 'atmosphere_delete_comment_record' );
 	// Clear the legacy hook name in case an earlier PR-6 build scheduled it.
 	\wp_clear_scheduled_hook( 'atmosphere_sync_comments' );
 	\flush_rewrite_rules();
--- a/atmosphere.php
+++ b/atmosphere.php
@@ -64,16 +64,7 @@ function activate() {
  * Deactivation hook.
  */
 function deactivate() {
-	\wp_clear_scheduled_hook( 'atmosphere_refresh_token' );
-	\wp_clear_scheduled_hook( 'atmosphere_sync_reactions' );
-	\wp_clear_scheduled_hook( 'atmosphere_sync_publication' );
-	\wp_clear_scheduled_hook( 'atmosphere_delete_records' );
-	\wp_clear_scheduled_hook( 'atmosphere_publish_comment' );
-	\wp_clear_scheduled_hook( 'atmosphere_update_comment' );
-	\wp_clear_scheduled_hook( 'atmosphere_delete_comment' );
-	\wp_clear_scheduled_hook( 'atmosphere_delete_comment_record' );
-	// Clear the legacy hook name in case an earlier PR-6 build scheduled it.
-	\wp_clear_scheduled_hook( 'atmosphere_sync_comments' );
+	clear_scheduled_hooks();
 	\flush_rewrite_rules();
 }
 \register_deactivation_hook( __FILE__, __NAMESPACE__ . '\deactivate' );
--- a/includes/class-atmosphere.php
+++ b/includes/class-atmosphere.php
@@ -24,6 +24,31 @@ use Atmosphere\WP_Admin\Admin;
 class Atmosphere {
 
 	/**
+	 * Comment meta key tracking how many times publish has been
+	 * deferred waiting for a parent comment to publish first.
+	 *
+	 * @var string
+	 */
+	private const META_PUBLISH_ATTEMPTS = '_atmosphere_publish_attempts';
+
+	/**
+	 * Maximum re-schedule hops for a child comment waiting on a
+	 * not-yet-published parent. After this many deferrals the child
+	 * publishes as a top-level reply on the post (current fallback
+	 * behavior) so a stuck parent does not block it forever.
+	 *
+	 * @var int
+	 */
+	private const PARENT_DEFER_MAX_ATTEMPTS = 3;
+
+	/**
+	 * Seconds between parent-pending re-schedule hops.
+	 *
+	 * @var int
+	 */
+	private const PARENT_DEFER_DELAY_SECONDS = 30;
+
+	/**
 	 * Wire up all hooks.
 	 */
 	public function init(): void {
@@ -617,6 +642,10 @@ class Atmosphere {
 		 * bails when META_URI is absent (which it is pre-publish), so
 		 * without these guards a pre-cron unapprove would still
 		 * publish, and a pre-cron re-approve would still delete.
+		 *
+		 * Publisher WP_Error returns are logged rather than silently
+		 * dropped so a flaky PDS window or an expired refresh token
+		 * leaves a breadcrumb operators can find.
 		 */
 		\add_action(
 			'atmosphere_publish_comment',
@@ -628,7 +657,13 @@ class Atmosphere {
 				if ( ! self::should_publish_comment( $comment ) ) {
 					return;
 				}
-				Publisher::publish_comment( $comment );
+				if ( self::defer_when_parent_pending( $comment ) ) {
+					return;
+				}
+				\delete_comment_meta( $comment_id, self::META_PUBLISH_ATTEMPTS );
+
+				$result = Publisher::publish_comment( $comment );
+				self::log_cron_error( 'publish_comment', $comment_id, $result );
 			}
 		);
 
@@ -642,7 +677,9 @@ class Atmosphere {
 				if ( ! self::should_publish_comment( $comment ) ) {
 					return;
 				}
-				Publisher::update_comment( $comment );
+
+				$result = Publisher::update_comment( $comment );
+				self::log_cron_error( 'update_comment', $comment_id, $result );
 			}
 		);
 
@@ -658,19 +695,125 @@ class Atmosphere {
 				if ( self::should_publish_comment( $comment ) ) {
 					return;
 				}
-				Publisher::delete_comment( $comment );
+
+				$result = Publisher::delete_comment( $comment );
+				self::log_cron_error( 'delete_comment', $comment_id, $result );
 			}
 		);
 
 		\add_action(
 			'atmosphere_delete_comment_record',
 			static function ( string $tid ): void {
-				if ( '' !== $tid ) {
-					Publisher::delete_comment_by_tid( $tid );
+				if ( '' === $tid ) {
+					return;
+				}
+
+				$result = Publisher::delete_comment_by_tid( $tid );
+
+				if ( \is_wp_error( $result ) ) {
+					// Worst-case path: the WP comment row is already gone,
+					// so operators need the TID to clean up the orphan
+					// record manually.
+					\error_log( // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
+						\sprintf(
+							'[atmosphere] delete_comment_record tid=%s failed: %s — %s',
+							$tid,
+							$result->get_error_code(),
+							$result->get_error_message()
+						)
+					);
 				}
 			},
 			10,
 			1
+		);
+	}
+
+	/**
+	 * Defer a child comment publish when its parent is eligible but
+	 * has not published to the PDS yet.
+	 *
+	 * Comments are scheduled as independent single events with no
+	 * dependency ordering: if a user approves a parent and its reply
+	 * together, the child's cron event can fire first, see
+	 * resolve_parent_ref() return null, and publish flat as a
+	 * top-level reply on the root post. This defers the child a short
+	 * interval (up to PARENT_DEFER_MAX_ATTEMPTS hops) to give the
+	 * parent time to publish first. After the cap the child publishes
+	 * anyway using the root fallback — a stuck parent must not block
+	 * the child forever.
+	 *
+	 * @param \WP_Comment $comment Comment being published.
+	 * @return bool True when the publish was deferred, false to proceed now.
+	 */
+	private static function defer_when_parent_pending( \WP_Comment $comment ): bool {
+		$parent_id = (int) $comment->comment_parent;
+
+		if ( $parent_id <= 0 ) {
+			return false;
+		}
+
+		$parent = \get_comment( $parent_id );
+
+		if ( ! $parent instanceof \WP_Comment ) {
+			return false;
+		}
+
+		if ( ! self::should_publish_comment( $parent ) ) {
+			// Parent is ineligible (anon, rejected, etc.); resolve_parent_ref
+			// will fall back to root, which is the correct behavior.
+			return false;
+		}
+
+		if ( ! empty( \get_comment_meta( $parent_id, Comment::META_URI, true ) ) ) {
+			// Parent is already published — nothing to defer for.
+			return false;
+		}
+
+		$comment_id = (int) $comment->comment_ID;
+		$attempts   = (int) \get_comment_meta( $comment_id, self::META_PUBLISH_ATTEMPTS, true );
+
+		if ( $attempts >= self::PARENT_DEFER_MAX_ATTEMPTS ) {
+			// Give up and publish with root as parent; clear the counter
+			// so a future re-publish gets a fresh deferral budget.
+			\delete_comment_meta( $comment_id, self::META_PUBLISH_ATTEMPTS );
+			return false;
+		}
+
+		\update_comment_meta( $comment_id, self::META_PUBLISH_ATTEMPTS, $attempts + 1 );
+		\wp_schedule_single_event(
+			\time() + self::PARENT_DEFER_DELAY_SECONDS,
+			'atmosphere_publish_comment',
+			array( $comment_id )
+		);
+
+		return true;
+	}
+
+	/**
+	 * Log a WP_Error returned from a comment cron Publisher call.
+	 *
+	 * `wp_schedule_single_event` does not retry, so a silent drop
+	 * here would lose the breadcrumb operators need to diagnose
+	 * auth, transport, or PDS-side failures.
+	 *
+	 * @param string $op         One of 'publish_comment' | 'update_comment' | 'delete_comment'.
+	 * @param int    $comment_id Comment ID.
+	 * @param mixed  $result     Publisher call result.
+	 */
+	private static function log_cron_error( string $op, int $comment_id, $result ): void {
+		if ( ! \is_wp_error( $result ) ) {
+			return;
+		}
+
+		\error_log( // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
+			\sprintf(
+				'[atmosphere] %s %d failed: %s — %s',
+				$op,
+				$comment_id,
+				$result->get_error_code(),
+				$result->get_error_message()
+			)
 		);
 	}
 }
--- a/includes/class-atmosphere.php
+++ b/includes/class-atmosphere.php
@@ -10,7 +10,9 @@ namespace Atmosphere;
 \defined( 'ABSPATH' ) || exit;
 
 use Atmosphere\OAuth\Client;
+use Atmosphere\Transformer\Comment;
 use Atmosphere\Transformer\Document;
+use Atmosphere\Transformer\Post;
 use Atmosphere\Transformer\Publication;
 use Atmosphere\Transformer\TID;
 use Atmosphere\Integrations\Load;
@@ -53,6 +55,12 @@ class Atmosphere {
 
 		// Catch permanent deletes (bypassing trash or emptying trash).
 		\add_action( 'before_delete_post', array( $this, 'on_before_delete' ) );
+
+		// Comment lifecycle hooks.
+		\add_action( 'transition_comment_status', array( $this, 'on_comment_status_change' ), 10, 3 );
+		\add_action( 'comment_post', array( $this, 'on_comment_insert' ), 10, 2 );
+		\add_action( 'edit_comment', array( $this, 'on_comment_edit' ) );
+		\add_action( 'delete_comment', array( $this, 'on_comment_before_delete' ) );
 
 		// Auto-sync publication when site identity changes.
 		\add_action( 'update_option_blogname', array( $this, 'schedule_publication_sync' ) );
@@ -308,6 +316,206 @@ class Atmosphere {
 	}
 
 	/**
+	 * Handle a comment transitioning between approval states.
+	 *
+	 * @param string      $new_status New comment_approved value.
+	 * @param string      $old_status Previous comment_approved value.
+	 * @param \WP_Comment $comment    Comment object.
+	 */
+	public function on_comment_status_change( string $new_status, string $old_status, \WP_Comment $comment ): void {
+		if ( $new_status === $old_status ) {
+			return;
+		}
+
+		if ( 'approved' === $new_status ) {
+			$this->schedule_comment_publish( $comment );
+			return;
+		}
+
+		if ( 'approved' === $old_status ) {
+			$this->schedule_comment_delete( $comment );
+		}
+	}
+
+	/**
+	 * Handle a newly-inserted comment.
+	 *
+	 * Covers the case where a comment lands already-approved (trusted
+	 * author), for which transition_comment_status does not fire.
+	 *
+	 * @param int        $comment_id       Comment ID.
+	 * @param int|string $comment_approved Approval status (1, 0, or 'spam').
+	 */
+	public function on_comment_insert( int $comment_id, int|string $comment_approved ): void {
+		if ( 1 !== (int) $comment_approved ) {
+			return;
+		}
+
+		$comment = \get_comment( $comment_id );
+		if ( $comment instanceof \WP_Comment ) {
+			$this->schedule_comment_publish( $comment );
+		}
+	}
+
+	/**
+	 * Handle a comment edit by updating its bsky record.
+	 *
+	 * @param int $comment_id Comment ID.
+	 */
+	public function on_comment_edit( int $comment_id ): void {
+		$comment = \get_comment( $comment_id );
+
+		if ( ! $comment instanceof \WP_Comment ) {
+			return;
+		}
+
+		if ( ! self::should_publish_comment( $comment ) ) {
+			return;
+		}
+
+		$hook = empty( \get_comment_meta( $comment_id, Comment::META_TID, true ) )
+			? 'atmosphere_publish_comment'
+			: 'atmosphere_update_comment';
+
+		if ( ! \wp_next_scheduled( $hook, array( $comment_id ) ) ) {
+			\wp_schedule_single_event( \time(), $hook, array( $comment_id ) );
+		}
+	}
+
+	/**
+	 * Capture a comment's TID before it is permanently deleted.
+	 *
+	 * Runs on delete_comment which fires before the row and meta are
+	 * removed, so the TID is still reachable. The TID-only cron variant
+	 * lets the async worker issue the PDS delete without re-reading
+	 * state that no longer exists.
+	 *
+	 * @param int $comment_id Comment ID.
+	 */
+	public function on_comment_before_delete( int $comment_id ): void {
+		if ( ! is_connected() ) {
+			return;
+		}
+
+		$tid = \get_comment_meta( $comment_id, Comment::META_TID, true );
+
+		if ( empty( $tid ) ) {
+			return;
+		}
+
+		\wp_schedule_single_event(
+			\time(),
+			'atmosphere_delete_comment_record',
+			array( (string) $tid )
+		);
+	}
+
+	/**
+	 * Eligibility gate for outbound comment publishing.
+	 *
+	 * @param \WP_Comment $comment Comment object.
+	 * @return bool
+	 */
+	public static function should_publish_comment( \WP_Comment $comment ): bool {
+		$should = self::is_comment_eligible( $comment );
+
+		/**
+		 * Filters whether a comment should be published to Bluesky.
+		 *
+		 * @param bool        $should  Whether to publish.
+		 * @param \WP_Comment $comment Comment object.
+		 */
+		return (bool) \apply_filters( 'atmosphere_should_publish_comment', $should, $comment );
+	}
+
+	/**
+	 * Core comment eligibility checks, pre-filter.
+	 *
+	 * @param \WP_Comment $comment Comment object.
+	 * @return bool
+	 */
+	private static function is_comment_eligible( \WP_Comment $comment ): bool {
+		if ( ! is_connected() ) {
+			return false;
+		}
+
+		if ( \in_array( (string) $comment->comment_type, array( 'trackback', 'pingback' ), true ) ) {
+			return false;
+		}
+
+		if ( (int) $comment->user_id <= 0 ) {
+			return false;
+		}
+
+		if ( '1' !== (string) $comment->comment_approved ) {
+			return false;
+		}
+
+		if ( 'atproto' === \get_comment_meta( (int) $comment->comment_ID, Reaction_Sync::META_PROTOCOL, true ) ) {
+			return false;
+		}
+
+		/*
+		 * Defence in depth: Reaction_Sync writes META_PROTOCOL after
+		 * wp_insert_comment, so if any caller ever fires comment_post
+		 * between the insert and the meta write, the gate above would
+		 * miss it. The sync always stamps its own agent string, which
+		 * is set before the insert — use it as a belt-and-braces check.
+		 */
+		if ( 0 === \strpos( (string) $comment->comment_agent, 'ATmosphere/' ) ) {
+			return false;
+		}
+
+		$post_uri = \get_post_meta( (int) $comment->comment_post_ID, Post::META_URI, true );
+		if ( empty( $post_uri ) ) {
+			return false;
+		}
+
+		return true;
+	}
+
+	/**
+	 * Schedule a publish or update event for a comment.
+	 *
+	 * @param \WP_Comment $comment Comment object.
+	 */
+	private function schedule_comment_publish( \WP_Comment $comment ): void {
+		if ( ! self::should_publish_comment( $comment ) ) {
+			return;
+		}
+
+		$comment_id = (int) $comment->comment_ID;
+		$hook       = empty( \get_comment_meta( $comment_id, Comment::META_TID, true ) )
+			? 'atmosphere_publish_comment'
+			: 'atmosphere_update_comment';
+
+		if ( \wp_next_scheduled( $hook, array( $comment_id ) ) ) {
+			return;
+		}
+
+		\wp_schedule_single_event( \time(), $hook, array( $comment_id ) );
+	}
+
+	/**
+	 * Schedule a delete event when a published comment leaves approved state.
+	 *
+	 * @param \WP_Comment $comment Comment object.
+	 */
+	private function schedule_comment_delete( \WP_Comment $comment ): void {
+		$comment_id = (int) $comment->comment_ID;
+
+		if ( empty( \get_comment_meta( $comment_id, Comment::META_TID, true ) ) ) {
+			return;
+		}
+
+		if ( \wp_next_scheduled( 'atmosphere_delete_comment', array( $comment_id ) ) ) {
+			return;
+		}
+
+		\wp_schedule_single_event( \time(), 'atmosphere_delete_comment', array( $comment_id ) );
+	}
+
+	/**
 	 * Schedule an async publication sync.
 	 */
 	public function schedule_publication_sync(): void {
@@ -340,7 +548,7 @@ class Atmosphere {
 			static function ( int $post_id ): void {
 				$post = \get_post( $post_id );
 				if ( $post && 'publish' === $post->post_status ) {
-					Publisher::publish( $post );
+					Publisher::publish_post( $post );
 				}
 			}
 		);
@@ -350,7 +558,7 @@ class Atmosphere {
 			static function ( int $post_id ): void {
 				$post = \get_post( $post_id );
 				if ( $post && 'publish' === $post->post_status ) {
-					Publisher::update( $post );
+					Publisher::update_post( $post );
 				}
 			}
 		);
@@ -360,7 +568,7 @@ class Atmosphere {
 			static function ( int $post_id ): void {
 				$post = \get_post( $post_id );
 				if ( $post ) {
-					Publisher::delete( $post );
+					Publisher::delete_post( $post );
 				}
 			}
 		);
@@ -375,10 +583,51 @@ class Atmosphere {
 		\add_action(
 			'atmosphere_delete_records',
 			static function ( string $bsky_tid, string $doc_tid ): void {
-				Publisher::delete_by_tids( $bsky_tid, $doc_tid );
+				Publisher::delete_post_by_tids( $bsky_tid, $doc_tid );
 			},
 			10,
 			2
+		);
+
+		\add_action(
+			'atmosphere_publish_comment',
+			static function ( int $comment_id ): void {
+				$comment = \get_comment( $comment_id );
+				if ( $comment instanceof \WP_Comment ) {
+					Publisher::publish_comment( $comment );
+				}
+			}
+		);
+
+		\add_action(
+			'atmosphere_update_comment',
+			static function ( int $comment_id ): void {
+				$comment = \get_comment( $comment_id );
+				if ( $comment instanceof \WP_Comment ) {
+					Publisher::update_comment( $comment );
+				}
+			}
+		);
+
+		\add_action(
+			'atmosphere_delete_comment',
+			static function ( int $comment_id ): void {
+				$comment = \get_comment( $comment_id );
+				if ( $comment instanceof \WP_Comment ) {
+					Publisher::delete_comment( $comment );
+				}
+			}
+		);
+
+		\add_action(
+			'atmosphere_delete_comment_record',
+			static function ( string $tid ): void {
+				if ( '' !== $tid ) {
+					Publisher::delete_comment_by_tid( $tid );
+				}
+			},
+			10,
+			1
 		);
 	}
 }
--- a/includes/class-atmosphere.php
+++ b/includes/class-atmosphere.php
@@ -319,8 +319,17 @@ class Atmosphere {
 	/**
 	 * Schedule AT Protocol record deletion before a post is permanently deleted.
 	 *
-	 * Captures TIDs from post meta before they're lost, then schedules
-	 * an async delete via cron.
+	 * Captures the post TIDs and the TIDs of every published outbound
+	 * comment reply, then schedules a single async batch delete via
+	 * cron. Comment TIDs must be collected here, while WP still has
+	 * the comment rows: `wp_delete_post( $id, true )` fires
+	 * `before_delete_post` first and only then iterates child comments,
+	 * so this is the last opportunity to read them.
+	 *
+	 * The trash path (`Publisher::delete_post()`) already cascades
+	 * comment deletes; this keeps the permanent-delete path symmetric
+	 * so unpublishing or hard-deleting a post does not orphan its
+	 * outbound replies on the PDS.
 	 *
 	 * @param int $post_id Post ID being deleted.
 	 */
@@ -338,8 +347,17 @@ class Atmosphere {
 		$bsky_tid = \get_post_meta( $post_id, Transformer\Post::META_TID, true );
 		$doc_tid  = \get_post_meta( $post_id, Transformer\Document::META_TID, true );
 
-		if ( $bsky_tid || $doc_tid ) {
-			\wp_schedule_single_event( \time(), 'atmosphere_delete_records', array( $bsky_tid, $doc_tid ) );
+		$comment_tids = \array_column(
+			Publisher::collect_published_comment_tids( $post_id ),
+			'tid'
+		);
+
+		if ( $bsky_tid || $doc_tid || ! empty( $comment_tids ) ) {
+			\wp_schedule_single_event(
+				\time(),
+				'atmosphere_delete_records',
+				array( $bsky_tid, $doc_tid, $comment_tids )
+			);
 		}
 	}
 
@@ -632,11 +650,15 @@ class Atmosphere {
 
 		\add_action(
 			'atmosphere_delete_records',
-			static function ( string $bsky_tid, string $doc_tid ): void {
-				Publisher::delete_post_by_tids( $bsky_tid, $doc_tid );
+			static function ( string $bsky_tid, string $doc_tid, $comment_tids = array() ): void {
+				Publisher::delete_post_by_tids(
+					$bsky_tid,
+					$doc_tid,
+					\is_array( $comment_tids ) ? $comment_tids : array()
+				);
 			},
 			10,
-			2
+			3
 		);
 
 		/*
--- a/includes/class-atmosphere.php
+++ b/includes/class-atmosphere.php
@@ -373,7 +373,7 @@ class Atmosphere {
 			return;
 		}
 
-		$hook = empty( \get_comment_meta( $comment_id, Comment::META_TID, true ) )
+		$hook = empty( \get_comment_meta( $comment_id, Comment::META_URI, true ) )
 			? 'atmosphere_publish_comment'
 			: 'atmosphere_update_comment';
 
@@ -386,14 +386,22 @@ class Atmosphere {
 	 * Capture a comment's TID before it is permanently deleted.
 	 *
 	 * Runs on delete_comment which fires before the row and meta are
-	 * removed, so the TID is still reachable. The TID-only cron variant
-	 * lets the async worker issue the PDS delete without re-reading
-	 * state that no longer exists.
+	 * removed, so the TID is still reachable. META_URI gates whether
+	 * there is anything to delete — a TID without URI is from a failed
+	 * earlier publish (no record on the PDS). The TID-only cron
+	 * variant lets the async worker issue the PDS delete without
+	 * re-reading state that no longer exists.
 	 *
 	 * @param int $comment_id Comment ID.
 	 */
 	public function on_comment_before_delete( int $comment_id ): void {
 		if ( ! is_connected() ) {
+			return;
+		}
+
+		$uri = \get_comment_meta( $comment_id, Comment::META_URI, true );
+
+		if ( empty( $uri ) ) {
 			return;
 		}
 
@@ -403,11 +411,14 @@ class Atmosphere {
 			return;
 		}
 
-		\wp_schedule_single_event(
-			\time(),
-			'atmosphere_delete_comment_record',
-			array( (string) $tid )
-		);
+		$tid  = (string) $tid;
+		$args = array( $tid );
+
+		if ( \wp_next_scheduled( 'atmosphere_delete_comment_record', $args ) ) {
+			return;
+		}
+
+		\wp_schedule_single_event( \time(), 'atmosphere_delete_comment_record', $args );
 	}
 
 	/**
@@ -466,8 +477,12 @@ class Atmosphere {
 			return false;
 		}
 
-		$post_uri = \get_post_meta( (int) $comment->comment_post_ID, Post::META_URI, true );
-		if ( empty( $post_uri ) ) {
+		$post_id  = (int) $comment->comment_post_ID;
+		$post_uri = \get_post_meta( $post_id, Post::META_URI, true );
+		$post_cid = \get_post_meta( $post_id, Post::META_CID, true );
+
+		// Both URI and CID are required to build a valid reply.root strongRef.
+		if ( empty( $post_uri ) || empty( $post_cid ) ) {
 			return false;
 		}
 
@@ -485,7 +500,7 @@ class Atmosphere {
 		}
 
 		$comment_id = (int) $comment->comment_ID;
-		$hook       = empty( \get_comment_meta( $comment_id, Comment::META_TID, true ) )
+		$hook       = empty( \get_comment_meta( $comment_id, Comment::META_URI, true ) )
 			? 'atmosphere_publish_comment'
 			: 'atmosphere_update_comment';
 
@@ -502,9 +517,13 @@ class Atmosphere {
 	 * @param \WP_Comment $comment Comment object.
 	 */
 	private function schedule_comment_delete( \WP_Comment $comment ): void {
+		if ( ! is_connected() ) {
+			return;
+		}
+
 		$comment_id = (int) $comment->comment_ID;
 
-		if ( empty( \get_comment_meta( $comment_id, Comment::META_TID, true ) ) ) {
+		if ( empty( \get_comment_meta( $comment_id, Comment::META_URI, true ) ) ) {
 			return;
 		}
 
--- a/includes/class-atmosphere.php
+++ b/includes/class-atmosphere.php
@@ -608,13 +608,27 @@ class Atmosphere {
 			2
 		);
 
+		/*
+		 * Cron handlers re-evaluate eligibility at fire time so state
+		 * changes between enqueue and execution (approve→unapprove,
+		 * unapprove→re-approve, user deleted, etc.) are respected. The
+		 * separate transition hooks only schedule; they cannot cancel
+		 * an already-queued event, and schedule_comment_delete itself
+		 * bails when META_URI is absent (which it is pre-publish), so
+		 * without these guards a pre-cron unapprove would still
+		 * publish, and a pre-cron re-approve would still delete.
+		 */
 		\add_action(
 			'atmosphere_publish_comment',
 			static function ( int $comment_id ): void {
 				$comment = \get_comment( $comment_id );
-				if ( $comment instanceof \WP_Comment ) {
-					Publisher::publish_comment( $comment );
+				if ( ! $comment instanceof \WP_Comment ) {
+					return;
 				}
+				if ( ! self::should_publish_comment( $comment ) ) {
+					return;
+				}
+				Publisher::publish_comment( $comment );
 			}
 		);
 
@@ -622,9 +636,13 @@ class Atmosphere {
 			'atmosphere_update_comment',
 			static function ( int $comment_id ): void {
 				$comment = \get_comment( $comment_id );
-				if ( $comment instanceof \WP_Comment ) {
-					Publisher::update_comment( $comment );
+				if ( ! $comment instanceof \WP_Comment ) {
+					return;
 				}
+				if ( ! self::should_publish_comment( $comment ) ) {
+					return;
+				}
+				Publisher::update_comment( $comment );
 			}
 		);
 
@@ -632,9 +650,15 @@ class Atmosphere {
 			'atmosphere_delete_comment',
 			static function ( int $comment_id ): void {
 				$comment = \get_comment( $comment_id );
-				if ( $comment instanceof \WP_Comment ) {
-					Publisher::delete_comment( $comment );
+				if ( ! $comment instanceof \WP_Comment ) {
+					return;
 				}
+				// If the comment is eligible again by the time cron
+				// fires, another transition has superseded the delete.
+				if ( self::should_publish_comment( $comment ) ) {
+					return;
+				}
+				Publisher::delete_comment( $comment );
 			}
 		);
 
--- a/includes/class-atmosphere.php
+++ b/includes/class-atmosphere.php
@@ -753,11 +753,27 @@ class Atmosphere {
 		\add_action(
 			'atmosphere_delete_records',
 			static function ( $bsky_tids, string $doc_tid, $comment_tids = array() ): void {
-				Publisher::delete_post_by_tids(
-					$bsky_tids,
-					$doc_tid,
-					\is_array( $comment_tids ) ? $comment_tids : array()
-				);
+				$comment_tids = \is_array( $comment_tids ) ? $comment_tids : array();
+				$result       = Publisher::delete_post_by_tids( $bsky_tids, $doc_tid, $comment_tids );
+
+				if ( \is_wp_error( $result ) ) {
+					/*
+					 * One-shot cron event with no retry: dropping this error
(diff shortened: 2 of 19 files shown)


includes/class-publisher.php
```diff
@@ -259,14 +297,145 @@ public static function sync_publication(): array|\WP_Error {
...
+	 *
+	 * @param \WP_Comment $comment WordPress comment.
+	 * @return array|\WP_Error
+	 */
+	public static function update_comment( \WP_Comment $comment ): array|\WP_Error {
+		$comment_id = (int) $comment->comment_ID;
+		$tid        = \get_comment_meta( $comment_id, Comment::META_TID, true );
+
+		if ( empty( $tid ) ) {
+			return self::publish_comment( $comment );
```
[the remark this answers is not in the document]
pfefferle: Confirmed and fixed in 4372ef0. `update_comment` and `delete_comment` now key off `META_URI` (set only after a successful API call), matching the post flow's reasoning. Added `test_update_comment_retries_create_when_tid_persisted_but_no_uri` as a regression guard for the stale-TID loop you described.

includes/class-atmosphere.php
```diff
@@ -307,6 +315,206 @@ public function on_before_delete( int $post_id ): void {
...
+		 * between the insert and the meta write, the gate above would
+		 * miss it. The sync always stamps its own agent string, which
+		 * is set before the insert — use it as a belt-and-braces check.
+		 */
+		if ( 0 === \strpos( (string) $comment->comment_agent, 'ATmosphere/' ) ) {
+			return false;
+		}
+
+		$post_uri = \get_post_meta( (int) $comment->comment_post_ID, Post::META_URI, true );
+		if ( empty( $post_uri ) ) {
```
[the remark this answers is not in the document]
pfefferle: Fixed in 4372ef0. `is_comment_eligible` now requires both `Post::META_URI` and `Post::META_CID`. The reply-root strongRef needs both, and this keeps invalid records out of the publish path. `test_comment_on_post_without_cid_is_skipped` covers it.

includes/class-atmosphere.php
```diff
@@ -307,6 +315,206 @@ public function on_before_delete( int $post_id ): void {
...
+		if ( empty( \get_comment_meta( $comment_id, Comment::META_TID, true ) ) ) {
+			return;
+		}
+
+		if ( \wp_next_scheduled( 'atmosphere_delete_comment', array( $comment_id ) ) ) {
+			return;
+		}
+
+		\wp_schedule_single_event( \time(), 'atmosphere_delete_comment', array( $comment_id ) );
+	}
```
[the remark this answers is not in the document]
pfefferle: Fixed in 4372ef0 — added the `is_connected()` guard to `schedule_comment_delete`, matching the post flow. `test_disconnected_state_does_not_schedule_comment_delete` covers it.

includes/class-atmosphere.php
```diff
@@ -307,6 +315,206 @@ public function on_before_delete( int $post_id ): void {
...
+
+		if ( empty( $tid ) ) {
+			return;
+		}
+
+		\wp_schedule_single_event(
+			\time(),
+			'atmosphere_delete_comment_record',
+			array( (string) $tid )
+		);
```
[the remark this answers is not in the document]
pfefferle: Fixed in 4372ef0 with the `wp_next_scheduled` guard you suggested. `test_comment_before_delete_does_not_double_schedule` covers it.

includes/transformer/class-comment.php
```diff
@@ -0,0 +1,174 @@
...
+
+		$federated_uri = \get_comment_meta( $parent_id, Reaction_Sync::META_SOURCE_ID, true );
+		if ( empty( $federated_uri ) ) {
+			return null;
+		}
+
+		return array(
+			'uri' => (string) $federated_uri,
+			'cid' => (string) \get_comment_meta( $parent_id, Reaction_Sync::META_BSKY_CID, true ),
+		);
```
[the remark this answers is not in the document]
pfefferle: Fixed in f2731b0. `resolve_parent_ref` now requires both URI and CID for both local and federated parents — missing either returns null so the caller falls back to the root ref. Matches the post-root guard added in the previous round.

tests/phpunit/tests/class-test-atmosphere.php
```diff
@@ -0,0 +1,518 @@
...
+			'comment_post_ID'  => $post_id,
+			'comment_approved' => '1',
+			'comment_type'     => 'comment',
+			'user_id'          => self::factory()->user->create(),
+			'comment_content'  => 'Hello.',
+		);
+
+		$comment_id = self::factory()->comment->create( \array_merge( $defaults, $overrides ) );
+
+		return \get_comment( $comment_id );
```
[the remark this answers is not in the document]
pfefferle: Fixed in f2731b0. Dropped the static cache — `make_eligible_comment` now creates a fresh post per call so each test owns its own fixture and survives the rollback between tests.

tests/phpunit/tests/transformer/class-test-comment.php
```diff
@@ -0,0 +1,272 @@
...
+				'comment_parent'  => $parent_id,
+				'user_id'         => 1,
+			)
+		);
+
+		$record = ( new Comment( \get_comment( $child_id ) ) )->transform();
+
+		$this->assertSame( 'at://did:plc:stranger/app.bsky.feed.post/nocid', $record['reply']['parent']['uri'] );
+		$this->assertSame( '', $record['reply']['parent']['cid'] );
+	}
```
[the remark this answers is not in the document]
pfefferle: Updated in f2731b0. The test now asserts root fallback (consistent with the tightened strongRef requirement), and I added a matching `test_reply_to_local_parent_without_cid_falls_back_to_root` for the local-publish edge case.

includes/class-atmosphere.php
```diff
@@ -375,10 +602,51 @@ static function (): void {
...
 		);
+
+		\add_action(
+			'atmosphere_publish_comment',
+			static function ( int $comment_id ): void {
+				$comment = \get_comment( $comment_id );
+				if ( $comment instanceof \WP_Comment ) {
+					Publisher::publish_comment( $comment );
+				}
+			}
```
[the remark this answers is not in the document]
pfefferle: Fixed in 6875ba8. The `atmosphere_publish_comment` handler now calls `should_publish_comment()` at fire time and bails when the comment is no longer eligible. `test_publish_comment_cron_rechecks_eligibility` covers the approve → unapprove race.

includes/class-atmosphere.php
```diff
@@ -375,10 +602,51 @@ static function (): void {
...
+		);
+
+		\add_action(
+			'atmosphere_update_comment',
+			static function ( int $comment_id ): void {
+				$comment = \get_comment( $comment_id );
+				if ( $comment instanceof \WP_Comment ) {
+					Publisher::update_comment( $comment );
+				}
```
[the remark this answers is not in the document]
pfefferle: Fixed in 6875ba8 — same guard pattern as the publish handler. The `atmosphere_update_comment` handler re-runs `should_publish_comment()` at fire time and skips the API call when the comment is no longer eligible.

includes/class-atmosphere.php
```diff
@@ -375,10 +602,51 @@ static function (): void {
...
+			}
+		);
+
+		\add_action(
+			'atmosphere_delete_comment',
+			static function ( int $comment_id ): void {
+				$comment = \get_comment( $comment_id );
+				if ( $comment instanceof \WP_Comment ) {
+					Publisher::delete_comment( $comment );
+				}
```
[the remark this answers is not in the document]
pfefferle: Fixed in 6875ba8. The `atmosphere_delete_comment` handler now skips when the comment has become eligible again at fire time, so an unapprove → re-approve sequence before cron runs keeps the remote record intact. `test_delete_comment_cron_skips_when_eligible_again` covers it.

includes/class-publisher.php
```diff
@@ -1,10 +1,11 @@
 <?php
 /**
- * Orchestrates publishing WordPress posts to the AT Protocol.
+ * Orchestrates publishing WordPress content to the AT Protocol.
  *
- * Every publish creates both an app.bsky.feed.post and a
- * site.standard.document atomically via a single applyWrites call.
- * Updates and deletions follow the same pattern.
+ * Posts go out as a bsky post + standard.site document pair via a
```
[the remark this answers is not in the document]
pfefferle: Fixed in 6875ba8 — header now reads `app.bsky.feed.post + site.standard.document` to match the actual lexicon.

includes/transformer/class-comment.php
```diff
@@ -0,0 +1,177 @@
+<?php
+/**
+ * Transforms an approved WordPress comment into an app.bsky.feed.post
+ * reply record.
+ *
+ * Only comments authored by a registered WordPress user are published;
+ * the Comment_Scheduler enforces that gate before instantiating this
+ * transformer. Root is always the parent post's bsky record; parent
+ * resolves to a previously-published sibling comment, a federated
+ * reply ingested via Reaction_Sync, or falls through to the root.
```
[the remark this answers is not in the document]
pfefferle: Fixed in 6875ba8. Header now points at `Atmosphere::should_publish_comment()` as the real gate — the `Comment_Scheduler` class was dropped in an earlier refactor.

kraftbj (commented):
Static code review only — no manual testing performed. Findings consolidated from the pr-review-toolkit agents and a Codex adversarial review pass.

## Blocking

1. **Unpublishing a post orphans its published comment replies on the PDS.** `delete_post()` deletes only the root post + document records; reply records published from WordPress comments are not enumerated or deleted. AT Protocol deletes do not cascade, so those replies stay live after the root is gone. (Inline comment on `delete_post()`.)
2. **Nested comments can publish before their parent and stay permanently mis-threaded.** When a parent comment is eligible but has not yet published, the transformer silently falls back to making the root post the parent. Cron events have no dependency ordering, so the child can publish as a top-level reply and will not be re-parented once the actual parent publishes. (Inline comment on `build_reply_ref()`.)
3. **Every new comment cron handler discards `WP_Error` from `Publisher`.** Transient PDS 5xx, expired OAuth, DPoP nonce loops, and transport failures silently drop the event. WordPress state ends up inconsistent with Bluesky state with no log line, no retry, and no admin signal. `atmosphere_delete_comment_record` is the worst case: the WP comment and its meta are already gone when the handler fires, so a failed `apply_writes` orphans the PDS record with nothing pointing back. (Inline comments on all four handlers.)
4. **`store_comment_result()` falls back to a locally-constructed URI when the PDS response lacks `results[0].uri`.** That URI is written to both `META_URI` and `META_SOURCE_ID` as if publish succeeded — mis-routing future updates and deletes, and suppressing inbound-sync dedup. (Inline comment on `store_comment_result()`.)

## Important

5. Test-coverage gaps: schedule-layer hooks (`on_comment_status_change`, `on_comment_insert`, `on_comment_edit`) and Publisher comment error-path side-effects. Five existing tests use `markTestSkipped` when the auth layer short-circuits, masking CI regressions. (Inline comments on the two test files.)

## Suggestions

6. Misleading docblock on `on_comment_before_delete`: "a TID without URI is from a failed earlier publish" — actually that is the normal pre-publish state since `Comment::get_rkey()` writes the TID eagerly. (Inline comment.)

## Pre-existing, out-of-scope (worth a follow-up)

These came up in the review but target code not modified by this PR — flagging for a separate issue or PR:

- **Reaction sync cron is scheduled on construct but only cleared on plugin deactivation.** Revoking OAuth at runtime leaves an hourly orphan event. `Reaction_Sync::sync()` short-circuits on `! is_connected()` so it is harmless, but the event never self-cleans. Fix in the disconnect path.
- **`Reaction_Sync::paginate()` advances the watermark from the first item of the first page regardless of per-item success.** A run of >10 item-level failures at the top of the stream permanently drops those items (`WATERMARK_GRACE = 10`). Track the highest *successfully*-processed URI separately, or explicitly document the 10-item grace as the only retry budget.
- **`Reaction_Sync::resolve_author()` swallows `getProfile` errors and caches no negative result.** A rate-limited or flapping profile endpoint produces blank-authored comments and amplifies the rate-limit pressure. Log the error and cache a short negative result.
- **`Reaction_Sync::process_reply()` / `process_subject_reaction()` cannot distinguish a dedup hit from a structurally malformed record.** Both return `false` with no breadcrumb. Add `error_log` on the non-dedup branches.

## Strengths

- Defence-in-depth eligibility gate with full test coverage.
- Cron fire-time re-eligibility check correctly wired with explicit regression tests for approve→unapprove and unapprove→re-approve races.
- `update_comment()` correctly keys off `META_URI` rather than `META_TID`, with a named regression test.
- Publish-loop prevention via `META_SOURCE_ID` mirroring plus the `ATmosphere/` agent sentinel, with explicit coverage.
- Changelog entry is punctuated and user-facing.

includes/class-publisher.php
```diff
@@ -130,7 +174,7 @@ public static function update( \WP_Post $post ): array|\WP_Error {
 	 * @param \WP_Post $post WordPress post.
 	 * @return array|\WP_Error
 	 */
-	public static function delete( \WP_Post $post ): array|\WP_Error {
+	public static function delete_post( \WP_Post $post ): array|\WP_Error {
```
kraftbj, line 901: **Blocking — cascade delete on post unpublish** (Codex adversarial review)

`delete_post()` only deletes the root `app.bsky.feed.post` and `site.standard.document` records. Outbound comment replies (records in `app.bsky.feed.post` keyed by `Comment::META_TID`) are independent repo records in the AT Protocol; the root delete does not cascade to them.

When a published post transitions out of `publish`, `on_status_change` schedules this path while the approved WordPress comments remain approved. Their reply records stay live on the PDS after the root is gone, with no remaining pointer from WordPress that could trigger cleanup.

**Suggested fix:** before building the root writes, query comments on this post for `Comment::META_TID` / `Comment::META_URI` and include delete writes for each in the same `apply_writes` batch (or a prior batch, reverting if it fails). Clear the comment meta on success. Add tests for publish→draft and publish→trash with published comments.

(Not verified via manual testing — static review inference.)

includes/transformer/class-comment.php
```diff
@@ -0,0 +1,178 @@
...
+			$resolved = $this->resolve_parent_ref( $parent_id );
+			if ( null !== $resolved ) {
+				return array(
+					'root'   => $root,
+					'parent' => $resolved,
+				);
+			}
+		}
+
+		return array(
```
kraftbj, line 135: **Blocking — parent-before-child ordering** (Codex adversarial review)

When a parent comment exists but has no stored URI+CID yet, `resolve_parent_ref()` returns `null` and this block silently falls back to making the root post the parent.

Comment publishes are scheduled as independent single events with no dependency ordering. Approving a parent and a child before cron runs can fire the child handler first, publishing it as a top-level reply. Nothing goes back and updates the child's `parent` ref once the true parent publishes later, so the remote thread is permanently flat.

**Suggested fix:** when `$parent_id > 0` and `resolve_parent_ref()` returns `null`, defer the publish — return a `WP_Error` that the cron handler can act on by rescheduling the child a short interval later, or detect the pending-parent state in `schedule_comment_publish()` and delay-enqueue the child.

Add a test that approves a parent and its child before any cron fires, then runs cron twice, and asserts the published child's `parent.uri` equals the parent comment's URI (not the post's).

includes/class-atmosphere.php
```diff
@@ -375,10 +602,75 @@ static function (): void {
...
+			'atmosphere_publish_comment',
+			static function ( int $comment_id ): void {
+				$comment = \get_comment( $comment_id );
+				if ( ! $comment instanceof \WP_Comment ) {
+					return;
+				}
+				if ( ! self::should_publish_comment( $comment ) ) {
+					return;
+				}
+				Publisher::publish_comment( $comment );
```
kraftbj: **Blocking — `WP_Error` returns from `Publisher::*` are discarded in all four comment cron handlers (this one and the three below).**

Every `Publisher` call on these four paths can return `WP_Error` for OAuth refresh failure, DPoP proof creation / decryption failure, DPoP nonce loops, transport failures (`wp_remote_request`), PDS 4xx/5xx, and `atmosphere_missing_tid`. All of them are thrown away here.

`wp_schedule_single_event` does not retry, so a transient PDS outage or an expired refresh token permanently drops the comment sync with no admin notice, no log breadcrumb, and no stored failure state.

**Suggested shape** (apply to publish, update, delete, and delete_record analogously):

```suggestion
				$result = Publisher::publish_comment( $comment );
				if ( \is_wp_error( $result ) ) {
					\error_log( // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
						\sprintf(
							'[atmosphere] publish_comment %d failed: %s — %s',
							$comment_id,
							$result->get_error_code(),
							$result->get_error_message()
						)
					);
				}
```

Consider follow-up: for transport-class failures (WP_HTTP errors, PDS 5xx), reschedule with backoff instead of logging-and-dropping — otherwise a flaky PDS window orphans every comment published during the outage.

includes/class-atmosphere.php
```diff
@@ -375,10 +602,75 @@ static function (): void {
...
+			'atmosphere_update_comment',
+			static function ( int $comment_id ): void {
+				$comment = \get_comment( $comment_id );
+				if ( ! $comment instanceof \WP_Comment ) {
+					return;
+				}
+				if ( ! self::should_publish_comment( $comment ) ) {
+					return;
+				}
+				Publisher::update_comment( $comment );
```
kraftbj: Same `WP_Error` swallowing pattern as the `atmosphere_publish_comment` handler above — needs `error_log` on the `Publisher::update_comment()` result.

includes/class-atmosphere.php
```diff
@@ -375,10 +602,75 @@ static function (): void {
...
+				$comment = \get_comment( $comment_id );
+				if ( ! $comment instanceof \WP_Comment ) {
+					return;
+				}
+				// If the comment is eligible again by the time cron
+				// fires, another transition has superseded the delete.
+				if ( self::should_publish_comment( $comment ) ) {
+					return;
+				}
+				Publisher::delete_comment( $comment );
```
kraftbj: Same `WP_Error` swallowing pattern. The "eligible again" re-check is good; just add the error-path log after `Publisher::delete_comment()`.

includes/class-atmosphere.php
```diff
@@ -375,10 +602,75 @@ static function (): void {
...
+			}
+		);
+
+		\add_action(
+			'atmosphere_delete_comment_record',
+			static function ( string $tid ): void {
+				if ( '' !== $tid ) {
+					Publisher::delete_comment_by_tid( $tid );
```
kraftbj: **Worst-case silent failure.** By the time this handler fires, the WP comment row and its meta are gone. If `delete_comment_by_tid()` returns `WP_Error`, the PDS record is orphaned with no remaining pointer from WordPress. Log the TID alongside the error code and message so it can be cleaned up manually.

includes/class-publisher.php
```diff
@@ -297,6 +487,30 @@ private static function store_results( int $post_id, array $result, Post $bsky_t
...
+	 * that when listRecords feeds our own reply back through the inbound
+	 * sync, find_comment_by_source_id() matches this row and skips it.
+	 *
+	 * @param int     $comment_id  WordPress comment ID.
+	 * @param array   $result      applyWrites response.
+	 * @param Comment $transformer Comment transformer.
+	 */
+	private static function store_comment_result( int $comment_id, array $result, Comment $transformer ): void {
+		$first = $result['results'][0] ?? array();
+		$uri   = $first['uri'] ?? $transformer->get_uri();
```
kraftbj: **Blocking — silent fallback to locally-constructed URI.**

On a non-`WP_Error` response with no `results[0].uri` field (malformed server response, partial failure that still returned 2xx, spec drift), `$uri` falls through to `$transformer->get_uri()` — an AT-URI built locally from `at://<did>/<collection>/<tid>`. That synthesized URI is then written to both `META_URI` and `META_SOURCE_ID` as though the publish succeeded.

Consequences:
- A malformed response is indistinguishable from a clean publish in WP state.
- `update_comment()` will take the `#update` branch on the next edit against a record that may not exist.
- `delete_comment()` will pass its URI gate for a record that may not exist.
- `find_comment_by_source_id()` will dedup-match the synthesized URI during inbound sync.

**Suggested fix:** treat missing `results[0].uri` as a failure — return a `WP_Error` to the caller (change return type to `\WP_Error|null`), and have `publish_comment()` / `update_comment()` propagate it so the cron handler can log it.

```suggestion
		$first = $result['results'][0] ?? array();
		$uri   = $first['uri'] ?? '';
		$cid   = $first['cid'] ?? '';

		if ( '' === $uri ) {
			return new \WP_Error(
				'atmosphere_missing_uri',
				\__( 'applyWrites response did not include a record URI.', 'atmosphere' )
			);
		}

		\update_comment_meta( $comment_id, Comment::META_URI, $uri );
		\update_comment_meta( $comment_id, Reaction_Sync::META_SOURCE_ID, $uri );
```

(The same `?? $transformer->get_uri()` pattern exists in `store_post_result` on trunk, but there the TID-keyed URI is deterministic enough to be safe. It does not carry over cleanly to comments.)

includes/class-atmosphere.php
```diff
@@ -307,6 +315,225 @@ public function on_before_delete( int $post_id ): void {
...
+			\wp_schedule_single_event( \time(), $hook, array( $comment_id ) );
+		}
+	}
+
+	/**
+	 * Capture a comment's TID before it is permanently deleted.
+	 *
+	 * Runs on delete_comment which fires before the row and meta are
+	 * removed, so the TID is still reachable. META_URI gates whether
+	 * there is anything to delete — a TID without URI is from a failed
```
kraftbj: **Suggestion — misleading docblock.**

This says "a TID without URI is from a failed earlier publish (no record on the PDS)." That is not accurate: `Comment::get_rkey()` persists the TID eagerly (before the `apply_writes` call) to guarantee idempotent rkey generation. So "TID without URI" is the *normal pre-publish state*, not exclusively a failed-publish state — it matches any comment whose transformer has been instantiated but whose publish has not yet completed.

The `Publisher::update_comment` docblock already explains this correctly. Align this docblock with that one — e.g.:

> `META_URI` is the only reliable signal that a record exists on the PDS; the TID is written eagerly by `Comment::get_rkey()` before the API call, so a TID alone does not mean anything was published.

tests/phpunit/tests/class-test-publisher.php
```diff
@@ -219,9 +221,328 @@ public function test_delete_errors_without_tids() {
...
+			'at://did:plc:test123/app.bsky.feed.post/newtid',
+			'bafynew'
+		);
+
+		$result = Publisher::publish_comment( \get_comment( $comment_id ) );
+		\remove_all_filters( 'pre_http_request' );
+
+		if ( \is_wp_error( $result ) ) {
+			$this->markTestSkipped( 'API layer rejected request before stub: ' . $result->get_error_message() );
+		}
```
kraftbj, line 469: **Important — test-coverage gaps for Publisher comment error paths.**

Two concerns:

1. **`markTestSkipped` on auth short-circuit** (here and in four sibling tests). When the auth layer blocks the request before the stub fires, the test silently skips and the body assertions never run. If auth begins blocking for an unexpected reason, these tests will skip in CI without anyone noticing. Mock `API::apply_writes` directly, or fail loudly when the stub was not invoked.

2. **No error-path assertions for comment writes.** There is no analogue to `test_delete_preserves_meta_on_api_error` for comments. Add:
   - `publish_comment` on API error: must not write `META_URI`, `META_CID`, or `META_SOURCE_ID` (ties to the `store_comment_result` issue).
   - `update_comment` on API error: must not overwrite `META_CID`.
   - `delete_comment` on API error: must not clear `META_URI`/`META_CID`/`META_TID`.

Approach: use `pre_http_request` returning a `WP_Error` or 500 JSON body, then assert post-call meta state.

tests/phpunit/tests/class-test-atmosphere.php
```diff
@@ -0,0 +1,588 @@
...
+
+		\update_option(
+			'atmosphere_connection',
+			array(
+				'access_token' => 'encrypted-token',
+				'did'          => 'did:plc:test123',
+				'pds_endpoint' => 'https://pds.example.com',
+			)
+		);
+	}
```
kraftbj, line 50: **Important — schedule-layer test gaps.**

The cron-handler tests (`test_publish_comment_cron_rechecks_eligibility`, `test_delete_comment_cron_skips_when_eligible_again`) exercise the fire-time path but the *schedule* path is not asserted for the common transitions. A regression making a scheduler stop enqueueing would pass all current tests.

Suggested additions:

- `on_comment_status_change` unapproved→approved: assert `wp_next_scheduled( 'atmosphere_publish_comment', [ $id ] )` is non-false.
- `on_comment_status_change` approved→unapproved: assert `atmosphere_delete_comment` is enqueued.
- `on_comment_insert` with `$comment_approved = 1` on an eligible post: assert `atmosphere_publish_comment` enqueued.
- `on_comment_insert` with `$comment_approved = 0` or `'spam'`: assert nothing enqueued.
- `on_comment_edit` with `META_URI` set: assert `atmosphere_update_comment` enqueued.
- `on_comment_edit` without `META_URI` on an approved comment: assert `atmosphere_publish_comment` enqueued.
- `on_comment_edit` on an unapproved comment: assert nothing enqueued.
- `on_comment_before_delete` with TID but no URI: assert nothing enqueued (guards the failed-publish state from double-scheduling a delete against a non-existent record).

includes/class-publisher.php
pfefferle: Fixed in 867d0d7. `Publisher::delete_post` now collects every comment on the post with a `Comment::META_URI` (set only after a successful publish, so stale TIDs from prior failed publishes are excluded) and adds an `app.bsky.feed.post#delete` write for each comment's TID into the same `applyWrites` batch. Comment meta is cleared on success. `test_delete_post_cascades_comment_replies` verifies the batch contents and the stale-TID exclusion.

includes/transformer/class-comment.php
pfefferle: Fixed in 867d0d7. The `atmosphere_publish_comment` handler now calls `defer_when_parent_pending()` before publishing: if the parent is eligible but its `META_URI` is not yet written, the child is re-scheduled 30s later (up to `PARENT_DEFER_MAX_ATTEMPTS = 3` hops, tracked via a `_atmosphere_publish_attempts` comment meta counter). After the cap the child publishes with the root fallback so a stuck parent does not block it forever. `test_publish_comment_defers_when_parent_pending` and `test_publish_comment_proceeds_after_parent_defer_cap` cover both paths.

includes/class-atmosphere.php
pfefferle: Fixed in 867d0d7 across all four handlers. Added a `log_cron_error()` helper that emits the same `[atmosphere] …` breadcrumb format `Reaction_Sync::paginate` uses, so auth / transport / PDS failures leave a trail operators can search. Transport-class retry/backoff is a reasonable follow-up — tracked for a later PR.

includes/class-atmosphere.php
pfefferle: Fixed in 867d0d7 via the shared `log_cron_error()` helper — `update_comment` now logs WP_Error returns identically to the publish handler.

includes/class-atmosphere.php
pfefferle: Fixed in 867d0d7 via the shared `log_cron_error()` helper — `delete_comment` now logs WP_Error returns identically to the publish and update handlers.

includes/class-atmosphere.php
pfefferle: Fixed in 867d0d7. The `atmosphere_delete_comment_record` handler now logs the TID, error code, and message on failure — since the WP row is gone by that point, the TID itself is the only handle operators have for cleaning up the orphan manually.

includes/class-publisher.php
pfefferle: Fixed in a135192. `store_comment_result` now returns `true|\WP_Error` — a response without `results[0].uri` yields `atmosphere_missing_uri`, which `publish_comment` / `update_comment` propagate to the cron handler (which logs it via `log_cron_error`). No more locally-synthesized URIs written to `META_URI` or `META_SOURCE_ID`. `test_publish_comment_errors_on_response_without_uri` covers the malformed-response path and `test_publish_comment_writes_no_meta_on_api_error` covers the transport-error path.

includes/class-atmosphere.php
pfefferle: Fixed in a135192. The docblock now reads: META_URI is the only reliable signal a record exists on the PDS; TID-only matches both the normal pre-publish state and a publish that failed after TID allocation, so neither case should schedule a delete. Aligned with `Publisher::update_comment`'s phrasing.

tests/phpunit/tests/class-test-publisher.php
pfefferle: Fixed the "no error-path assertions" half in a135192 — three new tests assert meta preservation / non-writing on API error (`test_publish_comment_writes_no_meta_on_api_error`, `test_update_comment_preserves_meta_on_api_error`, `test_delete_comment_preserves_meta_on_api_error`). These use the natural auth failure so they run assertions rather than skipping.

The `markTestSkipped` pattern on the existing payload-shape tests is a real concern, but fixing it properly means making the encryption layer mockable in tests — out of scope for this PR. Flagging as follow-up.

tests/phpunit/tests/class-test-atmosphere.php
pfefferle: Fixed in a135192. Added all eight suggested schedule-path tests on `Test_Atmosphere`: approve / unapprove transitions, approved / unapproved / spam inserts, edit with / without META_URI, edit on an unapproved comment, and the TID-without-URI before-delete guard. A regression that silently stops enqueueing on any of these paths would now fail a test.

tests/phpunit/tests/class-test-atmosphere.php
```diff
@@ -0,0 +1,846 @@
...
+			5,
+			3
+		);
+
+		\do_action( 'atmosphere_delete_comment', $comment_id );
+		\remove_all_filters( 'pre_http_request' );
+
+		$this->assertFalse( $captured, 'applyWrites#delete must not be called for a re-approved comment.' );
+		\remove_all_actions( 'atmosphere_delete_comment' );
+	}
```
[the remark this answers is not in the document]
pfefferle: Fixed in 2309ee6. The three `remove_all_actions` calls were leftovers from an earlier revision that also called `Atmosphere::register_async_hooks()` in the tests — when I removed the register calls I missed the matching cleanup. The plugin bootstrap registers these handlers once for the whole run, so the tests just clear the scheduled events they created and leave the global action alone.

tests/phpunit/tests/class-test-atmosphere.php
```diff
@@ -0,0 +1,846 @@
...
+			'Child must be rescheduled when parent is pending.'
+		);
+		$this->assertSame(
+			'1',
+			\get_comment_meta( $child_id, '_atmosphere_publish_attempts', true ),
+			'Deferral counter must be incremented on each hop.'
+		);
+
+		\remove_all_actions( 'atmosphere_publish_comment' );
+		\wp_clear_scheduled_hook( 'atmosphere_publish_comment', array( $child_id ) );
```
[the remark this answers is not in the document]
pfefferle: Fixed in 2309ee6. The three `remove_all_actions` calls were leftovers from an earlier revision that also called `Atmosphere::register_async_hooks()` in the tests — when I removed the register calls I missed the matching cleanup. The plugin bootstrap registers these handlers once for the whole run, so the tests just clear the scheduled events they created and leave the global action alone.

tests/phpunit/tests/class-test-atmosphere.php
```diff
@@ -0,0 +1,846 @@
...
+			5,
+			3
+		);
+
+		\do_action( 'atmosphere_publish_comment', $comment_id );
+		\remove_all_filters( 'pre_http_request' );
+
+		$this->assertFalse( $captured, 'applyWrites must not be called for a no-longer-eligible comment.' );
+		\remove_all_actions( 'atmosphere_publish_comment' );
+	}
```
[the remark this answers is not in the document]
pfefferle: Fixed in 2309ee6. The three `remove_all_actions` calls were leftovers from an earlier revision that also called `Atmosphere::register_async_hooks()` in the tests — when I removed the register calls I missed the matching cleanup. The plugin bootstrap registers these handlers once for the whole run, so the tests just clear the scheduled events they created and leave the global action alone.

includes/class-atmosphere.php
```diff
@@ -378,10 +633,193 @@ static function (): void {
...
+		 * bails when META_URI is absent (which it is pre-publish), so
+		 * without these guards a pre-cron unapprove would still
+		 * publish, and a pre-cron re-approve would still delete.
+		 *
+		 * Publisher WP_Error returns are logged rather than silently
+		 * dropped so a flaky PDS window or an expired refresh token
+		 * leaves a breadcrumb operators can find.
+		 */
+		\add_action(
+			'atmosphere_publish_comment',
```
kraftbj, line 797: **🔴 Must-fix — teardown regression vs PR #35.**

All four of these new cron hooks (this one and the three below) are missing from `atmosphere.php` `deactivate()` and from `uninstall.php`'s cleanup. PR #35 (which this branch just merged from trunk) added explicit `wp_clear_scheduled_hook()` calls for every other plugin-owned cron and called this out in its changelog. The four hooks added here were not extended into that pattern.

Consequences:
- After deactivate, queued events linger in `wp_options`, polluting the cron array.
- After uninstall + reinstall, a stale `atmosphere_delete_comment_record` event from a prior install fires against the new install's connection. `Publisher::delete_comment_by_tid()` does **not** verify the TID belongs to the current connection — this is a silent cross-install PDS delete attempt.
- A queued `atmosphere_publish_comment` after reactivation fires with a `comment_id` that may have been edited or unapproved in the meantime.

**Suggested fix** — add to `atmosphere.php`'s `deactivate()` and the corresponding block in `uninstall.php`:

```php
\wp_clear_scheduled_hook( 'atmosphere_publish_comment' );
\wp_clear_scheduled_hook( 'atmosphere_update_comment' );
\wp_clear_scheduled_hook( 'atmosphere_delete_comment' );
\wp_clear_scheduled_hook( 'atmosphere_delete_comment_record' );
```

`wp_clear_scheduled_hook()` without `$args` removes every queued event for that hook regardless of arguments (WP 4.9+), so no per-comment-id iteration is needed.

Also add a `commentmeta` sweep to `uninstall.php` for the new keys (the existing sweep is `postmeta`-only):

```php
$atmosphere_comment_meta_keys = array(
	'_atmosphere_bsky_tid',
	'_atmosphere_bsky_uri',
	'_atmosphere_bsky_cid',
	'_atmosphere_publish_attempts',
);
foreach ( $atmosphere_comment_meta_keys as $atmosphere_key ) {
	$wpdb->delete( $wpdb->commentmeta, array( 'meta_key' => $atmosphere_key ) );
}
```

(Inline anchor here because `atmosphere.php` and `uninstall.php` aren't in this PR's diff — the fix lives in those two files.)

includes/class-publisher.php
```diff
@@ -156,6 +206,15 @@ public static function delete( \WP_Post $post ): array|\WP_Error {
 			);
 		}
 
+		$comment_tids = self::collect_published_comment_tids( $post->ID );
```
kraftbj: **🟠 Should-fix — asymmetric cascade between trash and permanent-delete paths.**

The trash path is correct: `Publisher::delete_post()` enumerates `collect_published_comment_tids()` and includes them in the same `applyWrites` batch (this line). Good.

The permanent-delete path is not symmetric. `on_before_delete()` (`includes/class-atmosphere.php:327-344`, unchanged on this branch) reads only the post's two TIDs and schedules `atmosphere_delete_records` → `Publisher::delete_post_by_tids( $bsky_tid, $doc_tid )`, which has no knowledge of comment TIDs. By the time that cron fires the post is gone, so `get_comments( post_id )` returns nothing. The reply records orphan on Bluesky.

WordPress core does fire `delete_comment` for each comment before `before_delete_post` inside `wp_delete_post( $id, true )`, which de-facto papers over this — `on_comment_before_delete` schedules per-comment record deletes via `atmosphere_delete_comment_record`. But:

1. Nothing in this PR's tests asserts that ordering.
2. A plugin or extension that calls `wp_delete_post` with manual comment reassignment, or a future WP core refactor, breaks the cascade silently.
3. `on_before_delete()` short-circuits when `$post->post_type` isn't in `Backfill::syncable_post_types()`. If a site admin changes that filter after some comments are published, the cascade silently breaks for those posts.

**Suggested options:**

- Have `on_before_delete()` enumerate comment TIDs itself (before WP deletes them) and pass them through to `delete_post_by_tids()`.
- Or add an explicit test for `wp_delete_post( $id, true )` with published comments that asserts every reply record was deleted.

The class-level docblock on `Publisher::delete_post()` correctly identifies the cascade risk but only closes it for the trash path, not the permanent-delete path.

includes/class-atmosphere.php
```diff
@@ -378,10 +633,193 @@ static function (): void {
...
+				}
+				if ( ! self::should_publish_comment( $comment ) ) {
+					return;
+				}
+				if ( self::defer_when_parent_pending( $comment ) ) {
+					return;
+				}
+				\delete_comment_meta( $comment_id, self::META_PUBLISH_ATTEMPTS );
+
+				$result = Publisher::publish_comment( $comment );
```
kraftbj, line 811: **🟡 Follow-up — one-shot cron events permanently drop on transient API failures** (Codex adversarial review).

`Publisher::publish_comment()` (and the update / delete handlers below) can return `WP_Error` for OAuth refresh failure, DPoP nonce mismatch, transport flaps, PDS 4xx/5xx. `log_cron_error()` correctly logs the failure, but `wp_schedule_single_event` does not retry — there is no failed-state marker, no rescheduling, and no admin signal.

`Comment::get_rkey()` persists a TID before the network call to make the rkey idempotent. So a transient `applyWrites` failure leaves an approved comment in a half-published local state (TID present, URI absent) and no automatic event ever retries it unless a human edits or re-transitions the comment. Delete handlers have the mirror divergence — WP state has changed but the Bluesky record stays live.

**Suggestion (follow-up ticket OK):** persist a retryable failure state for outbound comment ops. Either reschedule failed publish/update/delete jobs with backoff (and a max-attempts cap), or maintain a durable pending-actions queue retried until success or surfaced in the admin UI.

includes/class-atmosphere.php
```diff
@@ -378,10 +633,193 @@ static function (): void {
...
+		);
+
+		\add_action(
+			'atmosphere_delete_comment_record',
+			static function ( string $tid ): void {
+				if ( '' === $tid ) {
+					return;
+				}
+
+				$result = Publisher::delete_comment_by_tid( $tid );
```
kraftbj, line 861: **🟡 Follow-up — hard-delete loses durable state on transient API failure** (Codex adversarial review).

This handler is the worst-case path. By the time it fires, `on_comment_before_delete()` has already let WordPress destroy the comment row and its meta. The TID survives only as the cron event's argument. If `Publisher::delete_comment_by_tid()` returns `WP_Error`:

- The PDS record is orphaned with no remaining WordPress pointer.
- The cron event itself does not survive once the handler returns.
- The only artifact is the `error_log` line, which a typical WordPress operator may never see.

**Suggestion (follow-up ticket OK):** capture pending hard-deletes in a durable plugin option/table *before* the comment row is removed (e.g. a `atmosphere_pending_deletes` array of TIDs), retried by a periodic cron until the remote delete succeeds. Logging stays useful, but recovery should not depend solely on it.

includes/class-publisher.php
```diff
@@ -156,6 +206,15 @@ public static function delete( \WP_Post $post ): array|\WP_Error {
...
+		foreach ( $comment_tids as $comment_tid ) {
+			$writes[] = array(
+				'$type'      => 'com.atproto.repo.applyWrites#delete',
+				'collection' => 'app.bsky.feed.post',
+				'rkey'       => $comment_tid['tid'],
+			);
+		}
+
 		$result = API::apply_writes( $writes );
```
kraftbj: **🟡 Follow-up — unbounded atomic `applyWrites` batch on post-delete** (Codex adversarial review).

`delete_post()` now appends one `applyWrites#delete` per published comment reply (lines 209–216) and submits everything in a single atomic request at this line. The batch size scales with thread length, has no chunking, and no fallback. On a popular post, this can:

- Exceed PDS write-count or payload limits.
- Fail under load with no partial-progress recovery.
- Block deletion of the root post + document together with every reply, since the batch is atomic.

Matters more on permanent-delete than trash, since trash can be retried by re-trashing.

**Suggestion (follow-up ticket OK):**

- Chunk comment deletions to a bounded batch size (e.g., 25–50 writes) and submit sequentially.
- Decouple the post + document deletes from the comment-reply deletes — root cleanup should not be blocked by a single oversized comment cascade.
- If the AT Protocol or your target PDS has a documented write-count limit, enforce it here and add a test for the overflow path.

Reference:`https://atproto.com/specs/atp` does not currently spec a hard limit, but most PDS implementations do.

kraftbj (commented):
Re-review after the trunk merges (PRs #33, #35, #37) and the two follow-up commits addressing my earlier review.

**Static code review only — no manual testing.** Findings consolidated from the pr-review-toolkit agents and a Codex adversarial review pass.

## ✅ Earlier-review items resolved

- `delete_post()` now cascades comment replies via `collect_published_comment_tids()`.
- `defer_when_parent_pending()` defers child publishes when the parent is eligible-but-unpublished, with a 3-hop cap; covered by tests.
- `log_cron_error()` is wired for all four comment cron handlers.
- `store_comment_result()` returns `WP_Error( 'atmosphere_missing_uri' )` instead of synthesizing a URI.
- Misleading docblock on `on_comment_before_delete` rewritten.
- Schedule-layer + Publisher comment error-path test coverage added.

Nice work on the follow-up.

## 🔴 Must-fix (regression vs PR #35, which this branch just merged)

Neither file is in this PR's diff, so the inline anchor is on the hook-registration site. Both need updating in this branch:

- **`atmosphere.php` `deactivate()`** (lines ~66–74 on this branch's tip): PR #35 added `wp_clear_scheduled_hook()` for every plugin-owned cron and explicitly stated the goal in its changelog ("so leftover jobs don't linger after the plugin is removed"). The four new comment crons (`atmosphere_publish_comment`, `atmosphere_update_comment`, `atmosphere_delete_comment`, `atmosphere_delete_comment_record`) were added by this branch but not added to the cleanup list. Worst case after deactivate→uninstall→reinstall: a queued `atmosphere_delete_comment_record` from the previous install fires against the new install's connection, attempting a PDS delete on a record from a different repo (`Publisher::delete_comment_by_tid()` doesn't verify the TID belongs to the current connection).
- **`uninstall.php`** needs the same four `wp_clear_scheduled_hook()` calls **and** a `commentmeta` sweep for the new keys (`_atmosphere_bsky_tid`, `_atmosphere_bsky_uri`, `_atmosphere_bsky_cid`, `_atmosphere_publish_attempts`). The existing sweep operates on `$wpdb->postmeta` only.

## 🟠 Should-fix

- Asymmetric cascade between the two delete paths — see inline comment on `Publisher::delete_post()`.

## 🟡 Follow-up tickets OK (Codex adversarial review)

- One-shot cron events permanently drop on transient API failures.
- Hard-delete loses durable state; only `error_log` evidence remains.
- Unbounded `applyWrites` batch on post-delete with many replies.

See inline comments on each.

## Secondary observations (out of scope for this review pass)

Not raised inline since you only asked for items 1–3 of the recommended action order, but flagging here for your judgement:

- `defer_when_parent_pending()` clears `_atmosphere_publish_attempts` only on the cap path. Non-defer early returns (`parent gone`, `parent ineligible`, `parent already published`) leave the counter stale, so a later edit-driven re-publish has a reduced defer budget. Low impact.
- `on_comment_before_delete()` silently returns when `META_URI` is empty but `META_TID` is present — this is correct for the normal pre-publish state, but masks the rare case where `META_URI` was cleared after a successful publish (filter, wp-cli, or external plugin). A breadcrumb on that branch would help operators surface orphans.
- `store_comment_result()` returning `atmosphere_missing_uri` on a 2xx response that omits `results[0].uri` is more correct than the prior synthesized-URI fallback, but the PDS write actually succeeded — that becomes an orphaned record on Bluesky with no WP-side pointer and no automatic recovery. Consider including the locally-generated TID in the WP_Error so logs can map back to the orphan, paving the way for a reconciliation job.

includes/class-atmosphere.php
pfefferle: Fixed in 0974054. `deactivate()` and `uninstall.php` now clear all four comment crons (`atmosphere_publish_comment`, `atmosphere_update_comment`, `atmosphere_delete_comment`, `atmosphere_delete_comment_record`). `uninstall.php` also gets the `commentmeta` sweep for `_atmosphere_bsky_tid`, `_atmosphere_bsky_uri`, `_atmosphere_bsky_cid`, `_atmosphere_publish_attempts`.

includes/class-publisher.php
pfefferle: Fixed in 612fa03 — went with the symmetric option. `on_before_delete()` now enumerates published comment TIDs via `Publisher::collect_published_comment_tids()` while the comment rows are still readable (before WP iterates child comments) and threads them through `atmosphere_delete_records` → `delete_post_by_tids()`, which appends one `applyWrites#delete` per reply to the same atomic batch. Tests cover the cascade + the unchanged no-comments path.

Kept META intact rather than clearing it after capture — the per-comment `atmosphere_delete_comment_record` crons still fire via WP's natural cascade. On the happy path that's noisy ("record not found" → `log_cron_error`), but on a failed atomic `applyWrites` the per-comment events give every reply an independent retry chance instead of orphaning the lot.

includes/class-atmosphere.php
pfefferle: Acknowledged as follow-up. Retry-with-backoff or a durable pending-actions queue is the right shape but out of scope for this PR. Tracked in #46.

includes/class-atmosphere.php
pfefferle: Acknowledged as follow-up. The cron-arg-only state is the weak link — a durable pending-deletes table replayed by a periodic worker is the right design. Out of scope here. Tracked in #47.

includes/class-publisher.php
pfefferle: Acknowledged as follow-up. The decoupling angle (post+doc delete should not block on a large comment cascade) is the more impactful piece — chunking only matters once threads grow. Tracked in #48.

pfefferle:
works quite well: https://bsky.app/profile/pfefferle.org/post/3mjz3o3gtjk2o

includes/class-atmosphere.php
```diff
@@ -366,9 +409,240 @@ public function on_before_delete( int $post_id ): void {
...
+		if ( 0 === \strpos( (string) $comment->comment_agent, 'ATmosphere/' ) ) {
+			return false;
+		}
+
+		$post_id  = (int) $comment->comment_post_ID;
+		$post_uri = \get_post_meta( $post_id, Post::META_URI, true );
+		$post_cid = \get_post_meta( $post_id, Post::META_CID, true );
+
+		// Both URI and CID are required to build a valid reply.root strongRef.
+		if ( empty( $post_uri ) || empty( $post_cid ) ) {
```
kraftbj, line 596: **🟡 Note — not blocking, but worth a deliberate decision** (Codex adversarial review)

`is_comment_eligible()` checks for the parent post's `Post::META_URI` / `META_CID` — i.e. "the post was at some point published to the PDS" — but does not verify the post's *current* status or whether it's still a supported post type. So an outbound comment can publish for a post that's now a draft, in the trash, switched to private, or whose post type was removed from the allowlist (as long as the post still has stale `_atmosphere_bsky_*` meta).

This is consistent with PR #38's split publish/cleanup gates philosophy — "once published, keep being maintained" — and an explicit decision is fine. Some questions worth resolving before merge:

- Should new outbound comment replies continue to flow for a post that's been moved to draft? (Most users probably don't expect new public Bluesky activity attached to unpublished WP content.)
- The cleanup path already correctly handles draft/trash transitions (PR #38), so unpublishing a post tears the records down — but `is_comment_eligible()` runs at the moderate-comment moment, which can be hours later. There's a window between unpublish and cleanup where a newly-approved comment publishes against a since-deleted root.

Not adding the `post_status === 'publish'` and `is_supported_post_type()` checks here would benefit from a one-line code comment explaining the trade-off (so a future contributor doesn't "fix" it). Adding them would close the window and is symmetric with `on_status_change`'s scheduling gate.

Not blocking — flagging for awareness.

includes/class-publisher.php
```diff
@@ -972,6 +1096,164 @@ public static function sync_publication(): array|\WP_Error {
...
+			),
+		);
+
+		$result = API::apply_writes( $writes );
+
+		if ( \is_wp_error( $result ) ) {
+			return $result;
+		}
+
+		$stored = self::store_comment_result( (int) $comment->comment_ID, $result );
```
kraftbj, line 1194: **🔴 In-flight publish can survive moderation delete** (Codex adversarial review)

The sequence:

1. Comment is approved → `atmosphere_publish_comment` cron fires.
2. `Comment::get_rkey()` persists the TID (in `transform()`).
3. `apply_writes` is in flight — the PDS is committing the record.
4. Admin unapproves or deletes the comment in WordPress.
5. `transition_comment_status` runs `schedule_comment_delete` → `on_comment_before_delete`. Both of those gates require `Comment::META_URI` to be present. **It isn't**, because we haven't reached `store_comment_result()` yet. So both gates short-circuit; nothing schedules a cleanup.
6. `apply_writes` returns 2xx. This call writes `META_URI` for a comment that's no longer approved (or no longer exists).

Result: a public Bluesky reply under the site identity with no scheduled cleanup event. The user has no way to know the record exists.

**Suggested fix:** after `apply_writes` returns success but before `store_comment_result()`, re-fetch the comment and re-run `should_publish_comment()`. If it's now ineligible (or `\get_comment()` returns null), immediately delete-by-TID using the rkey we already have, and skip the meta write. Add a regression test that flips the comment status during a `pre_http_request` filter so the API call appears to take time.

Correctness gap, not data-corruption — but a moderator-deleted comment showing up as a live Bluesky post under the site DID is a real concern.

includes/class-atmosphere.php
```diff
@@ -478,11 +752,198 @@ static function (): void {
 
 		\add_action(
 			'atmosphere_delete_records',
-			static function ( $bsky_tids, string $doc_tid ): void {
-				Publisher::delete_by_tids( $bsky_tids, $doc_tid );
+			static function ( $bsky_tids, string $doc_tid, $comment_tids = array() ): void {
+				Publisher::delete_post_by_tids(
```
kraftbj: **🔴 `atmosphere_delete_records` cron handler swallows `WP_Error` from `Publisher::delete_post_by_tids()`.**

Every other cron handler in this block (and in the comment block below) routes the `Publisher::*` return value through `log_cron_error()` or an inline `error_log()`. This one was untouched when commit `612fa037` started routing comment TIDs through it. The result is silently dropped on:

- Transport failures (`wp_remote_request` → DNS/TLS/timeout).
- PDS 4xx/5xx (including `InvalidRequest` for batch-too-large — see the unbounded-batch comment on `Publisher::delete_post_by_tids()`).
- Expired refresh token, DPoP nonce drift.

Commit `612fa037` materially raised the cost of swallowing this error: a failed batch now leaks every reply record on the post in addition to the root post and document.

**Suggested fix:**

```suggestion
		\add_action(
			'atmosphere_delete_records',
			static function ( $bsky_tids, string $doc_tid, $comment_tids = array() ): void {
				$result = Publisher::delete_post_by_tids(
					$bsky_tids,
					$doc_tid,
					\is_array( $comment_tids ) ? $comment_tids : array()
				);
				if ( \is_wp_error( $result ) ) {
					\error_log( // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
						\sprintf(
							'[atmosphere] delete_records failed (bsky=%d, doc=%s, comments=%d): %s — %s',
							\is_array( $bsky_tids ) ? \count( $bsky_tids ) : 1,
							$doc_tid ? 'yes' : 'no',
							\is_array( $comment_tids ) ? \count( $comment_tids ) : 0,
							$result->get_error_code(),
							$result->get_error_message()
						)
					);
				}
			},
			10,
			3
		);
```

includes/class-publisher.php
```diff
@@ -927,6 +1043,14 @@ public static function delete_by_tids( $bsky_tids, string $doc_tid ): array|\WP_
...
 
+		foreach ( $comment_tids as $comment_tid ) {
+			$writes[] = array(
+				'$type'      => 'com.atproto.repo.applyWrites#delete',
+				'collection' => 'app.bsky.feed.post',
+				'rkey'       => $comment_tid,
+			);
+		}
+
 		return API::apply_writes( $writes );
```
kraftbj: **🔴 Unbounded `applyWrites` batch on post-delete cascade** (Codex adversarial review)

`collect_published_comment_tids()` returns *every* match. `delete_post()` (line 888) and `delete_post_by_tids()` (this function) both append every result to a single `$writes` array and submit them in one atomic `apply_writes` call.

The AT Protocol `com.atproto.repo.applyWrites` lexicon caps writes per call at 200. A high-traffic post with 200+ published comment replies fails the entire batch with `InvalidRequest`. Combined with the silent-error issue on `atmosphere_delete_records` (above), this is invisible to the operator.

Because the batch is atomic, an oversized request blocks deletion of the root post + document together with every reply record.

**Suggested fix:** chunk to a bounded batch size (e.g. `array_chunk( $writes, 100 )`), submit sequentially, and aggregate results. Surface partial-failure cleanly so an operator can know which chunk failed. Reference: AT Protocol lexicon `com.atproto.repo.applyWrites` (`writes` array maxLength: 200).

This was previously flagged as a follow-up; commit `612fa037` materially worsened the impact by adding the comment cascade through the same atomic call.

atmosphere.php
```diff
@@ -68,6 +68,10 @@ function deactivate() {
 	\wp_clear_scheduled_hook( 'atmosphere_sync_reactions' );
 	\wp_clear_scheduled_hook( 'atmosphere_sync_publication' );
 	\wp_clear_scheduled_hook( 'atmosphere_delete_records' );
```
kraftbj: **🔴 `deactivate()` is missing three post crons that `uninstall.php` clears.**

`uninstall.php` (lines 22–24) clears `atmosphere_publish_post`, `atmosphere_update_post`, and `atmosphere_delete_post`. `deactivate()` does not. After `deactivate → reactivate` (or `deactivate → reconnect to a different account → reactivate`), queued post-publish/update/delete cron events fire against the now-running plugin. If the connection has changed between the two activations (different DID), the events publish or delete records on the wrong repo.

This is the same cross-install risk that motivated commit `0974054` for the comment crons; the post crons were missed. The two lists should be in lock-step.

**Suggested fix** — add to `deactivate()`:

```suggestion
	\wp_clear_scheduled_hook( 'atmosphere_refresh_token' );
	\wp_clear_scheduled_hook( 'atmosphere_sync_reactions' );
	\wp_clear_scheduled_hook( 'atmosphere_sync_publication' );
	\wp_clear_scheduled_hook( 'atmosphere_publish_post' );
	\wp_clear_scheduled_hook( 'atmosphere_update_post' );
	\wp_clear_scheduled_hook( 'atmosphere_delete_post' );
	\wp_clear_scheduled_hook( 'atmosphere_delete_records' );
```

Follow-up consideration: extract a shared `atmosphere_clear_all_crons()` helper used by both `deactivate()` and `uninstall.php` so this can't drift again.

atmosphere.php
```diff
@@ -68,6 +68,10 @@ function deactivate() {
...
+	\wp_clear_scheduled_hook( 'atmosphere_publish_comment' );
+	\wp_clear_scheduled_hook( 'atmosphere_update_comment' );
+	\wp_clear_scheduled_hook( 'atmosphere_delete_comment' );
+	\wp_clear_scheduled_hook( 'atmosphere_delete_comment_record' );
 	// Clear the legacy hook name in case an earlier PR-6 build scheduled it.
 	\wp_clear_scheduled_hook( 'atmosphere_sync_comments' );
 	\flush_rewrite_rules();
```
kraftbj, line 68: **🔴 `Client::disconnect()` does not clear any crons** (verified in `includes/oauth/class-client.php:511-513`).

Attaching here because `class-client.php` is not in this PR's diff, but it has the same cross-install risk that motivated commit `0974054` for `deactivate()` / `uninstall.php`.

Current `Client::disconnect()`:

```php
public static function disconnect(): void {
    \delete_option( 'atmosphere_connection' );
}
```

A disconnect → reconnect cycle (different account, same install) leaves every queued cron event in place. Most are safe because they re-check `is_connected()` or `should_publish_comment()`, but two are not:

- `atmosphere_delete_comment_record` (`class-atmosphere.php:835`) calls `Publisher::delete_comment_by_tid()` with no connection check. After reconnect to a new account, this issues `applyWrites#delete` against TIDs that don't belong to the new account's repo.
- `atmosphere_delete_records` (`class-atmosphere.php:753`) similarly proceeds without a connection check, and now silently swallows the resulting error (see other comment).

**Suggested fix:** have `Client::disconnect()` clear the same crons `deactivate()` does. Alternatively, gate `delete_comment_by_tid()` and `delete_post_by_tids()` on `is_connected()` — but the `Client::disconnect()` cleanup is the symmetric fix.

The commit message for `0974054` explicitly cites the deactivate→uninstall→reinstall cross-install risk; the disconnect→reconnect path has the same shape and was missed.

kraftbj (commented):
Third-pass review after `0974054` (deactivate/uninstall cron clearing + commentmeta sweep) and `612fa037` (cascade through permanent-delete path), plus the trunk merges of PR #34 (long-form teaser-thread) and PR #38 (split publish/cleanup gates).

**Static code review only — no manual testing.** Findings consolidated from the pr-review-toolkit agents and a Codex adversarial review.

## ✅ Prior must-fix / should-fix items: all resolved

- `deactivate()` clears the four new comment crons; `uninstall.php` clears them and sweeps the new commentmeta keys.
- Permanent-delete cascade now enumerates published comment TIDs in `on_before_delete()` and threads them through `delete_post_by_tids()`.
- Trunk merges (PR #34, PR #38) integrate cleanly with comment publishing (comments correctly stay outside long-form composition, and the publish/cleanup gate split is consistent with the META_URI-based comment gate).

## 🔴 New blocking findings (Codex + silent-failure-hunter)

See inline comments. Three of these are post-merge regressions; one is a Codex adversarial finding; two are teardown-completeness gaps that mirror the cleanup work in `0974054`.

## 🟡 Note (not blocking)

Codex flagged that `is_comment_eligible()` doesn't gate on the parent post's `post_status` or `is_supported_post_type()` — it only checks for stale META_URI/META_CID. See inline comment for context. Likely intentional given the publish/cleanup gate philosophy from PR #38, but worth a deliberate decision.

## Still open from prior reviews (out of scope for this pass)

Not raised inline this round — already documented in earlier reviews:
- One-shot cron events permanently dropped on transient API failures.
- Hard-delete loses durable state on transient API failure.
- `defer_when_parent_pending` counter leak on non-defer return paths.
- `on_comment_before_delete` silent drop conflates safe and orphan states.
- `store_comment_result` produces a wedged comment state on 2xx-without-uri.

kraftbj (approved):
Reviewed — no remaining concerns. LGTM.

PR #36, opened 2026-04-24
Update license to correct GPLv2-or-greater link

jeffpaul:
## Proposed changes:

This corrects the license link from the GPLv2 version to the SPDX one for GPLv2 or greater: https://spdx.org/licenses/GPL-2.0-or-later.html.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?

## Testing instructions:

* Verify the `License URI` in `atmosphere.php` and `readme.txt` matches the License header.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,25 @@
+ATmosphere - Publish WordPress posts to AT Protocol (Bluesky and standard.site)
+
+Copyright (C) 2025 Automattic and ATmosphere contributors
+
+This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


pfefferle:
@jeffpaul thanks for pointing us to the issue, but the real problem is the "-or-later", because the license file is GPL 2.0

See #43

pfefferle:
I am sorry, it seems that WordPress is also 2.0 or later! Let's also update the license file then!

kraftbj (dismissed):

PR #38, opened 2026-04-24
Split publish/cleanup gates so allowlist changes don't orphan remote records

kraftbj:
Fixes the rollback hazard where narrowing \`atmosphere_syncable_post_types\` (directly or via the \`activitypub_support_post_types\` option projected into that filter, as FOSSE's \`Post_Types\` projector does) orphans the remote AT Protocol records of posts that were already synced under the wider configuration.

## Proposed changes:

- **\`on_status_change()\`** — split the allowlist gate so it governs only the new-publish / update decisions. Unpublish now consults \`_atmosphere_bsky_tid\` / \`_atmosphere_doc_tid\` directly and schedules \`atmosphere_delete_post\` whenever either is present, regardless of whether the post type is currently in the allowlist. The \`did_action/do_action\` recursion guard moves with the scheduling block so non-syncable unpublishes without TIDs still early-return without firing it.
- **\`on_before_delete()\`** — drop the allowlist check entirely. Permanent delete is a pure cleanup path: if the post has Atmosphere publication metadata it was synced at some point, and the remote records must be removed even if the post type has since been dropped from the allowlist. A comment on the method explains the omission.

### Rationale

The bug is a gate-conflation issue. \`Backfill::syncable_post_types()\` answers "what post types is this site willing to publish to Bluesky?" — a publish-time eligibility question. Both cleanup paths were reusing it as a proxy for "was this post ever synced?", which only coincides when the allowlist never narrows. A site owner unchecking a post type on the AP side (where FOSSE's new projector surfaces that control in a way most sites already used) becomes a normal operation, not an edge case — at which point the conflation silently leaks orphaned records.

FOSSE PR Automattic/fosse#31 surfaces this path by making the allowlist editable via AP's existing checkbox UI, but the bug pre-exists in every standalone Atmosphere site that filters \`atmosphere_syncable_post_types\` in PHP. Fixing it here — not in a FOSSE-side workaround — is correct per Atmosphere's existing upstream-first positioning: post-type-agnostic correctness lives here.

### Other information:

Three new PHPUnit cases in \`tests/phpunit/tests/class-test-status-change.php\`:

1. \`test_unpublish_of_previously_synced_non_syncable_post_schedules_delete\` — publish→draft of a \`page\` with synthetic \`META_TID\` / \`META_TID\` (document) set; asserts \`atmosphere_delete_post\` is scheduled even though \`page\` isn't in the default allowlist.
2. \`test_before_delete_of_previously_synced_non_syncable_post_schedules_delete_records\` — \`on_before_delete\` on the same kind of post; asserts \`atmosphere_delete_records\` is scheduled with the captured TIDs.
3. \`test_new_publish_respects_allowlist_even_when_filter_narrows\` — regression guard: narrowing via \`atmosphere_syncable_post_types\` filter still blocks a new publish of a type the filter excludes, so the split gate doesn't accidentally loosen publish-time semantics.

Full suite runs clean (97/97).

## Testing instructions:

- Check out this branch, \`composer install\`, \`npm install\`, \`npm run env-start\`.
- Run \`npx wp-env run tests-cli --env-cwd="wp-content/plugins/wordpress-atmosphere" vendor/bin/phpunit --filter=Test_Status_Change\`. Expect 14/14.
- (Or \`npm run env-test\` for the full 97-case suite.)
- Manual: connect a site, publish a \`post\`, then add \`atmosphere_syncable_post_types\` filter returning \`[]\`, trash the post; confirm \`wp cron event list\` shows an \`atmosphere_delete_post\` scheduled.

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

-   [x] Fixed - for any bug fixes

#### Message

Preserve remote cleanup of already-synced posts when their post type is removed from the syncable allowlist.

</details>

---

Companion references:

- Downstream trigger scenario: Automattic/fosse#31 (FOSSE's \`Post_Types\` projector)
- Linear issue: https://linear.app/a8c/issue/DOTCOM-16875

### DIFF
--- a/includes/class-atmosphere.php
+++ b/includes/class-atmosphere.php
@@ -251,8 +251,36 @@ class Atmosphere {
 			return;
 		}
 
-		if ( ! \in_array( $post->post_type, Backfill::syncable_post_types(), true ) ) {
+		$is_new_publish = 'publish' === $new_status && 'publish' !== $old_status;
+		$is_update      = 'publish' === $new_status && 'publish' === $old_status;
+		$is_unpublish   = 'publish' === $old_status && 'publish' !== $new_status;
+
+		if ( ! $is_new_publish && ! $is_update && ! $is_unpublish ) {
+			// Transition between two non-publish states; nothing to schedule.
 			return;
+		}
+
+		/*
+		 * Publish-time decisions respect the allowlist so sites only sync
+		 * the post types they've opted into. Unpublish is a cleanup path
+		 * for records that were already synced, so narrowing the allowlist
+		 * later must not orphan those remote records: unpublish defers to
+		 * publication metadata (TIDs on the post) instead of the current
+		 * allowlist.
+		 */
+		if ( ( $is_new_publish || $is_update )
+			&& ! \in_array( $post->post_type, Backfill::syncable_post_types(), true )
+		) {
+			return;
+		}
+
+		if ( $is_unpublish ) {
+			$bsky_tid = \get_post_meta( $post->ID, Transformer\Post::META_TID, true );
+			$doc_tid  = \get_post_meta( $post->ID, Transformer\Document::META_TID, true );
+			if ( ! $bsky_tid && ! $doc_tid ) {
+				// Unpublish of a post that was never synced — nothing to clean up.
+				return;
+			}
 		}
 
 		// Prevent infinite loops from meta updates.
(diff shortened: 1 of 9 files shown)


includes/class-atmosphere.php
```diff
@@ -251,35 +251,57 @@ public function on_status_change( string $new_status, string $old_status, \WP_Po
 			return;
 		}
 
-		if ( ! \in_array( $post->post_type, Backfill::syncable_post_types(), true ) ) {
+		$is_new_publish = 'publish' === $new_status && 'publish' !== $old_status;
+		$is_update      = 'publish' === $new_status && 'publish' === $old_status;
+		$is_unpublish   = 'publish' === $old_status && 'publish' !== $new_status;
```
pfefferle, line 256: Do we want to delete posts that were put into draft mode!? Is it possible to re-publish posts in Bluesky? Mastodon is very picky with that... once deleted, it can't be published (with the same ID) again.

kraftbj: Those are good questions. I would think that if someone made a post a draft, it should be deleted. But, the how should we act on re-publish is fair. Let me think about it a bit and checkout the AT docs a bit.

kraftbj: Dug in. Good news on the AT Protocol vs. Mastodon front: rkeys can be reused after deletion, so a re-publish gets a fresh TID and publishes cleanly — the existing meta cleanup on successful delete already handles that path.

The real footgun isn't republish failure, it's engagement loss. Likes, reposts, replies, and quote posts all attach to a specific record's AT-URI/CID. Delete and republish, and that engagement is pointed at a tombstone. Someone hitting "Switch to Draft" to fix a typo loses every interaction.

That's a real UX problem, but it's orthogonal to this PR — the draft → delete behavior pre-dates this branch, trunk does the same thing today. All this PR changes is *when* cleanup is allowed to run (decoupling it from the allowlist).

My take: keep this PR scoped to the allowlist bug, and I'll open a separate issue to think through draft-vs-trash semantics properly.

kraftbj: Filed as #39 for the broader discussion.

pfefferle (approved):

PR #41, opened 2026-04-28
Choose which post types publish to AT Protocol

pfefferle:
## Proposed changes

- New **Post types** field in *Settings → ATmosphere* with a checkbox per public post type. Stored as `atmosphere_support_post_types` (default `['post']`).
- Plugins and themes can opt their custom post types in via the WordPress-canonical `\add_post_type_support( 'book', 'atmosphere' )` — picked up automatically.
- Replaces `Backfill::syncable_post_types()` with `Atmosphere\Post_Types` + bridge functions `Atmosphere\get_supported_post_types()` and `Atmosphere\is_supported_post_type()`.
- `atmosphere_syncable_post_types` filter preserved; now runs after the option and native opt-ins are merged so callers can add or remove from either source.

### Other information

## Testing instructions

1. Visit *Settings → ATmosphere*. Confirm a new **Post types** field appears in the Publishing section with a checkbox per public post type. **Posts** is checked by default.
2. Check **Pages**, save, edit any page. The ATmosphere meta box should appear.
3. Uncheck everything, save, and confirm no meta box renders on any post-edit screen.
4. In a small test plugin, add `add_post_type_support( 'book', 'atmosphere' );` for a custom `book` CPT — confirm `Atmosphere\get_supported_post_types()` includes it even when the option is empty.
5. `composer lint` and `npm run env-test` should pass.

### DIFF
--- a/includes/class-atmosphere.php
+++ b/includes/class-atmosphere.php
@@ -98,7 +98,7 @@ class Atmosphere {
 			return;
 		}
 
-		if ( ! \in_array( $post->post_type, Backfill::syncable_post_types(), true ) ) {
+		if ( ! is_supported_post_type( $post->post_type ) ) {
 			return;
 		}
 
@@ -220,7 +220,7 @@ class Atmosphere {
 			return;
 		}
 
-		if ( ! \in_array( $post->post_type, Backfill::syncable_post_types(), true ) ) {
+		if ( ! is_supported_post_type( $post->post_type ) ) {
 			\status_header( 404 );
 			exit;
 		}
@@ -261,16 +261,14 @@ class Atmosphere {
 		}
 
 		/*
-		 * Publish-time decisions respect the allowlist so sites only sync
-		 * the post types they've opted into. Unpublish is a cleanup path
-		 * for records that were already synced, so narrowing the allowlist
-		 * later must not orphan those remote records: unpublish defers to
-		 * publication metadata (TIDs on the post) instead of the current
-		 * allowlist.
+		 * Publish-time decisions respect the supported list so sites
+		 * only sync the post types they've opted into. Unpublish is a
+		 * cleanup path for records that were already synced, so
+		 * narrowing support later must not orphan those remote records:
+		 * unpublish defers to publication metadata (TIDs on the post)
+		 * instead of the current support list.
 		 */
-		if ( ( $is_new_publish || $is_update )
-			&& ! \in_array( $post->post_type, Backfill::syncable_post_types(), true )
-		) {
+		if ( ( $is_new_publish || $is_update ) && ! is_supported_post_type( $post->post_type ) ) {
 			return;
 		}
 
@@ -325,12 +323,12 @@ class Atmosphere {
 		}
 
 		/*
-		 * No allowlist check here. Permanent delete is a cleanup path: if
+		 * No support check here. Permanent delete is a cleanup path: if
 		 * the post has Atmosphere publication metadata it was synced at
 		 * some point, and the remote records must be removed even if the
-		 * post type has since been dropped from the syncable allowlist.
-		 * Gating this on the current allowlist would orphan already-
-		 * published records whenever a site narrows its configuration.
+		 * post type has since been removed from the supported list.
+		 * Gating this on current support would orphan already-published
+		 * records whenever a site narrows its configuration.
 		 */
 		$bsky_tid = \get_post_meta( $post_id, Transformer\Post::META_TID, true );
 		$doc_tid  = \get_post_meta( $post_id, Transformer\Document::META_TID, true );
--- a/includes/class-atmosphere.php
+++ b/includes/class-atmosphere.php
@@ -366,11 +366,19 @@ class Atmosphere {
 	 * Register async action hooks (called by WP-Cron).
 	 */
 	public static function register_async_hooks(): void {
+		/*
+		 * Publish/update cron callbacks re-check post-type support.
+		 * A user (or downstream filter) can disable a post type after a
+		 * cron event was queued, and we must not still publish it.
+		 *
+		 * The delete callback intentionally skips this check so cleanup
+		 * still runs after support is removed.
+		 */
 		\add_action(
 			'atmosphere_publish_post',
 			static function ( int $post_id ): void {
 				$post = \get_post( $post_id );
-				if ( $post && 'publish' === $post->post_status ) {
+				if ( $post && 'publish' === $post->post_status && is_supported_post_type( $post->post_type ) ) {
 					Publisher::publish( $post );
 				}
 			}
@@ -380,7 +388,7 @@ class Atmosphere {
 			'atmosphere_update_post',
 			static function ( int $post_id ): void {
 				$post = \get_post( $post_id );
-				if ( $post && 'publish' === $post->post_status ) {
+				if ( $post && 'publish' === $post->post_status && is_supported_post_type( $post->post_type ) ) {
 					Publisher::update( $post );
 				}
 			}
--- a/includes/class-backfill.php
+++ b/includes/class-backfill.php
@@ -113,12 +113,14 @@ class Backfill {
 			\wp_send_json_error( 'No post IDs provided.' );
 		}
 
-		$results = array();
+		// Resolve supported post types once for the whole batch.
+		$supported = get_supported_post_types();
+		$results   = array();
 
 		foreach ( $post_ids as $post_id ) {
 			$post = \get_post( $post_id );
 
-			if ( ! $post || 'publish' !== $post->post_status || ! is_supported_post_type( $post->post_type ) ) {
+			if ( ! $post || 'publish' !== $post->post_status || ! \in_array( $post->post_type, $supported, true ) ) {
 				$results[] = array(
 					'id'      => $post_id,
 					'success' => false,
--- a/includes/class-backfill.php
+++ b/includes/class-backfill.php
@@ -39,6 +39,20 @@ class Backfill {
 
 		$post_types = get_supported_post_types();
 
+		/*
+		 * Short-circuit when no post types are enabled. Passing an empty
+		 * array to get_posts() falls back to the default `post` query,
+		 * which would surface posts that nothing is configured to sync.
+		 */
+		if ( empty( $post_types ) ) {
+			\wp_send_json_success(
+				array(
+					'total'    => 0,
+					'post_ids' => array(),
+				)
+			);
+		}
+
 		/**
 		 * Filters the maximum number of posts to backfill.
 		 *
@@ -104,7 +118,7 @@ class Backfill {
 		foreach ( $post_ids as $post_id ) {
 			$post = \get_post( $post_id );
 
-			if ( ! $post || 'publish' !== $post->post_status ) {
+			if ( ! $post || 'publish' !== $post->post_status || ! is_supported_post_type( $post->post_type ) ) {
 				$results[] = array(
 					'id'      => $post_id,
 					'success' => false,
--- a/includes/class-backfill.php
+++ b/includes/class-backfill.php
@@ -37,7 +37,7 @@ class Backfill {
 
 		\check_ajax_referer( 'atmosphere_backfill', 'nonce' );
 
-		$post_types = self::syncable_post_types();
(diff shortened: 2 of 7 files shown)


includes/class-atmosphere.php
```diff
@@ -251,7 +251,7 @@ public function on_status_change( string $new_status, string $old_status, \WP_Po
 			return;
 		}
 
-		if ( ! \in_array( $post->post_type, Backfill::syncable_post_types(), true ) ) {
+		if ( ! is_supported_post_type( $post->post_type ) ) {
 			return;
```
kraftbj: [P1] Cleanup must not use current support

`is_supported_post_type()` is now user-mutable through the new settings and through FOSSE's projection from `activitypub_support_post_types`. If a post was already published and the user later removes its post type, this early return prevents publish-to-draft/trash cleanup from scheduling deletes; the same pattern in `on_before_delete()` skips permanent-delete cleanup. This matches the orphan-record hazard documented in FOSSE's `sdd/post-type-sync/notes.md`; pull in the split publish/cleanup gate from `fix/cleanup-gate-split`/PR #38 so cleanup keys off existing TIDs rather than current publish eligibility.

kraftbj: [P2] Cron publish ignores later opt-out

The scheduling path checks post-type support, but these async handlers publish/update any still-published post when cron fires. With the new settings UI, and especially with FOSSE writing AP post-type settings directly, a user can disable a post type after an event is queued and still have that content sent to Bluesky. Re-check `is_supported_post_type()` in the publish/update cron callbacks, while keeping delete callbacks metadata-driven for cleanup.

includes/wp-admin/class-admin.php
```diff
@@ -247,6 +273,57 @@ public static function render_auto_publish_field(): void {
...
+		?>
+		<fieldset>
+			<legend class="screen-reader-text">
+				<?php \esc_html_e( 'Post types', 'atmosphere' ); ?>
+			</legend>
+			<?php
+			foreach ( $post_types as $post_type ) :
+				$is_saved    = \in_array( $post_type->name, $saved, true );
+				$is_external = ! $is_saved && \in_array( $post_type->name, $effective, true );
+				?>
```
kraftbj: [P3] Settings can show filtered-out types as enabled

The UI marks checkboxes from the saved Atmosphere option only, but the effective list can be replaced by `atmosphere_syncable_post_types`. Under FOSSE, AP's option is the source of truth, so an Atmosphere-saved/default `post` can render checked even when FOSSE's filter returns only `page`. That makes the settings page claim a type publishes when runtime will skip it. Compare `$saved` with `$effective` both ways and show saved-but-disabled types as externally overridden, or disable this field when a filter owns the list.

includes/class-atmosphere.php
kraftbj: [P1] Cleanup must bypass current allowlist

This support check now reflects mutable settings and FOSSE's AP-option projection. If a post was already published and the user later removes its post type, publish-to-draft/trash and permanent delete skip remote cleanup, leaving AT records orphaned. Pull in the split publish/cleanup gate from PR #38: publish/update should use the allowlist, cleanup should key off existing Atmosphere TIDs.

kraftbj: [P2] Cron publish should re-check support

The schedule path checks post-type support, but these async handlers publish or update any still-published post when cron fires. With a settings UI, a user can disable a post type after an event is queued and still have it sent to Bluesky. Re-check `is_supported_post_type()` in publish/update cron callbacks; keep delete callbacks metadata-driven for cleanup.

includes/wp-admin/class-admin.php
kraftbj: [P3] UI can show filtered-out saved types as active

The field now distinguishes saved types from externally enabled types, but it still does not show the inverse case: saved-but-filtered-out. FOSSE replaces the effective list from AP's option, so `post` can remain checked from Atmosphere's saved/default option while runtime support is only `page`. Add a saved-but-disabled state or make the field clearly read-only/overridden when a filter owns the list.

includes/class-post-types.php
```diff
@@ -0,0 +1,78 @@
...
+	 */
+	public static function sanitize( $value ): array {
+		if ( empty( $value ) ) {
+			return array();
+		}
+
+		$allowed = \get_post_types( array( 'public' => true ) );
+		$value   = \array_map( 'sanitize_text_field', (array) $value );
+
+		return \array_values( \array_intersect( $value, $allowed ) );
```
kraftbj: [P3] Normalize saved post type option

`sanitize()` still uses `sanitize_text_field()` and returns `array_intersect()` without dropping duplicates or normalizing to the same slug format used by `get_supported()`. Use `sanitize_key()`, filter non-strings/empties, and `array_unique()` before storing so the option remains a canonical string list.

kraftbj (commented):
Combined review comments from the full PR pass, including the FOSSE-specific concerns and the remaining existing review concerns.

includes/class-backfill.php
```diff
@@ -104,7 +118,7 @@ public static function handle_batch(): void {
 		foreach ( $post_ids as $post_id ) {
 			$post = \get_post( $post_id );
 
-			if ( ! $post || 'publish' !== $post->post_status ) {
+			if ( ! $post || 'publish' !== $post->post_status || ! is_supported_post_type( $post->post_type ) ) {
 				$results[] = array(
```
kraftbj, line 124: [P3] Avoid resolving support once per backfill ID

`is_supported_post_type()` recomputes the option, native support list, filter result, and normalization for each submitted ID. Backfill batches can include many IDs, so resolve `get_supported_post_types()` once before the loop and use `in_array()` against that local list.

pfefferle:
@kraftbj it was not meant to fix what you fixed with your PR already, this was simply meant to extract the functionality to a dedicated class/function instead of a weird artifact that lives in the Backfill structure (which I am not even sure we need). I am happy to point the PR to your branch instead or directly work on your branch.

pfefferle:
I rebased it to your pr.

includes/wp-admin/class-admin.php
```diff
@@ -247,6 +273,38 @@ public static function render_auto_publish_field(): void {
...
+					<label>
+						<input
+							type="checkbox"
+							name="atmosphere_support_post_types[]"
+							value="<?php echo \esc_attr( $post_type->name ); ?>"
+							<?php \checked( \in_array( $post_type->name, $enabled, true ) ); ?>
+						>
+						<?php echo \esc_html( $post_type->label ); ?>
+						<code><?php echo \esc_html( $post_type->name ); ?></code>
+					</label>
```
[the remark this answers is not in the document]
pfefferle: Addressed in 1132250 — checkboxes now read from the saved option; effective-only types render with an "Enabled by another plugin or theme" note. 0685c47 extends this with the saved-but-filtered-out case.

includes/class-post-types.php
```diff
@@ -0,0 +1,71 @@
...
+	 * in via `\add_post_type_support( $post_type, 'atmosphere' )`.
+	 *
+	 * @return string[]
+	 */
+	public static function get_supported(): array {
+		$configured = (array) \get_option( 'atmosphere_support_post_types', array( 'post' ) );
+		$native     = \get_post_types_by_support( 'atmosphere' );
+
+		$post_types = \array_values( \array_unique( \array_merge( $configured, $native ) ) );
+
```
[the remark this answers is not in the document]
pfefferle: Filter-output normalisation (`sanitize_key` + `array_filter` + `array_unique`) was added in 1132250. Intentionally not intersecting against public types here: native `add_post_type_support()` and the `atmosphere_syncable_post_types` filter are valid extension points for non-public types. The public-only filter is only on the option `sanitize()` path.

includes/class-backfill.php
```diff
@@ -37,7 +37,7 @@ public static function handle_count(): void {
 
 		\check_ajax_referer( 'atmosphere_backfill', 'nonce' );
 
-		$post_types = self::syncable_post_types();
+		$post_types = get_supported_post_types();
```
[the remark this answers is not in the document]
pfefferle: Addressed in 1132250 — `handle_count()` short-circuits to 0 when the supported list is empty; `handle_batch()` re-checks support per ID.

includes/class-post-types.php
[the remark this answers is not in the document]
pfefferle: Fixed in 0685c47 — `sanitize_key`, drop non-strings/empties, `array_unique`.

includes/wp-admin/class-admin.php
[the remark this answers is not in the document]
pfefferle: Fixed in 0685c47 — added an `is_filtered_out` state with "Disabled by another plugin or theme — this post type will not be published."

includes/class-backfill.php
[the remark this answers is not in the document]
pfefferle: Fixed in 0685c47 — `handle_batch()` resolves `get_supported_post_types()` once before the loop and uses `in_array` against the local list.

includes/class-atmosphere.php
pfefferle: Done — rebased onto #38 (`fix/cleanup-gate-split`) and pointed PR #41 base at it. Cleanup paths now bypass current support.

pfefferle: Done — rebased onto #38 (`fix/cleanup-gate-split`) and pointed PR #41 base at it. Cleanup paths now bypass current support.

pfefferle: Fixed in 0685c47 — `atmosphere_publish_post` and `atmosphere_update_post` callbacks re-check `is_supported_post_type()`. Delete callback intentionally skips the check so cleanup still runs.

pfefferle: Fixed in 0685c47 — `atmosphere_publish_post` and `atmosphere_update_post` callbacks re-check `is_supported_post_type()`. Delete callback intentionally skips the check so cleanup still runs.

includes/wp-admin/class-admin.php
pfefferle: Fixed in 0685c47 — saved-but-filtered-out types now render "Disabled by another plugin or theme — this post type will not be published."

pfefferle: Fixed in 0685c47 — saved-but-filtered-out types now render "Disabled by another plugin or theme — this post type will not be published."

includes/class-post-types.php
pfefferle: Fixed in 0685c47 — `sanitize_key`, drop non-strings/empties, `array_unique`.

includes/class-backfill.php
pfefferle: Fixed in 0685c47 — resolved once before the loop, `in_array` per ID.

kraftbj:
Verified `0685c47` against all eight unresolved review threads and resolved them. Summary of what I checked against the code at that SHA:

- **P1 cleanup must bypass current allowlist** (both threads): `on_before_delete()` no longer gates on `is_supported_post_type()` and only checks for existing TIDs; `on_status_change()` restricts the support check to the publish/update branches and lets unpublish defer to TID metadata. PR 41 is rebased onto `fix/cleanup-gate-split` (merge-base matches that branch's head), so this depends on PR 38 landing first.
- **P2 cron publish should re-check support** (both threads): `register_async_hooks()` re-checks `is_supported_post_type( $post->post_type )` inside both the `atmosphere_publish_post` and `atmosphere_update_post` callbacks. The `atmosphere_delete_post` callback intentionally omits the check so cleanup keeps running.
- **P3 settings UI shows filtered-out types as enabled** (both threads): `render_support_post_types_field()` now distinguishes `$is_external` and `$is_filtered_out` and renders "Disabled by another plugin or theme — this post type will not be published." for the saved-but-filtered case.
- **P3 normalize saved post type option**: `Post_Types::sanitize()` drops non-strings, runs `sanitize_key`, filters empties, and dedupes via `array_unique` before intersecting against public post types.
- **P3 avoid resolving support once per backfill ID**: `Backfill::handle_batch()` resolves `get_supported_post_types()` once before the loop and uses `in_array( $post->post_type, $supported, true )` per ID.

All eight threads resolved; ready for formal approval.

kraftbj (approved):

PR #42, opened 2026-04-28
Settings UI for the long-form composition strategy

pfefferle:
Builds on top of #34. Targeting that branch so the setting is testable end-to-end alongside the strategies. **Re-target to trunk once #34 lands.**

## Proposed changes

- New **Long-form posts** field in *Settings → ATmosphere*, Publishing section. Radio group with three options matching the strategies #34 introduces:
  - **Link card** (default) — unchanged from today.
  - **Truncated post with link** — body text + inline permalink, no card.
  - **Teaser thread** — two-post hook + CTA-with-link.
- Stored as `atmosphere_long_form_composition`. REST schema enumerates the three valid values; sanitize callback falls back to the default for unknown / non-string input.
- `Atmosphere::init()` registers a low-priority (priority 1) filter that seeds `atmosphere_long_form_composition` from the option, leaving the default-priority filter slot open for downstream per-post overrides (FOSSE, custom plugins, etc.).

### Other information

## Testing instructions

1. After #34 is merged into this branch, visit *Settings → ATmosphere*. The new **Long-form posts** field shows three radio options. **Link card** is selected by default.
2. Switch to **Teaser thread**, save, and publish a long-form post (500+ words, has a title). Confirm two `app.bsky.feed.post` records land with `reply` refs.
3. Add `add_filter( 'atmosphere_long_form_composition', fn() => 'truncate-link' );` in a small test plugin while the setting is **Teaser thread**. Confirm the filter wins.
4. Manually corrupt the option (`update_option( 'atmosphere_long_form_composition', 'bogus' )`). Confirm publishing falls back to **link-card**.
5. `composer lint` and `npm run env-test` pass (5 new tests, 143 total).

### DIFF
--- a/includes/class-atmosphere.php
+++ b/includes/class-atmosphere.php
@@ -34,6 +34,23 @@ class Atmosphere {
 		\add_action( 'init', array( Admin::class, 'register' ), 5 );
 		\add_action( 'init', array( Backfill::class, 'register' ), 5 );
 
+		/*
+		 * Seed the long-form composition strategy from the user's
+		 * setting. Priority 1 so any downstream filter at the default
+		 * priority can still override it per post.
+		 */
+		\add_filter(
+			'atmosphere_long_form_composition',
+			static function ( string $strategy ): string {
+				$option = (string) \get_option( 'atmosphere_long_form_composition', 'link-card' );
+
+				return \in_array( $option, array( 'link-card', 'truncate-link', 'teaser-thread' ), true )
+					? $option
+					: $strategy;
+			},
+			1
+		);
+
 		// REST route (always active for client-metadata).
 		\add_action( 'rest_api_init', array( Admin::class, 'register_rest_routes' ) );
 
--- a/includes/class-atmosphere.php
+++ b/includes/class-atmosphere.php
@@ -22,6 +22,12 @@ use Atmosphere\WP_Admin\Admin;
 class Atmosphere {
 
 	/**
+	 * Allowed values for the long-form composition strategy filter and
+	 * the matching `atmosphere_long_form_composition` option.
+	 */
+	public const LONG_FORM_STRATEGIES = array( 'link-card', 'truncate-link', 'teaser-thread' );
+
+	/**
 	 * Wire up all hooks.
 	 */
 	public function init(): void {
@@ -39,17 +45,7 @@ class Atmosphere {
 		 * setting. Priority 1 so any downstream filter at the default
 		 * priority can still override it per post.
 		 */
-		\add_filter(
-			'atmosphere_long_form_composition',
-			static function ( string $strategy ): string {
-				$option = (string) \get_option( 'atmosphere_long_form_composition', 'link-card' );
-
-				return \in_array( $option, array( 'link-card', 'truncate-link', 'teaser-thread' ), true )
-					? $option
-					: $strategy;
-			},
-			1
-		);
+		\add_filter( 'atmosphere_long_form_composition', array( self::class, 'seed_long_form_composition' ), 1 );
 
 		// REST route (always active for client-metadata).
 		\add_action( 'rest_api_init', array( Admin::class, 'register_rest_routes' ) );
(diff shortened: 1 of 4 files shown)


tests/phpunit/tests/class-test-long-form-composition-setting.php
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Long-form composition setting tests.
+ */
+class Test_Long_Form_Composition_Setting extends WP_UnitTestCase {
+
+	/**
+	 * Reset state between tests.
+	 */
+	public function tear_down(): void {
```
kraftbj, line 42: Heads-up on a test-isolation fragility: the seed filter is registered once via `Atmosphere::init()` on `plugins_loaded`. `class-test-publisher.php` (line 58) and `transformer/class-test-post.php` (lines 27, 750) call `\remove_all_filters( 'atmosphere_long_form_composition' )` in their `tear_down`, and PHPUnit doesn't re-fire `plugins_loaded` between classes. Once either of those classes runs first, the seed callback is gone for the rest of the suite.

Today the suite stays green by alphabetical accident (this class sorts first). If the order ever shifts:

- `test_option_seeds_filter` would fail.
- `test_corrupt_option_falls_through_to_default` keeps passing for the wrong reason — with no seed filter on the chain, `apply_filters` returns the unchanged seed argument, which happens to equal `link-card`.

Easiest fix: re-register the seed in this class's `set_up()` and add `\remove_all_filters( 'atmosphere_long_form_composition' )` to its own `tear_down`. Even cleaner: extract the seed closure to a static method on `Atmosphere` so production and tests share one definition.

Non-blocking.

includes/class-atmosphere.php
```diff
@@ -34,6 +34,23 @@ public function init(): void {
...
+		 * priority can still override it per post.
+		 */
+		\add_filter(
+			'atmosphere_long_form_composition',
+			static function ( string $strategy ): string {
+				$option = (string) \get_option( 'atmosphere_long_form_composition', 'link-card' );
+
+				return \in_array( $option, array( 'link-card', 'truncate-link', 'teaser-thread' ), true )
+					? $option
+					: $strategy;
```
kraftbj: Worth aligning with the existing prior art at `includes/transformer/class-post.php` line 407, which `error_log()`s every silent strategy downgrade specifically so operators can tell a fallback from a configuration choice.

Today, if the stored option is corrupt (third-party write, or schema drift between releases), the seed silently returns `$strategy` and every subsequent publish quietly uses the wrong strategy with no breadcrumb. A rate-limited `error_log()` (transient-gated to once per hour) would mirror the convention without log spam.

Non-blocking.

includes/wp-admin/class-admin.php
```diff
@@ -68,6 +68,22 @@ public static function register_settings(): void {
...
+			'atmosphere',
+			'atmosphere_long_form_composition',
+			array(
+				'type'              => 'string',
+				'description'       => 'Composition strategy for long-form Bluesky posts.',
+				'default'           => 'link-card',
+				'sanitize_callback' => array( self::class, 'sanitize_long_form_composition' ),
+				'show_in_rest'      => array(
+					'schema' => array(
+						'enum' => array( 'link-card', 'truncate-link', 'teaser-thread' ),
```
kraftbj: The literal `array( 'link-card', 'truncate-link', 'teaser-thread' )` now appears in four places: this REST `enum`, the seed filter (`class-atmosphere.php` line 47), the sanitize allowlist (`class-admin.php` line 328), and the radio render `$choices` keys (`class-admin.php` lines 280, 284, 288). They all agree today, but adding a fourth strategy is one missed copy-paste away from drift.

A single shared constant (e.g. `Admin::LONG_FORM_STRATEGIES`) referenced from all four sites would tighten this up.

Non-blocking.

kraftbj (approved):
Reviewed. None of these strike me as blockers — leaving them as suggestions for a follow-up.

kraftbj:
Thanks! Looks good from here.

PR #43, opened 2026-04-28
Align License header with the GPL-2.0 URI

pfefferle:
## Proposed changes:

The `License URI` in both `atmosphere.php` and `readme.txt` points to https://www.gnu.org/licenses/gpl-2.0.html — the canonical GPL v2 license text. Update the `License` declaration from `GPL-2.0-or-later` to `GPL-2.0` so the SPDX identifier matches the URI it links to.

Replaces #36 (which addressed the same mismatch by updating the URI to SPDX's "or later" page; this PR aligns the declaration with the URI instead).

### Other information:

- [x] Skip Changelog (metadata fix only).

## Testing instructions:

* Verify `License: GPL-2.0` appears in both `atmosphere.php` (header docblock) and `readme.txt`.
* Verify `License URI` continues to point to https://www.gnu.org/licenses/gpl-2.0.html in both files.

### DIFF
--- a/atmosphere.php
+++ b/atmosphere.php
@@ -6,7 +6,7 @@
  * Version: unreleased
  * Author: Automattic
  * Author URI: https://automattic.com
- * License: GPL-2.0-or-later
+ * License: GPL-2.0
  * License URI: https://www.gnu.org/licenses/gpl-2.0.html
  * Text Domain: atmosphere
  * Requires PHP: 8.2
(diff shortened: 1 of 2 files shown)


kraftbj:
Is the intention to stick to the GPL 2 and disallow GPL 3 usage? WP itself is 2.0-or-later.

pfefferle:
Oh is it? I thought core is strict 2.0

PR #55, opened 2026-05-08
Use site domain as Bluesky handle

pfefferle:
Fixes #

## Proposed changes:

* Add a new `Atmosphere\Handle` service that calls `com.atproto.identity.updateHandle` to set the site's hostname as the connected Bluesky handle. Snapshots the previous handle to `atmosphere_previous_handle` so disconnect can revert.
* Render a confirm panel inside the connected-state Settings section. Only offered when the site is at the host root, the user is connected, and the current handle differs from the site host. Hidden via the `atmosphere_domain_handle_enabled` filter.
* Wire `admin_post_atmosphere_set_domain_handle` to verify capability + nonce, dispatch `Handle::set_handle()`, and redirect back to the settings page with a notice already populated by `Handle`.
* Best-effort revert in the disconnect handler: call `Handle::maybe_revert_on_disconnect()` BEFORE `Client::disconnect()` so the OAuth token is still valid for the XRPC call. Clear the snapshot afterward so it cannot be revived by a future reconnect to a different account.
* Bind snapshot lifetime to the connection: also delete `atmosphere_previous_handle` in `uninstall.php`.

Ports the domain-handle service from [FOSSE PR #97](https://github.com/Automattic/fosse/pull/97). Builds on the `identity:handle` OAuth scope added in #53.

### Other information:

23 PHPUnit tests in `tests/phpunit/tests/class-test-handle.php` cover eligibility gating, the short-circuit filter (`atmosphere_pre_update_handle`), success / WP_Error propagation, and the revert flow. The XRPC layer is exercised through the short-circuit filter so tests do not require DPoP keys.

## Testing instructions:

1. On a site at host root (not a subdirectory install), connect a Bluesky account whose handle is something like `alice.bsky.social`.
2. Visit Settings → ATmosphere. Confirm the new "Use your domain as your Bluesky handle" panel renders with the destructive-warning copy and a button labelled "Use \<site host> as my Bluesky handle".
3. Click the button. The XRPC call goes to your PDS; on success a notice "Your Bluesky handle is now \<host>." appears.
4. Verify on bsky.app or via `com.atproto.identity.resolveHandle` that the handle changed.
5. Click Disconnect. Confirm the previous handle is restored ("Restored your previous Bluesky handle: alice.bsky.social.") before the disconnect notice.
6. On a subdirectory install (`example.com/blog`), confirm the panel does NOT render.
7. With `add_filter( 'atmosphere_domain_handle_enabled', '__return_false' )`, confirm the panel does NOT render and disconnect does NOT attempt revert.

### Changelog entry

<details>

#### Significance

-   [x] Minor

#### Type

#### Message

Use your site domain as your Bluesky handle with one click from the ATmosphere settings page.

</details>

### DIFF
--- a/includes/class-handle.php
+++ b/includes/class-handle.php
@@ -83,4 +83,51 @@ class Handle {
 
 		return null === $path || '' === $path || '/' === $path;
 	}
+
+	/**
+	 * Compute the handle that the site would advertise.
+	 *
+	 * Reads the host portion of `home_url()`. Returns empty when the host
+	 * cannot be resolved so callers refuse to send an empty payload.
+	 *
+	 * @return string
+	 */
+	public static function get_target_handle(): string {
+		$host = \wp_parse_url( \home_url(), PHP_URL_HOST );
+
+		return \is_string( $host ) ? \strtolower( $host ) : '';
+	}
+
+	/**
+	 * Whether the confirm-handle UI should render for the given status.
+	 *
+	 * @param array<string, mixed> $status Connection status snapshot with at
+	 *                                     least `connected` and `handle`.
+	 * @return bool
+	 */
+	public static function should_offer( array $status ): bool {
+		if ( ! self::is_enabled() ) {
+			return false;
+		}
+
+		if ( ! self::is_root_install() ) {
+			return false;
+		}
+
+		$target = self::get_target_handle();
+		if ( '' === $target ) {
+			return false;
+		}
+
+		if ( empty( $status['connected'] ) ) {
+			return false;
+		}
+
+		$current = isset( $status['handle'] ) ? (string) $status['handle'] : '';
+		if ( '' !== $current && \strtolower( $current ) === $target ) {
+			return false;
+		}
+
+		return true;
+	}
 }
--- a/includes/class-handle.php
+++ b/includes/class-handle.php
@@ -76,12 +76,39 @@ class Handle {
 	 * feature must skip them — Bluesky's PDS would reject the handle change
 	 * because it cannot fetch `/.well-known/atproto-did` at the host root.
 	 *
+	 * Also rejects split installs where `home_url` is at the host root but
+	 * `site_url` lives in a subdirectory (e.g. `WP_HOME=https://example.com`,
+	 * `WP_SITEURL=https://example.com/wp`). The `.well-known/atproto-did`
+	 * rewrite resolves through WordPress's request parser rooted at
+	 * `site_url`, so the verification fetch would 404 even though `home_url`
+	 * looks correct — and the PDS would have already accepted the new
+	 * handle by the time we found out.
+	 *
 	 * @return bool
 	 */
 	public static function is_root_install(): bool {
-		$path = \wp_parse_url( \home_url(), PHP_URL_PATH );
+		$home = \wp_parse_url( \home_url() );
+		$site = \wp_parse_url( \site_url() );
 
-		return null === $path || '' === $path || '/' === $path;
+		if ( ! \is_array( $home ) || ! \is_array( $site ) ) {
+			return false;
+		}
+
+		$home_path = isset( $home['path'] ) ? (string) $home['path'] : '';
+		$site_path = isset( $site['path'] ) ? (string) $site['path'] : '';
+
+		if ( '' !== $home_path && '/' !== $home_path ) {
+			return false;
+		}
+
+		if ( '' !== $site_path && '/' !== $site_path ) {
+			return false;
+		}
+
+		$home_host = isset( $home['host'] ) ? \strtolower( (string) $home['host'] ) : '';
+		$site_host = isset( $site['host'] ) ? \strtolower( (string) $site['host'] ) : '';
+
+		return '' !== $home_host && $home_host === $site_host;
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


pfefferle:
Addressed both Copilot review comments in 35cd3a7:

- After a successful `com.atproto.identity.updateHandle`, both `Handle::set_handle()` and `Handle::maybe_revert_on_disconnect()` now mirror the new value into `atmosphere_connection['handle']` via a new private `sync_connection_handle()` helper. The Settings page reflects the new handle immediately and `should_offer()` correctly stops returning true once `current === target`.
- Tests extended:
  - Success path asserts `atmosphere_connection['handle'] === 'example.com'` after `set_handle()`.
  - Failure path asserts the connection handle is unchanged on `WP_Error`.
  - Revert success asserts the connection mirrors back to the previous handle.

Also restructured the panel as an `add_settings_field` on the existing `atmosphere_connection` section instead of inlining it — the previous nested `<form>` inside the settings page's outer `<form action="options.php">` was invalid HTML. The button is now a GET-nonce link styled as a button, matching the existing Disconnect pattern.

includes/class-handle.php
```diff
@@ -0,0 +1,350 @@
...
+	 * Whether the site is at the root of its domain.
+	 *
+	 * Subdirectory installs (e.g. `https://example.com/blog/`) cannot serve
+	 * the AT Protocol verification endpoint at the domain root, so the
+	 * feature must skip them — Bluesky's PDS would reject the handle change
+	 * because it cannot fetch `/.well-known/atproto-did` at the host root.
+	 *
+	 * @return bool
+	 */
+	public static function is_root_install(): bool {
```
kraftbj, line 89: Worth confirming end-to-end on a `home_url at host root, site_url in subdirectory` install (e.g. `WP_HOME=https://example.com`, `WP_SITEURL=https://example.com/wp`).

`is_root_install` here only checks `home_url`'s path, but the `^\.well-known/atproto-did$` rewrite rule registered by `Atmosphere` resolves through WordPress's request parser, which is rooted at `site_url`. On that layout the rewrite may not catch `/.well-known/atproto-did` at the host root, Bluesky's verification fetch 404s, and the user's handle change ends up invalid downstream — but the panel will have happily offered the button and `updateHandle` succeeded against the PDS first.

If that layout is broken, this is the place to catch it (compare `home_url` and `site_url` hosts/paths) so we don't write to the PDS for a configuration we can't actually verify. If it works, expand the testing-instructions step 6 to explicitly cover that case so it doesn't regress.

includes/class-handle.php
```diff
@@ -0,0 +1,350 @@
...
+	public static function get_target_handle(): string {
+		$host = \wp_parse_url( \home_url(), PHP_URL_HOST );
+
+		return \is_string( $host ) ? \strtolower( $host ) : '';
+	}
+
+	/**
+	 * Replace the connected user's Bluesky handle with the site host.
+	 *
+	 * Caller MUST have verified capability + nonce before invoking. The
```
kraftbj: Fragile contract on a public static method: "Caller MUST have verified capability + nonce." Today the only caller is `Admin::handle_set_domain_handle` which does both, so this is fine — but anything else that ever wires this up (a future cron job, REST endpoint, WP-CLI command) inherits the obligation by docblock alone.

If you want to keep the public surface thin, consider either renaming to `set_handle_unsafe()` to make the obligation visible at the call site, or having `set_handle()` itself short-circuit when `! is_admin() || ! current_user_can( 'manage_options' )` (it would still need a separate entry point for non-admin contexts later, but at least the default is fail-safe). Not blocking — happy to keep as-is if there's no plan to add other callers.

includes/class-handle.php
```diff
@@ -0,0 +1,350 @@
...
+	 * Mirror a successful PDS handle change into the local connection option.
+	 *
+	 * The PDS is now serving the new handle, but `atmosphere_connection`
+	 * still holds the value captured at OAuth time. Without this update
+	 * the Settings page shows the stale handle and {@see self::should_offer()}
+	 * keeps offering the panel because `current !== target`.
+	 *
+	 * @param string $handle New handle to record locally.
+	 */
+	private static function sync_connection_handle( string $handle ): void {
```
kraftbj, line 340: Minor: on the disconnect-with-revert path this writes the previous handle into `atmosphere_connection`, and then `Client::disconnect()` immediately deletes the whole option a few lines later in `Admin::handle_disconnect`. The write is wasted in that flow.

It's defensible — `Handle` correctly stays decoupled from disconnect ordering, and the sync is genuinely useful in the (today hypothetical) future where someone calls `maybe_revert_on_disconnect` outside the disconnect handler. Worth a one-line comment here so the apparent redundancy doesn't look like an oversight when read against the disconnect handler.

includes/wp-admin/class-admin.php
```diff
@@ -551,6 +631,26 @@ public static function handle_oauth_callback(): void {
...
 	}
 
+	/**
+	 * Handle the explicit "use my domain as my Bluesky handle" submission.
+	 *
+	 * Verifies capability + nonce, defers to {@see Handle::set_handle()} for
+	 * the actual call, then redirects back to the Settings page with a notice
+	 * already populated by Handle.
+	 */
+	public static function handle_set_domain_handle(): void {
```
kraftbj, line 656: **Missing test coverage:** the two security gates here (`current_user_can` + `check_admin_referer`) currently have no PHPUnit coverage. The disconnect handler probably has the same gap, but this PR is the right time to pin the contract before this admin handler becomes the front door for an XRPC `updateHandle` call against the user's connected PDS.

Minimal coverage I'd want:
- Hit the action without `manage_options` → expect `wp_die` (use `WP_UnitTestCase::expectException( WPDieException::class )` from the WP test suite).
- Hit the action with no/bad nonce → expect rejection before `Handle::set_handle()` runs (assert via a spying filter on `atmosphere_pre_update_handle` that it never fires).
- Happy path with both gates passing → assert redirect URL.

Belt-and-braces, but cheap, and pins the security contract for whoever touches this next.

tests/phpunit/tests/class-test-handle.php
```diff
@@ -0,0 +1,334 @@
...
+		\remove_filter( Handle::FILTER_ENABLED, '__return_false' );
+	}
+
+	/**
+	 * Test that is_root_install returns true for a root domain.
+	 */
+	public function test_is_root_install_for_root(): void {
+		\add_filter( 'home_url', static fn() => 'https://example.com' );
+		$this->assertTrue( Handle::is_root_install() );
+		\remove_all_filters( 'home_url' );
```
kraftbj: `\remove_all_filters( 'home_url' )` removes core's filters too, not just the closure this test added — `home_url` has callbacks attached from multisite, `wp_force_ssl_admin`, and various plugins.

Within a `WP_UnitTestCase` run that's mostly fine (state resets between tests), but if any future bootstrap reattaches at suite-load time rather than per-test, you'll get cross-test bleed that's a nightmare to debug. Capture the closure and pair it with a targeted `\remove_filter`, or move the add/remove into `setUp`/`tearDown`. Same pattern repeats throughout this file.

kraftbj (dismissed):
Approving — this is well-shaped: small surface, conservative defaults, snapshot-then-revert lifecycle is the right model, and the test coverage on `Handle` in isolation is solid.

A handful of notes inline. None are blockers — happy for any of them to land here or in a follow-up. If I had to pick the two I'd suggest tackling first as follow-ups, it's:

1. **Confirm `is_root_install` behavior on the `home_url at host root + site_url in subdirectory` install layout.** The current check only looks at `home_url`'s path, but the well-known endpoint actually resolves through WordPress's rewrite layer rooted at `site_url`. If verification 404s on that layout, the panel will offer the action while the subsequent Bluesky verification fails downstream. Either gate that case in `is_root_install` or document it as unsupported in testing step 6.
2. **One integration test for `handle_set_domain_handle`'s capability + nonce gates.** The `Handle` unit tests are great, but the admin handler's two security gates currently have no coverage. Worth pinning before this is the front door for an XRPC call against the user's connected account.

Everything else is polish or "nice if you're already in the file."

kraftbj (approved):

PR #58, opened 2026-05-14
Skip update_post fresh-publish fallback for unsynced legacy posts

### LINKED ISSUE #145: Updating and saving an existing WP post creates a post on Bsky
I was going back and editing some old posts in WordPress. I made an edit and saved, made another edit and saved. When I looked at Bsky, I had two new posts with the same content from WordPress.

I was expecting that it would cross-post for something new, not editing an old post.

kraftbj:
## Proposed changes:

`Publisher::update_post()` currently falls back to a fresh `publish_post()` whenever the post has no stored thread records:

```php
if ( empty( $stored ) ) {
    // Never successfully published — do a fresh publish.
    return self::publish_post( $post );
}
```

That branch fires on every routine edit of a WordPress post that pre-dates the Atmosphere connection (or pre-dates the post type becoming supported). Saving a small typo fix on a years-old WP post silently mints a brand-new Bluesky record with the original post date. Surprising, not what the user asked for, and easy to repeat — every subsequent edit before the cron job lands the metadata creates another duplicate. Fixes Automattic/fosse#145.

This PR distinguishes the two cases that reach the empty-stored branch:

- **Pristine post** — `Post::META_TID` is also unset. `publish_post()` has never run for this post (no `Transformer::get_rkey()` call to reserve the rkey). The user never asked Atmosphere to mirror it; doing so on a routine edit is the bug. **Now: skip the publish, fire `atmosphere_update_skipped_unsynced_post` so subscribers can observe the skip, and return an empty array.** The deliberate path for retro-syncing existing posts remains the Backfill admin workflow.
- **Failed prior attempt** — `Post::META_TID` was set by `get_rkey()` during a previous publish attempt but `META_URI` is empty, so `stored_thread_records()` still returns empty (see its existing rationale at `class-publisher.php:1540-1545`). The reserved TID must be reused on the next attempt. **Unchanged behavior: fall back to `publish_post()` so the publish retries.**

The `META_TID` check is the marker between the two: it's only ever written by `Transformer::get_rkey()` once `publish_post()` (or `rewrite_thread()`) has started running.

### Other information:

`test_update_skips_post_with_no_publication_history` pins the new behavior:

- Pristine `publish` post (no `META_TID`, no `META_URI`).
- Calls `Publisher::update_post()`.
- Asserts: empty-array return, `atmosphere_update_skipped_unsynced_post` action fired with the right post, no `applyWrites` attempted (via the `atmosphere_pre_apply_writes` capture used elsewhere in this file), and no `META_TID` was written (which would prove `publish_post()` reached `get_rkey()`).

The existing `test_update_falls_back_to_publish_without_uris` documents and protects the retry case (`META_TID` set, `META_URI` empty → still falls through to `publish_post()`). It continues to pass unchanged.

## Testing instructions:

1. Connect Atmosphere to a Bluesky DID.
2. Publish a new WP post — verify it appears on Bluesky as today.
3. Edit and save that post — verify Bluesky shows an `applyWrites#update` (in-place update), not a second record.
4. Find a WP post that was published before connecting (or temporarily set its `_atmosphere_bsky_tid` and `_atmosphere_doc_tid` meta to empty and clear `_atmosphere_bsky_records_v2`).
5. Edit and save that post — **on trunk today** a fresh Bluesky record appears; **with this PR** nothing appears on Bluesky and `did_action( 'atmosphere_update_skipped_unsynced_post' )` returns 1.
6. To explicitly sync that legacy post, use the existing Backfill workflow.

### Changelog entry

---

Companion references:

- Downstream report: Automattic/fosse#145


pfefferle (approved):
Nice one!

PR #61, opened 2026-05-15
Fix: harden OAuth resolution chain and DPoP transient storage

pfefferle:
Fixes the three High findings from the pre-launch security audit.

### ATM-001 — SSRF / poisoned-URL surface in the resolution chain (`includes/oauth/class-resolver.php`)

The handle → DID → PDS → auth-server chain returns URLs (`serviceEndpoint`, `authorization_servers[0]`, `token_endpoint`, `authorization_endpoint`, `pushed_authorization_request_endpoint`) that are persisted into the connection. Every one of these is now validated before it can leave the resolver:

* **Handles and `did:web` hosts** are validated as DNS-style identifiers (`is_valid_handle()`). Rejects single-label hosts like `localhost`, percent-encoded forms, IP literals, and anything outside `[A-Za-z0-9-]`. Two+ labels required.
* **URLs in response bodies** are validated with `is_safe_https_url()` — `wp_http_validate_url()` (private-IP / loopback / link-local blocklist) plus an explicit `https://` scheme check. The HTTPS narrowing matters because WordPress's safe URL gate allows `http`/`https`/`ssl`, but the AT Protocol spec requires HTTPS.
* The helper is **only** used where we'd otherwise persist a URL — pre-fetch URL checks were redundant with the safe variants and were removed.

### ATM-002 — DPoP private key in plaintext transient (`includes/oauth/class-client.php`)

`authorize()` previously wrote the freshly-generated ES256 private JWK to `set_transient('atmosphere_oauth_dpop_jwk', $dpop_jwk, ...)` — in plaintext, for up to one hour. Any reader of `wp_options` (DB, object cache, WP-CLI) could lift the key.

Fix: encrypt with `Encryption::encrypt()` on write, decrypt on read in `handle_callback()`. Matches the encryption already applied to the same key once the connection is persisted.

**In-flight sessions across the upgrade** (user starts OAuth on old code, finishes on new code) are handled explicitly — the callback short-circuits with `atmosphere_expired` and a clear message asking the user to reconnect.

### ATM-003 — `wp_remote_*` → `wp_safe_remote_*` (resolver + client + api)

WordPress's safe variants apply the host-safety blocklist (RFC 1918, loopback, link-local) automatically; the unsafe variants don't. Every call in `includes/oauth/class-resolver.php`, `includes/oauth/class-client.php`, and `includes/class-api.php` now goes through the safe variant.

### Tests

New regression tests under `tests/phpunit/tests/oauth/`:

* `class-test-resolver.php` — 13 tests covering: invalid handle rejection (incl. single-label and percent-encoded), `did:web` host validation, internal-host PDS endpoints, non-HTTPS endpoints, exotic schemes (`file://`), unsafe issuer URLs, unsafe `token_endpoint` / `authorization_endpoint` / `pushed_authorization_request_endpoint`, clean-chain success path.
* `class-test-client-authorize.php` — DPoP JWK transient encryption round-trip (asserts the ciphertext doesn't contain the private `d` parameter), plus in-flight-upgrade case (plaintext transient → `atmosphere_expired`, never reaches the decryption path with stale key material).

### Other information:

- [x] Have you written new tests for your changes, if applicable? — yes, 14 new tests.

### Backwards compatibility

Safe for production wp.org installs targeting public AT Protocol providers (bsky.social, public custom PDS on HTTPS). Two edge cases worth calling out:

* **Self-hosted PDS at a private IP / localhost** — `wp_safe_remote_*` will refuse to connect. Admins can opt those hosts back in via `WP_ACCESSIBLE_HOSTS` or the `http_request_host_is_external` filter. This is the same behaviour as every other plugin that uses the safe variants.
* **In-flight OAuth sessions across the upgrade** — fail cleanly with `atmosphere_expired`, user reconnects. No data loss.

Already-connected sites are unaffected — the connection data was validated when established, and the per-request path only changes from `wp_remote_request` to `wp_safe_remote_request` on the trusted PDS URL.

No public API signatures changed.

## Testing instructions:

* `composer lint` passes (verified locally).
* `npm run env-test -- --group=oauth` should pass in CI. Local test runs weren't possible — Docker daemon isn't running on the dev machine, no local MySQL; CI will run them on PR open.
* Manual: connect to a real Bluesky account via the settings page; verify the OAuth round-trip still works end-to-end and a post gets cross-posted.
* Try connecting with an invalid handle (`localhost`, `not-a-handle`, `http://evil/`) — should surface `atmosphere_invalid_handle` cleanly.
* Read `wp_options` (or `wp option get _transient_atmosphere_oauth_dpop_jwk` mid-flow) — the transient value should be base64 ciphertext, not a serialised PHP array.

### Changelog entry

Already committed at `.github/changelog/security-audit-highs` (`Significance: patch`, `Type: security`).

### DIFF
--- a/includes/oauth/class-client.php
+++ b/includes/oauth/class-client.php
@@ -101,14 +101,21 @@ class Client {
 		 * in plaintext in `wp_options` — encrypt it the same way the
 		 * persisted connection encrypts the key after callback
 		 * (`Encryption::encrypt()` below at the option write).
+		 *
+		 * Compute the ciphertext BEFORE writing any transients.
+		 * `Encryption::encrypt()` ultimately calls `random_bytes()`
+		 * and `sodium_crypto_secretbox()`, both of which can throw.
+		 * If we wrote verifier + state first and the encrypt then
+		 * threw, those transients would sit orphaned for an hour and
+		 * the user would see `atmosphere_expired` on retry until they
+		 * expired naturally. Doing the encrypt first means a thrown
+		 * exception bubbles up cleanly with no orphans.
 		 */
+		$dpop_jwk_ciphertext = Encryption::encrypt( (string) \wp_json_encode( $dpop_jwk ) );
+
 		\set_transient( 'atmosphere_oauth_verifier', $verifier, HOUR_IN_SECONDS );
 		\set_transient( 'atmosphere_oauth_state', $state, HOUR_IN_SECONDS );
-		\set_transient(
-			'atmosphere_oauth_dpop_jwk',
-			Encryption::encrypt( (string) \wp_json_encode( $dpop_jwk ) ),
-			HOUR_IN_SECONDS
-		);
+		\set_transient( 'atmosphere_oauth_dpop_jwk', $dpop_jwk_ciphertext, HOUR_IN_SECONDS );
 		\set_transient(
 			'atmosphere_oauth_resolved',
 			array(
--- a/includes/oauth/class-client.php
+++ b/includes/oauth/class-client.php
@@ -95,10 +95,20 @@ class Client {
 		// State for CSRF.
 		$state = \wp_generate_password( 40, false );
 
-		// Persist transient data needed for callback.
+		/*
+		 * Persist transient data needed for callback. The DPoP JWK
+		 * contains the ES256 private `d` parameter and must never sit
+		 * in plaintext in `wp_options` — encrypt it the same way the
+		 * persisted connection encrypts the key after callback
+		 * (`Encryption::encrypt()` below at the option write).
+		 */
 		\set_transient( 'atmosphere_oauth_verifier', $verifier, HOUR_IN_SECONDS );
 		\set_transient( 'atmosphere_oauth_state', $state, HOUR_IN_SECONDS );
-		\set_transient( 'atmosphere_oauth_dpop_jwk', $dpop_jwk, HOUR_IN_SECONDS );
+		\set_transient(
+			'atmosphere_oauth_dpop_jwk',
+			Encryption::encrypt( (string) \wp_json_encode( $dpop_jwk ) ),
+			HOUR_IN_SECONDS
+		);
 		\set_transient(
 			'atmosphere_oauth_resolved',
 			array(
@@ -175,7 +185,7 @@ class Client {
 			'login_hint'            => $did,
 		);
 
-		$response = \wp_remote_post(
+		$response = \wp_safe_remote_post(
 			$par_url,
 			array(
 				'headers' => array(
@@ -211,7 +221,7 @@ class Client {
 				return new \WP_Error( 'atmosphere_dpop', \__( 'DPoP nonce retry failed.', 'atmosphere' ) );
 			}
 
-			$response = \wp_remote_post(
+			$response = \wp_safe_remote_post(
 				$par_url,
 				array(
 					'headers' => array(
@@ -270,13 +280,37 @@ class Client {
 		}
 		\delete_transient( 'atmosphere_oauth_verifier' );
 
-		$dpop_jwk = \get_transient( 'atmosphere_oauth_dpop_jwk' );
-		$resolved = \get_transient( 'atmosphere_oauth_resolved' );
+		$dpop_jwk_blob = \get_transient( 'atmosphere_oauth_dpop_jwk' );
+		$resolved      = \get_transient( 'atmosphere_oauth_resolved' );
 		\delete_transient( 'atmosphere_oauth_dpop_jwk' );
 		\delete_transient( 'atmosphere_oauth_resolved' );
 
-		if ( ! $dpop_jwk || ! $resolved ) {
+		if ( ! $dpop_jwk_blob || ! $resolved ) {
 			return new \WP_Error( 'atmosphere_expired', \__( 'OAuth session data missing. Please try again.', 'atmosphere' ) );
+		}
+
+		/*
+		 * Pre-encryption versions of the plugin stored the DPoP JWK as
+		 * a plain array in the transient. A user who started OAuth on
+		 * the old code and completes the callback after upgrading
+		 * sees a non-string value here. Treat it as an expired
+		 * session so the user is prompted to restart cleanly.
+		 */
+		if ( ! \is_string( $dpop_jwk_blob ) ) {
+			return new \WP_Error(
+				'atmosphere_expired',
+				\__( 'OAuth session predates the latest update. Please try connecting again.', 'atmosphere' )
+			);
+		}
+
+		$dpop_jwk_json = Encryption::decrypt( $dpop_jwk_blob );
+		if ( false === $dpop_jwk_json ) {
+			return new \WP_Error( 'atmosphere_decrypt', \__( 'Failed to decrypt OAuth session key.', 'atmosphere' ) );
+		}
+
+		$dpop_jwk = \json_decode( $dpop_jwk_json, true );
+		if ( ! \is_array( $dpop_jwk ) ) {
+			return new \WP_Error( 'atmosphere_expired', \__( 'OAuth session key is malformed. Please try again.', 'atmosphere' ) );
 		}
 
 		$token_endpoint = $resolved['auth_server']['token_endpoint'];
@@ -295,7 +329,7 @@ class Client {
 			'code_verifier' => $verifier,
 		);
 
-		$response = \wp_remote_post(
+		$response = \wp_safe_remote_post(
 			$token_endpoint,
 			array(
 				'headers' => array(
@@ -329,7 +363,7 @@ class Client {
 				return new \WP_Error( 'atmosphere_dpop', \__( 'DPoP nonce retry failed during token exchange.', 'atmosphere' ) );
 			}
 
-			$response = \wp_remote_post(
+			$response = \wp_safe_remote_post(
 				$token_endpoint,
 				array(
 					'headers' => array(
@@ -408,7 +442,7 @@ class Client {
 			'client_id'     => self::client_id(),
 		);
 
-		$response = \wp_remote_post(
+		$response = \wp_safe_remote_post(
 			$token_endpoint,
 			array(
 				'headers' => array(
@@ -442,7 +476,7 @@ class Client {
 				return new \WP_Error( 'atmosphere_dpop', \__( 'DPoP nonce retry failed during refresh.', 'atmosphere' ) );
 			}
 
-			$response = \wp_remote_post(
+			$response = \wp_safe_remote_post(
 				$token_endpoint,
 				array(
 					'headers' => array(
--- a/includes/oauth/class-client.php
+++ b/includes/oauth/class-client.php
@@ -300,12 +300,12 @@ class Client {
 		 * Pre-encryption versions of the plugin stored the DPoP JWK as
 		 * a plain array in the transient. A user who started OAuth on
 		 * the old code and completes the callback after upgrading
-		 * sees a non-string value here. Treat it as an expired
-		 * session so the user is prompted to restart cleanly.
+		 * sees a non-string value here. Distinct error code from
+		 * "transient expired" so support can tell the two cases apart.
 		 */
 		if ( ! \is_string( $dpop_jwk_blob ) ) {
 			return new \WP_Error(
-				'atmosphere_expired',
+				'atmosphere_legacy_session',
 				\__( 'OAuth session predates the latest update. Please try connecting again.', 'atmosphere' )
 			);
 		}
@@ -317,7 +317,10 @@ class Client {
 
 		$dpop_jwk = \json_decode( $dpop_jwk_json, true );
 		if ( ! \is_array( $dpop_jwk ) ) {
-			return new \WP_Error( 'atmosphere_expired', \__( 'OAuth session key is malformed. Please try again.', 'atmosphere' ) );
+			return new \WP_Error(
+				'atmosphere_session_malformed',
+				\__( 'OAuth session key is malformed. Please try again.', 'atmosphere' )
+			);
 		}
 
 		$token_endpoint = $resolved['auth_server']['token_endpoint'];
--- a/includes/oauth/class-client.php
+++ b/includes/oauth/class-client.php
@@ -51,6 +51,15 @@ class Client {
 	/**
 	 * OAuth callback URI (admin page with special query param).
 	 *
+	 * The return value of this method is sent to the auth server as
+	 * `redirect_uri` in the PAR / authorize request and also baked into
+	 * the public client metadata document. If a third-party filter
+	 * returns a non-admin URL, the auth server happily redirects the
+	 * authorization code to that destination on completion — which is
+	 * an OAuth token-leak primitive. We validate the filter return is
+	 * a string pointing at this site's admin and fall back to the
+	 * default URI on anything else.
+	 *
 	 * @return string
 	 */
 	public static function redirect_uri(): string {
@@ -61,11 +70,24 @@ class Client {
 		 *
 		 * Allows consumers (e.g. wrapper plugins) to set their own callback
 		 * URL so the OAuth flow returns to their admin page instead of the
-		 * default Atmosphere settings page.
+		 * default Atmosphere settings page. The return MUST be a string
+		 * pointing at this site's admin area — anything else is rejected
+		 * to prevent a malicious or buggy filter from redirecting OAuth
+		 * codes off-site.
 		 *
 		 * @param string $uri Default redirect URI.
 		 */
-		return \apply_filters( 'atmosphere_oauth_redirect_uri', $uri );
+		$filtered = \apply_filters( 'atmosphere_oauth_redirect_uri', $uri );
+
+		if ( ! \is_string( $filtered ) || '' === $filtered ) {
+			return $uri;
+		}
+
+		if ( ! \str_starts_with( $filtered, \admin_url( '' ) ) ) {
+			return $uri;
+		}
+
+		return $filtered;
 	}
 
 	/**
--- a/includes/oauth/class-resolver.php
+++ b/includes/oauth/class-resolver.php
@@ -195,8 +195,17 @@ class Resolver {
 		 * Malformed JSON (or a valid scalar/null payload like
 		 * `"foo"` or `null`) decodes to a non-array. Bail before
 		 * indexing so PHP 8 doesn't TypeError on offset access.
+		 *
+		 * Guard the parent `authorization_servers` separately: on
+		 * PHP 8.1+ `empty( $resource['authorization_servers'][0] )`
+		 * still evaluates the inner offset access first, and if the
+		 * value is a scalar (`'foo'`, `42`) the offset read emits a
+		 * warning before `empty()` short-circuits. Checking
+		 * `is_array()` on the parent first sidesteps that.
 		 */
 		if ( ! \is_array( $resource )
+			|| ! isset( $resource['authorization_servers'] )
+			|| ! \is_array( $resource['authorization_servers'] )
 			|| empty( $resource['authorization_servers'][0] )
 			|| ! \is_string( $resource['authorization_servers'][0] )
 		) {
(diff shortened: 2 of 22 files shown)


includes/oauth/class-resolver.php
```diff
@@ -111,7 +129,16 @@ public static function pds_from_did_doc( array $did_doc ): string|\WP_Error {
 			$type = $service['type'] ?? '';
 
 			if ( '#atproto_pds' === $id && 'AtprotoPersonalDataServer' === $type && ! empty( $service['serviceEndpoint'] ) ) {
-				return $service['serviceEndpoint'];
+				$endpoint = $service['serviceEndpoint'];
+
+				if ( ! \is_string( $endpoint ) || ! self::is_safe_https_url( $endpoint ) ) {
```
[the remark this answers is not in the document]
pfefferle: Fixed in e6d2a11. Added two guards: reject the whole DID doc when `service` itself isn't an array (`atmosphere_invalid_did_doc`), and skip individual entries that aren't arrays during the foreach (falls through cleanly to `atmosphere_no_pds`). Two new regression tests pin both shapes — `test_pds_from_did_doc_tolerates_non_array_service_field` and `test_pds_from_did_doc_tolerates_scalar_service_entries`.

pfefferle:
Thanks for the review. Both summary findings (the `is_array` guards on `$resource` and `$meta` after `json_decode`) were actioned in 763d66f. The inline finding on `pds_from_did_doc` service-field type guards is fixed in e6d2a11. All have new regression tests.

includes/oauth/class-resolver.php
```diff
@@ -202,4 +300,79 @@ public static function resolve( string $handle ): array|\WP_Error {
...
+	 *    further than WordPress's `http`/`https`/`ssl` allowlist)
+	 *  - has a host
+	 *  - no embedded `user:pass@` credentials (a known URL-injection
+	 *    vector that would otherwise be carried into the persisted
+	 *    connection)
+	 *
+	 * @param mixed $url URL to validate.
+	 * @return bool
+	 */
+	private static function is_safe_https_url( $url ): bool {
```
kraftbj: **[HIGH] IP-literal hosts bypass this gate.**

`is_safe_https_url()` is the acceptance gate for every URL the resolver returns (PDS, issuer, token, authorization, PAR). It checks scheme + host presence + userinfo, but not whether the host is a literal address. So `https://169.254.169.254/...`, `https://127.0.0.1`, `https://[::1]`, and IPv6 ULAs all pass.

What the existing safety net actually catches:
- The PDS / token / PAR URLs are fetched via `wp_safe_remote_*`, which goes through `wp_http_validate_url()`. That blocks RFC1918 + loopback but **not** link-local 169.254/16 (cloud metadata) and is IPv4-centric overall.
- `authorization_endpoint` is consumed by `\wp_redirect( $auth_url )` in `Admin::sanitize_handle()` (`includes/wp-admin/class-admin.php:367`). No host-safety call fires on that path.

The regression tests use `http://169.254.169.254` (scheme=http), so they pass on the scheme check and give no evidence about the host-blocklist claim.

**Suggested fix:**

```php
if ( \filter_var( $host, FILTER_VALIDATE_IP ) !== false ) {
    return false;
}
```

Add this after the host-presence check. Then add regression cases for `https://169.254.169.254`, `https://127.0.0.1`, `https://[::1]`, and `https://[fd00::1]`. Each should be rejected before it reaches any HTTP transport.

Flagged by both the multi-persona review (adversarial reviewer, P1, conf 50) and the codex adversarial pass (HIGH). Cross-corroboration.

includes/oauth/class-resolver.php
```diff
@@ -202,4 +300,79 @@ public static function resolve( string $handle ): array|\WP_Error {
...
+		if ( ! \is_string( $url ) || '' === $url ) {
+			return false;
+		}
+
+		$parts = \wp_parse_url( $url );
+		if ( ! \is_array( $parts ) ) {
+			return false;
+		}
+
+		if ( ( $parts['scheme'] ?? '' ) !== 'https' ) {
```
kraftbj: Scheme comparison is case-sensitive — `HTTPS://example.com` is rejected even though RFC 3986 says scheme is case-insensitive.

```php
if ( \strtolower( $parts['scheme'] ?? '' ) !== 'https' ) {
    return false;
}
```

Not exploitable, just cleaner.

includes/oauth/class-resolver.php
```diff
@@ -126,44 +172,96 @@ public static function pds_from_did_doc( array $did_doc ): string|\WP_Error {
...
 		$resource = \json_decode( \wp_remote_retrieve_body( $response ), true );
 
-		if ( empty( $resource['authorization_servers'][0] ) ) {
+		/*
+		 * Malformed JSON (or a valid scalar/null payload like
+		 * `"foo"` or `null`) decodes to a non-array. Bail before
+		 * indexing so PHP 8 doesn't TypeError on offset access.
+		 */
+		if ( ! \is_array( $resource )
+			|| empty( $resource['authorization_servers'][0] )
```
kraftbj, line 209: On PHP 8.1+, when `$resource` is an array but `$resource['authorization_servers']` is `null` or a scalar, indexing `[0]` emits a `null offset` deprecation before `empty()` catches it. The chain still returns the right `WP_Error`, but the noise will hit production logs.

```php
if ( ! isset( $resource['authorization_servers'] ) || ! \is_array( $resource['authorization_servers'] ) ) {
    return new \WP_Error( 'atmosphere_no_auth_server', /* ... */ );
}
if ( empty( $resource['authorization_servers'][0] ) || ! \is_string( $resource['authorization_servers'][0] ) ) {
```

Guard the parent before indexing.

includes/oauth/class-resolver.php
```diff
@@ -202,4 +300,79 @@ public static function resolve( string $handle ): array|\WP_Error {
...
+	 *
+	 * @param string $host Hostname (handle or did:web domain).
+	 * @return bool
+	 */
+	private static function is_valid_handle( string $host ): bool {
+		if ( '' === $host || \strlen( $host ) > 253 ) {
+			return false;
+		}
+
+		$label = '[A-Za-z0-9](?:[A-Za-z0-9-]{0,61}[A-Za-z0-9])?';
```
kraftbj, line 363: Minor spec gap, not exploitable: `is_valid_handle()` accepts numeric-only TLDs (`a.b.123`) and reserved atproto TLDs (`.local`, `.localhost`, `.arpa`, `.internal`, `.invalid`, `.onion`, `.test`, `.example`, `.alt`). The AT Protocol handle spec rejects all of these.

For the SSRF angle the IP-literal fix above covers numeric labels. For spec compliance, reject the reserved TLD list explicitly.

Leaving as a P3 since runtime risk is bounded by the resolver chain.

includes/oauth/class-client.php
```diff
@@ -95,10 +95,20 @@ public static function authorize( string $handle ): string|\WP_Error {
...
+		 * Persist transient data needed for callback. The DPoP JWK
+		 * contains the ES256 private `d` parameter and must never sit
+		 * in plaintext in `wp_options` — encrypt it the same way the
+		 * persisted connection encrypts the key after callback
+		 * (`Encryption::encrypt()` below at the option write).
+		 */
 		\set_transient( 'atmosphere_oauth_verifier', $verifier, HOUR_IN_SECONDS );
 		\set_transient( 'atmosphere_oauth_state', $state, HOUR_IN_SECONDS );
-		\set_transient( 'atmosphere_oauth_dpop_jwk', $dpop_jwk, HOUR_IN_SECONDS );
+		\set_transient(
```
kraftbj: Ordering concern: `set_transient( 'atmosphere_oauth_verifier', ... )` and `set_transient( 'atmosphere_oauth_resolved', ... )` already ran before this `Encryption::encrypt()` call. `Encryption::encrypt()` can throw `RandomException` (or return WP_Error on key-derivation failure, depending on path). When it does, the verifier + resolved transients sit in storage for the full TTL with no JWK to pair against.

Not dangerous — the orphans are deleted next time `authorize()` runs — but produces confusing `atmosphere_expired` errors for an hour.

Fix: build the ciphertext first, only call `set_transient` after every encrypt succeeds. Or wrap in try/catch and `return new \WP_Error( 'atmosphere_encrypt', ... )` so the user sees the real failure.

includes/oauth/class-client.php
```diff
@@ -270,15 +280,39 @@ public static function handle_callback( string $code, string $state ): true|\WP_
...
-		$dpop_jwk = \get_transient( 'atmosphere_oauth_dpop_jwk' );
-		$resolved = \get_transient( 'atmosphere_oauth_resolved' );
+		$dpop_jwk_blob = \get_transient( 'atmosphere_oauth_dpop_jwk' );
+		$resolved      = \get_transient( 'atmosphere_oauth_resolved' );
 		\delete_transient( 'atmosphere_oauth_dpop_jwk' );
 		\delete_transient( 'atmosphere_oauth_resolved' );
 
-		if ( ! $dpop_jwk || ! $resolved ) {
+		if ( ! $dpop_jwk_blob || ! $resolved ) {
 			return new \WP_Error( 'atmosphere_expired', \__( 'OAuth session data missing. Please try again.', 'atmosphere' ) );
```
kraftbj, line 318: `atmosphere_expired` is doing triple duty in `handle_callback()`:
- line 289 — transient missing entirely (user took too long, normal expiry)
- line 301 — transient is a plain array (pre-encryption upgrade path)
- line 313 — decrypt succeeded but JSON didn't decode to an array (truly malformed)

These have different remediation. The pre-upgrade case wants "please reconnect." The malformed-JSON case wants "something corrupted the transient." Currently support can't distinguish.

Suggested split:
- keep `atmosphere_expired` for line 289
- `atmosphere_legacy_session` for line 301
- `atmosphere_session_malformed` for line 313

includes/oauth/class-client.php
```diff
@@ -270,15 +280,39 @@ public static function handle_callback( string $code, string $state ): true|\WP_
...
+		 */
+		if ( ! \is_string( $dpop_jwk_blob ) ) {
+			return new \WP_Error(
+				'atmosphere_expired',
+				\__( 'OAuth session predates the latest update. Please try connecting again.', 'atmosphere' )
+			);
+		}
+
+		$dpop_jwk_json = Encryption::decrypt( $dpop_jwk_blob );
+		if ( false === $dpop_jwk_json ) {
```
kraftbj, line 336: The `Encryption::decrypt() === false` branch is untested. The pre-upgrade test seeds a plaintext array, which short-circuits at line 299's `is_string` check before reaching this code.

The real-world trigger is `AUTH_KEY` / `AUTH_SALT` rotation between `authorize()` and the callback — the user is mid-OAuth, the secrets rotate, the in-flight transient won't decrypt. Right now nobody knows whether the resulting WP_Error code is what they expect.

```php
public function test_handle_callback_returns_decrypt_error_on_corrupted_jwk() {
    \set_transient( 'atmosphere_oauth_dpop_jwk', 'not-real-ciphertext', HOUR_IN_SECONDS );
    // also seed verifier + resolved with valid encrypted values, plus a callback state
    $result = Client::handle_callback( /* fake args matching state */ );
    $this->assertWPError( $result );
    $this->assertSame( 'atmosphere_decrypt', $result->get_error_code() );
}
```

If the code split above lands too, add a second test for the malformed-JSON branch (decrypt succeeds, plaintext is `'not-json'`).

tests/phpunit/tests/oauth/class-test-client-authorize.php
```diff
@@ -0,0 +1,122 @@
...
+	 * The DPoP JWK transient set during `authorize()` is encrypted
+	 * with `Encryption::encrypt()`. Decrypts back to a JWK array
+	 * with a private `d` parameter (ES256 key material).
+	 *
+	 * We don't exercise the full `authorize()` path here — it would
+	 * require mocking PAR + the full resolution chain. Instead this
+	 * test pins the storage contract: any plaintext JWK in the
+	 * transient is a regression of ATM-002.
+	 */
+	public function test_dpop_jwk_transient_encrypted_round_trips() {
```
kraftbj, line 74: Coverage gap: `test_dpop_jwk_transient_encrypted_round_trips` exercises `Encryption::encrypt()` directly — it never calls `Client::authorize()`. A refactor that dropped the `Encryption::encrypt()` wrapper in `authorize()` (writing the JWK plaintext again) would not break this test.

The assertion is real, just not pinned to the right entry point. Suggested end-to-end version:

```php
public function test_authorize_persists_encrypted_jwk_transient() {
    // Stub the resolver chain via pre_http_request so we don't need a network.
    Client::authorize( 'alice.example.com' );
    $stored = \get_transient( 'atmosphere_oauth_dpop_jwk' );
    $this->assertIsString( $stored );
    $this->assertStringNotContainsString( '"d":', $stored ); // no private key fragment
    $jwk = \json_decode( Encryption::decrypt( $stored ), true );
    $this->assertArrayHasKey( 'd', $jwk );
    $this->assertArrayHasKey( 'kty', $jwk );
}
```

Keep the current direct-round-trip test as a unit test for `Encryption`. Add this one for `authorize()`.

includes/oauth/class-resolver.php
```diff
@@ -202,4 +300,79 @@ public static function resolve( string $handle ): array|\WP_Error {
...
+	 * hyphens, single-label hosts (`localhost`), and characters
+	 * outside `[A-Za-z0-9-]`. This is the first line of defence
+	 * against percent-encoded host bypasses and SSRF via crafted
+	 * DIDs — the regex never matches anything that contains a `%`
+	 * or other reserved URL characters.
+	 *
+	 * @param string $host Hostname (handle or did:web domain).
+	 * @return bool
+	 */
+	private static function is_valid_handle( string $host ): bool {
```
kraftbj: Coordination note with PR 62 — `class-facet.php:245` is about to land a byte-for-byte copy of this method. The maintainer comment on that PR flags consolidation as a follow-up.

The two PRs landing together is the right window to fix this. Easiest path: bump this and `is_safe_https_url()` to `public static` here, then PR 62's `Facet::is_valid_handle()` becomes `return Resolver::is_valid_handle( $handle );`. No new file needed.

If either side drifts after merge (IDN support, length cap change, label rule tweak), the mention path and the resolver path quietly disagree. Not a blocker for this PR alone — flagging because the fix is cheap right now and gets more expensive once both are in trunk.

kraftbj (changes requested):
Cross-referenced two passes here: an opinionated multi-persona review (correctness, security, testing, reliability, maintainability, adversarial, previous-comments) plus an independent codex adversarial run. Both verified Copilot's earlier feedback was actually addressed at HEAD — the `is_array` guards on `$resource` (line 199) and `$meta` (line 231), and the `pds_from_did_doc` service-field guards (lines 135, 143). All present, all backed by regression tests. Nice.

One high blocker, and a handful of smaller items inline.

**Blocker — `is_safe_https_url()` doesn't reject IP literals.** The helper validates string + parse + `https` + non-empty host + no userinfo, but doesn't touch the host's actual address space. `https://169.254.169.254/...`, `https://127.0.0.1`, and IPv6 loopback all pass. For PDS / token / PAR URLs the safety net is `wp_safe_remote_*`, which catches RFC1918 + loopback but not link-local 169.254/16. For `authorization_endpoint` there's no safety net at all — it goes to `wp_redirect()` in `Admin::sanitize_handle()` (after PR 62's parse/scheme/host check, which is also IP-literal-blind). The existing internal-host tests use `http://169.254...`, so they trip the scheme check and never exercise the host-blocklist claim. Recommend an explicit IP-literal kill in `is_safe_https_url()` plus regression cases that use `https://`.

Everything else inline is P2 / P3 polish — test coverage gaps, a couple of error-code clarifications, and one PHP 8.1 deprecation. The cross-PR coordination item (sharing `is_valid_handle()` with PR 62's facet copy) is also flagged on the facet side.

For context: PR 62 is also reviewed in parallel with three of its own high blockers (client_metadata redirect_uris content validation, multisite uninstall, mention DNS egress). Worth landing both fix waves before tagging the security release.

includes/oauth/class-resolver.php
pfefferle: Fixed in 0627f55. `is_safe_https_url()` now bails when `FILTER_VALIDATE_IP` accepts the host. PHP's `parse_url()` keeps the brackets on IPv6 hosts (`[::1]`), so I strip those before the filter runs. New regression cases cover loopback / link-local / RFC1918 / unspecified for v4 (5 cases) and loopback / ULA for v6 (2 cases), all on `https://` so they exercise the new IP check rather than the existing scheme gate.

includes/oauth/class-resolver.php
pfefferle: Fixed in 8c7a6e9. `\strtolower( $parts['scheme'] ?? '' ) !== 'https'`. Added a regression test that round-trips `HTTPS://pds.example.com` through `pds_from_did_doc()` and asserts it's accepted.

includes/oauth/class-resolver.php
pfefferle: Fixed in 7710db5. Added an explicit `is_array( $resource['authorization_servers'] )` guard before the inner index access so a scalar payload doesn't force PHP to evaluate offset-0 of a string/int before `empty()` short-circuits. Regression test stubs `{ "authorization_servers": "not-a-list" }` and asserts `atmosphere_no_auth_server`.

includes/oauth/class-resolver.php
pfefferle: Fixed in 1c0ebe7. `is_valid_handle()` now rejects TLDs that start with a digit or are entirely digits, plus the AT Protocol reserved list (`alt`, `arpa`, `example`, `internal`, `invalid`, `local`, `localhost`, `onion`, `test`). Two data-providered regression tests cover each case. Belt-and-suspenders with the IP-literal fix for the numeric angle — the spec rule is the cheaper layer.

includes/oauth/class-resolver.php
pfefferle: Done in 5ced865 — bumped both `is_valid_handle()` and `is_safe_https_url()` to `public static` with no implementation changes. PR 62's `Facet::is_valid_handle()` can now become `return Resolver::is_valid_handle( $handle );` and the two paths stay in lock-step. I'll thread this through on the PR 62 side once both have landed.

includes/oauth/class-client.php
pfefferle: Fixed in f5ef184. `Encryption::encrypt()` now runs into a local before any `set_transient()` call — if `random_bytes()` or `sodium_crypto_secretbox()` throws, the verifier / state transients never get written, so no orphans. Went with the "build ciphertext first" path rather than try/catch + `WP_Error` because there's nothing meaningful the caller can do with an `atmosphere_encrypt` error here — a thrown exception is the right signal that the host's crypto is broken.

includes/oauth/class-client.php
pfefferle: Fixed in a2c75a5. Split into three codes:

 * `atmosphere_expired` — transient missing entirely (natural expiry).
 * `atmosphere_legacy_session` — pre-encryption transient shape (upgrade path).
 * `atmosphere_session_malformed` — decrypt succeeded but plaintext isn't a JWK.

Each carries its own message string. Existing upgrade-path test updated to assert the new code.

includes/oauth/class-client.php
pfefferle: Covered in 3e09cf8. Two new tests:

 * `test_handle_callback_returns_decrypt_error_on_corrupted_jwk` — seeds a non-ciphertext string into the transient, asserts `atmosphere_decrypt`.
 * `test_handle_callback_returns_malformed_error_on_non_json_plaintext` — encrypts `'not-json'`, asserts `atmosphere_session_malformed` (the new error code from the split above).

tests/phpunit/tests/oauth/class-test-client-authorize.php
pfefferle: Added in 2fe6a97. `test_authorize_persists_encrypted_jwk_transient` stubs the full resolution chain (handle → DID → DID doc → PDS → auth server) via `pre_http_request`, calls `Client::authorize()`, then inspects the transient `authorize()` actually wrote. Asserts it's a ciphertext string with no plaintext JWK markers (`"d":`, `"kty":`), and that decrypting it round-trips to a JWK with both `kty` and the private `d`. The original direct-`Encryption::encrypt()` round-trip stays in place as a unit test for the helper.

`markTestSkipped()` if the resolver chain ever rejects the stubbed handle in a future environment, rather than silently masking the storage assertion.

kraftbj (approved):

PR #62, opened 2026-05-15
Fix: address remaining medium-severity audit findings

pfefferle:
Fixes the five Medium findings from the pre-launch security audit (ATM-004 through ATM-008). Follow-up to #61.

### ATM-004 — Auth-URL validation before redirect (`class-admin.php`)

`Admin::sanitize_handle()` no longer hands the resolver-derived auth URL straight to `wp_redirect`. New behaviour:

* Parse the URL; reject anything without an `https` scheme and a real host with an admin notice.
* Switch from `wp_redirect` to `wp_safe_redirect`, with the auth-server host opt-ed into `allowed_redirect_hosts` for the lifetime of the dying request (we `exit` immediately after).
* The `phpcs:ignore WordPress.Security.SafeRedirect.wp_redirect_wp_redirect` suppression is gone.

### ATM-005 — `atmosphere_oauth_redirect_uri` filter validation (`class-client.php`)

`Client::redirect_uri()` validates the filter return is a non-empty string that starts with `admin_url('')`. Non-string, empty, off-site, and exotic-scheme (`javascript:`, `data:`) returns fall back to the default — a malicious or buggy filter can no longer redirect OAuth codes off-site.

### ATM-006 — Defence-in-depth DNS gate (`class-facet.php`)

`Facet::resolve_mention()` now runs the same DNS-name syntax check as the resolver before calling `dns_get_record()`. The mention regex in `Facet::mentions()` already enforces this — the second gate documents the invariant close to the network call so a future regex loosening can't silently expose it. Invalid handles return an empty DID and `mentions()` skips emitting the facet (preventing malformed records from reaching the PDS).

### ATM-007 — Transformer + client_metadata filter return validation

All four `atmosphere_transform_*` filters and the `atmosphere_client_metadata` filter validate their return shape:

* Transformer filters (`transform_bsky_post`, `transform_comment`, `transform_document`, `transform_publication`): non-array returns fall back to the pre-filter record.
* `atmosphere_client_metadata`: must be an array containing `client_id` and `redirect_uris`; otherwise the unfiltered metadata is served so the OAuth flow keeps working.

### ATM-008 — Complete `uninstall.php` cleanup

Now removes the option / meta / transient keys that were leaking after delete:

| Surface | Keys |
|---------|------|
| Options | `atmosphere_long_form_composition`, `atmosphere_support_post_types`, `atmosphere_last_seen_notification`, `atmosphere_last_seen_own_*` (wildcard) |
| Post meta | `_atmosphere_bsky_thread_records`, `_atmosphere_bsky_uri_index`, `_atmosphere_bsky_orphan_records`, `_atmosphere_doc_ref_pending` |
| Comment meta | `_atmosphere_author_did`, `_atmosphere_author_avatar`, and `protocol = 'atproto'` (scoped by value so we never delete `protocol` meta written by other plugins) |
| Transients | `atmosphere_invalid_long_form_composition_logged`, `atmo_dpop_nonce_*` (wildcard) |

### Tests

14 new regression tests:

* `class-test-redirect-uri-filter.php` — 6 cases covering the ATM-005 filter contract (default, external, in-site admin, non-string, empty string, `javascript:`).
* `class-test-transform-filter-validation.php` — 5 cases asserting the transformer fallbacks for non-array, null, scalar, object returns plus the happy path.
* `class-test-facet.php` — single-label mention emits no facet (regression test for the ATM-006 user-visible behaviour).

### Other information:

- [x] Have you written new tests for your changes, if applicable? — yes, 14 new tests.

## Follow-ups left for separate PRs

* **ATM-004 test coverage** — `Admin::sanitize_handle()` redirects + exits, which makes it untestable in PHPUnit without a small refactor (extract the URL-validation into a private static helper). Doable; out of scope for this PR.
* **Consolidate `is_valid_handle()`** — `Resolver::is_valid_handle()` (in PR #61) and `Facet::is_valid_handle()` (here) are identical. Once #61 merges, fold them into a shared helper.

## Backwards compatibility

* Existing well-behaved `atmosphere_oauth_redirect_uri` filters that return an admin URL are unaffected.
* Existing well-behaved `atmosphere_transform_*` and `atmosphere_client_metadata` filters that return arrays are unaffected. A filter that returned a non-array would have caused a downstream `applyWrites` fatal anyway — the new fallback turns that into a quiet recovery.
* Already-installed sites that get *deleted* (not just deactivated) will now have a more thorough cleanup. Sites that just deactivate aren't affected.

No public API signatures changed.

## Testing instructions:

* `composer lint` passes (verified locally).
* CI should run `npm run env-test` — verify the new test files pass.
* Manual: connect to a real Bluesky account; verify OAuth still works end-to-end.
* Try delete-uninstall on a test site and inspect `wp_options` / `wp_postmeta` / `wp_commentmeta` for leftover keys — none of the listed ones should remain.

### Changelog entry

Already committed at `.github/changelog/security-audit-mediums` (`Significance: patch`, `Type: security`).

### DIFF
--- a/includes/transformer/class-facet.php
+++ b/includes/transformer/class-facet.php
@@ -198,13 +198,22 @@ class Facet {
 	/**
 	 * Resolve a handle to a DID for mention facets.
 	 *
-	 * Falls back to did:web if DNS resolution fails. Handles that
-	 * don't match the AT Protocol DNS-style syntax never reach the
-	 * `dns_get_record` call — without that gate, a post containing
-	 * `@evil-attacker-controlled-tld.example` would trigger a server
-	 * DNS lookup against an attacker-controlled domain (low-bandwidth
-	 * but reliable side-channel for exfiltrating data via subdomain
-	 * encoding).
+	 * Falls back to `did:web` if DNS resolution fails. The
+	 * `is_valid_handle()` gate below ensures only DNS-syntactically
+	 * valid handles reach `dns_get_record()` — that closes the
+	 * "malformed handle as DNS query smuggling" angle (e.g. control
+	 * characters or percent-encoded segments injected through a
+	 * regex relaxation), it does NOT block lookups against
+	 * attacker-controlled but well-formed domains.
+	 *
+	 * That broader exposure is by design: mention resolution requires
+	 * a DNS lookup against the mentioned handle's authoritative server,
+	 * and any user (commenter included) can mention any well-formed
+	 * domain. If that DNS-egress surface becomes a concern, the right
+	 * fix is at the threat-model layer (skip mention resolution on
+	 * the commenter path, allowlist mention authorities, or move to
+	 * DoH with a hard timeout) rather than tightening the syntactic
+	 * gate further.
 	 *
 	 * @param string $handle AT Protocol handle.
 	 * @return string DID string, or empty string if the handle is malformed.
--- a/includes/transformer/class-facet.php
+++ b/includes/transformer/class-facet.php
@@ -132,6 +132,16 @@ class Facet {
 
 			$did = self::resolve_mention( $handle );
 
+			/*
+			 * `resolve_mention()` returns an empty string when the
+			 * handle fails its (defence-in-depth) syntax check. Skip
+			 * the facet entirely in that case — sending an empty
+			 * `did` to Bluesky would have the PDS reject the record.
+			 */
+			if ( '' === $did ) {
+				continue;
+			}
+
 			$facets[] = array(
 				'index'    => array(
 					'byteStart' => $start,
@@ -188,15 +198,25 @@ class Facet {
 	/**
 	 * Resolve a handle to a DID for mention facets.
 	 *
-	 * Falls back to did:web if DNS resolution fails.
+	 * Falls back to did:web if DNS resolution fails. Handles that
+	 * don't match the AT Protocol DNS-style syntax never reach the
+	 * `dns_get_record` call — without that gate, a post containing
+	 * `@evil-attacker-controlled-tld.example` would trigger a server
+	 * DNS lookup against an attacker-controlled domain (low-bandwidth
+	 * but reliable side-channel for exfiltrating data via subdomain
+	 * encoding).
 	 *
 	 * @param string $handle AT Protocol handle.
-	 * @return string DID string.
+	 * @return string DID string, or empty string if the handle is malformed.
 	 */
 	private static function resolve_mention( string $handle ): string {
 		$conn = get_connection();
 		if ( ! empty( $conn['handle'] ) && \strtolower( $handle ) === \strtolower( $conn['handle'] ) ) {
 			return $conn['did'];
+		}
+
+		if ( ! self::is_valid_handle( $handle ) ) {
+			return '';
 		}
 
 		$records = @\dns_get_record( '_atproto.' . $handle, DNS_TXT ); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
@@ -210,5 +230,25 @@ class Facet {
 		}
 
 		return 'did:web:' . $handle;
+	}
+
+	/**
+	 * RFC 1035-style DNS-name validation, mirroring
+	 * `Resolver::is_valid_handle()`. Rejects empty strings, oversized
+	 * labels, leading/trailing hyphens, single-label hosts, and any
+	 * character outside `[A-Za-z0-9-]` — including percent-encoded
+	 * forms.
+	 *
+	 * @param string $host Handle to validate.
+	 * @return bool
+	 */
+	private static function is_valid_handle( string $host ): bool {
+		if ( '' === $host || \strlen( $host ) > 253 ) {
+			return false;
+		}
+
+		$label = '[A-Za-z0-9](?:[A-Za-z0-9-]{0,61}[A-Za-z0-9])?';
+
+		return (bool) \preg_match( '/^' . $label . '(?:\.' . $label . ')+$/', $host );
 	}
 }
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -192,7 +192,16 @@ class Post extends Base {
 			)
 		);
 
-		return \is_array( $filtered ) ? $filtered : $record;
+		if ( ! \is_array( $filtered ) ) {
+			\_doing_it_wrong(
+				__METHOD__,
+				\esc_html__( 'atmosphere_transform_bsky_post must return an array; falling back to the unfiltered record.', 'atmosphere' ),
+				'0.1.0'
+			);
+			return $record;
+		}
+
+		return $filtered;
 	}
 
 	/**
@@ -1008,7 +1017,16 @@ class Post extends Base {
 		/** This filter is documented in Post::transform() above. */
 		$filtered = \apply_filters( 'atmosphere_transform_bsky_post', $record, $this->object, $context );
 
-		return \is_array( $filtered ) ? $filtered : $record;
+		if ( ! \is_array( $filtered ) ) {
+			\_doing_it_wrong(
+				__METHOD__,
+				\esc_html__( 'atmosphere_transform_bsky_post must return an array; falling back to the unfiltered record.', 'atmosphere' ),
+				'0.1.0'
+			);
+			return $record;
+		}
+
+		return $filtered;
 	}
 
 	/**
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -173,11 +173,15 @@ class Post extends Base {
 		 * lint hooks) should use the `$context` array to distinguish
 		 * single-post output from teaser-thread entries.
 		 *
+		 * Filters that return a non-array fall back to the pre-filter
+		 * record — protects the applyWrites batch from a misbehaving
+		 * listener.
+		 *
 		 * @param array    $record Bsky post record.
 		 * @param \WP_Post $post   WordPress post.
 		 * @param array    $context Additional composition context.
 		 */
-		return \apply_filters(
+		$filtered = \apply_filters(
 			'atmosphere_transform_bsky_post',
 			$record,
 			$this->object,
@@ -187,6 +191,8 @@ class Post extends Base {
 				'is_thread_reply' => false,
 			)
 		);
+
+		return \is_array( $filtered ) ? $filtered : $record;
 	}
 
 	/**
@@ -1000,7 +1006,9 @@ class Post extends Base {
 		);
 
 		/** This filter is documented in Post::transform() above. */
-		return \apply_filters( 'atmosphere_transform_bsky_post', $record, $this->object, $context );
+		$filtered = \apply_filters( 'atmosphere_transform_bsky_post', $record, $this->object, $context );
+
+		return \is_array( $filtered ) ? $filtered : $record;
 	}
 
 	/**
@@ -1039,7 +1047,7 @@ class Post extends Base {
 		}
 
 		/** This filter is documented in Post::transform() above. */
-		return \apply_filters(
+		$filtered = \apply_filters(
 			'atmosphere_transform_bsky_post',
 			$record,
 			$this->object,
@@ -1049,5 +1057,7 @@ class Post extends Base {
 				'is_thread_reply' => false,
 			)
 		);
+
+		return \is_array( $filtered ) ? $filtered : $record;
 	}
 }
--- a/includes/wp-admin/class-admin.php
+++ b/includes/wp-admin/class-admin.php
@@ -364,7 +364,42 @@ class Admin {
 			return '';
 		}
 
-		\wp_redirect( $auth_url ); // phpcs:ignore WordPress.Security.SafeRedirect.wp_redirect_wp_redirect
+		/*
+		 * `$auth_url` is built from the auth-server metadata returned
+		 * by the resolution chain. The resolver validates each URL it
+		 * persists, but defence-in-depth: re-check the scheme + host
+		 * before redirecting an admin so a misconfigured filter or
+		 * future code path can't slip a `javascript:` / `data:` URI
+		 * through.
+		 *
+		 * `wp_safe_redirect` would normally reject this destination —
+		 * it's intentionally off-site (the AT Protocol auth server).
+		 * Add the auth-server host to `allowed_redirect_hosts` for the
+		 * lifetime of this dying request so `wp_safe_redirect` lets
+		 * the redirect through; the `exit` below means the filter
+		 * never affects any subsequent request.
+		 */
+		$auth_host   = \is_string( $auth_url ) ? \wp_parse_url( $auth_url, PHP_URL_HOST ) : '';
+		$auth_scheme = \is_string( $auth_url ) ? \wp_parse_url( $auth_url, PHP_URL_SCHEME ) : '';
+
+		if ( empty( $auth_host ) || 'https' !== $auth_scheme ) {
+			\add_settings_error(
+				'atmosphere',
+				'auth_failed',
+				\__( 'Authorization URL is not a safe HTTPS target.', 'atmosphere' )
+			);
+			return '';
+		}
+
+		\add_filter(
+			'allowed_redirect_hosts',
+			static function ( $hosts ) use ( $auth_host ) {
+				$hosts[] = $auth_host;
+				return $hosts;
+			}
+		);
+
+		\wp_safe_redirect( $auth_url );
 		exit;
 	}
 
@@ -771,9 +806,13 @@ class Admin {
 			'grant_types'                => array( 'authorization_code', 'refresh_token' ),
 			'response_types'             => array( 'code' ),
 			'token_endpoint_auth_method' => 'none',
-			// MUST match the scope string requested by Client::authorize().
-			// The auth server validates the request scope against the metadata;
-			// a drift here silently downgrades to the smaller of the two.
+
+			/*
+			 * MUST match the scope string requested by
+			 * Client::authorize(). The auth server validates the
+			 * request scope against the metadata; a drift here
+			 * silently downgrades to the smaller of the two.
+			 */
 			'scope'                      => 'atproto transition:generic identity:handle',
 			'dpop_bound_access_tokens'   => true,
 			'application_type'           => 'web',
@@ -782,9 +821,21 @@ class Admin {
 		/**
 		 * Filters the OAuth client metadata served at the REST endpoint.
 		 *
+		 * Filters MUST return an array containing at minimum the
+		 * `client_id` and `redirect_uris` keys; otherwise the original
+		 * unfiltered metadata is served so the OAuth flow keeps
+		 * working even with a misbehaving listener.
+		 *
 		 * @param array $metadata Client metadata.
 		 */
-		$metadata = \apply_filters( 'atmosphere_client_metadata', $metadata );
+		$filtered = \apply_filters( 'atmosphere_client_metadata', $metadata );
+
+		if ( \is_array( $filtered )
+			&& ! empty( $filtered['client_id'] )
+			&& ! empty( $filtered['redirect_uris'] )
+		) {
+			$metadata = $filtered;
+		}
 
 		$response = new \WP_REST_Response( $metadata, 200 );
 
--- a/includes/wp-admin/class-admin.php
+++ b/includes/wp-admin/class-admin.php
@@ -375,9 +375,11 @@ class Admin {
 		 * `wp_safe_redirect` would normally reject this destination —
 		 * it's intentionally off-site (the AT Protocol auth server).
 		 * Add the auth-server host to `allowed_redirect_hosts` for the
-		 * lifetime of this dying request so `wp_safe_redirect` lets
-		 * the redirect through; the `exit` below means the filter
-		 * never affects any subsequent request.
+		 * `wp_safe_redirect` call, then immediately detach the filter
+		 * so it can't affect any subsequent redirect — the `exit`
+		 * makes that production-redundant, but pinning the invariant
+		 * here keeps it intact if a test or a `wp_die()` handler ever
+		 * intercepts the redirect before `exit` fires.
 		 */
 		$auth_host   = \is_string( $auth_url ) ? \wp_parse_url( $auth_url, PHP_URL_HOST ) : '';
 		$auth_scheme = \is_string( $auth_url ) ? \wp_parse_url( $auth_url, PHP_URL_SCHEME ) : '';
@@ -391,15 +393,14 @@ class Admin {
 			return '';
 		}
 
-		\add_filter(
-			'allowed_redirect_hosts',
-			static function ( $hosts ) use ( $auth_host ) {
-				$hosts[] = $auth_host;
-				return $hosts;
-			}
-		);
+		$allow_auth_host = static function ( $hosts ) use ( $auth_host ) {
+			$hosts[] = $auth_host;
+			return $hosts;
+		};
 
+		\add_filter( 'allowed_redirect_hosts', $allow_auth_host );
 		\wp_safe_redirect( $auth_url );
+		\remove_filter( 'allowed_redirect_hosts', $allow_auth_host );
 		exit;
 	}
 
--- a/includes/wp-admin/class-admin.php
+++ b/includes/wp-admin/class-admin.php
@@ -847,6 +847,19 @@ class Admin {
 
 		if ( self::client_metadata_filter_is_valid( $filtered ) ) {
 			$metadata = $filtered;
+		} elseif ( $filtered !== $metadata ) {
+			/*
+			 * Surface only when the filter actually fired and returned
+			 * something that failed validation — without this guard
+			 * every page load on a site with no filter would trip the
+			 * notice because the equality check above is the cheap
+			 * shorthand for "nothing changed".
+			 */
+			\_doing_it_wrong(
+				__METHOD__,
+				\esc_html__( 'atmosphere_client_metadata must return an array with a non-empty string client_id and a redirect_uris list of admin URLs; falling back to the unfiltered metadata.', 'atmosphere' ),
+				'0.1.0'
+			);
 		}
 
 		$response = new \WP_REST_Response( $metadata, 200 );
--- a/includes/wp-admin/class-admin.php
+++ b/includes/wp-admin/class-admin.php
@@ -821,17 +821,21 @@ class Admin {
 		/**
 		 * Filters the OAuth client metadata served at the REST endpoint.
 		 *
-		 * Filters MUST return an array containing at minimum the
-		 * `client_id` and `redirect_uris` keys; otherwise the original
-		 * unfiltered metadata is served so the OAuth flow keeps
-		 * working even with a misbehaving listener.
+		 * Filters MUST return an array containing a non-empty
+		 * `client_id` string and a non-empty `redirect_uris` array;
+		 * otherwise the original unfiltered metadata is served so the
+		 * OAuth flow keeps working even with a misbehaving listener.
 		 *
 		 * @param array $metadata Client metadata.
 		 */
 		$filtered = \apply_filters( 'atmosphere_client_metadata', $metadata );
 
 		if ( \is_array( $filtered )
-			&& ! empty( $filtered['client_id'] )
+			&& isset( $filtered['client_id'] )
+			&& \is_string( $filtered['client_id'] )
+			&& '' !== $filtered['client_id']
+			&& isset( $filtered['redirect_uris'] )
+			&& \is_array( $filtered['redirect_uris'] )
 			&& ! empty( $filtered['redirect_uris'] )
 		) {
 			$metadata = $filtered;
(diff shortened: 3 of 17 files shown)


tests/phpunit/tests/oauth/class-test-redirect-uri-filter.php
```diff
@@ -0,0 +1,121 @@
...
+	/**
+	 * A filter that points at an external host is rejected; the
+	 * default admin URL wins.
+	 */
+	public function test_external_url_filter_is_rejected() {
+		\add_filter(
+			'atmosphere_oauth_redirect_uri',
+			static fn() => 'https://evil.example.com/oauth/callback'
+		);
+
```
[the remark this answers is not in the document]
pfefferle: Dismissed. PHP allows extra positional arguments to closures at the language level — verified at runtime: `php -r '$fn = static fn() => 42; echo $fn("extra");'` → outputs `42` cleanly. WordPress also slices the args to `accepted_args` (default 1) inside `WP_Hook::apply_filters`, so a 1-arg `static fn() => ...` callback only ever sees ≤1 argument anyway. No `ArgumentCountError`.

tests/phpunit/tests/transformer/class-test-transform-filter-validation.php
```diff
@@ -0,0 +1,160 @@
...
+				'post_content' => 'Body.',
+				'post_status'  => 'publish',
+			)
+		);
+		$post    = \get_post( $post_id );
+
+		\add_filter( 'atmosphere_transform_bsky_post', static fn() => 'not-an-array' );
+
+		$transformer = new Post( $post );
+		$record      = $transformer->transform();
```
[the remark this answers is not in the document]
pfefferle: Same as the redirect_uri test reply — dismissed. PHP allows extra positional arguments to closures, and WordPress's `apply_filters` slices args to `accepted_args` (default 1). Confirmed at runtime.

includes/wp-admin/class-admin.php
```diff
@@ -782,9 +821,21 @@ public static function serve_client_metadata(): \WP_REST_Response {
...
+		 *
 		 * @param array $metadata Client metadata.
 		 */
-		$metadata = \apply_filters( 'atmosphere_client_metadata', $metadata );
+		$filtered = \apply_filters( 'atmosphere_client_metadata', $metadata );
+
+		if ( \is_array( $filtered )
+			&& ! empty( $filtered['client_id'] )
+			&& ! empty( $filtered['redirect_uris'] )
+		) {
```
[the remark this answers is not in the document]
pfefferle: Fixed in 19016ab. Validation now requires `client_id` to be a non-empty **string** AND `redirect_uris` to be a non-empty **array** (previously both were just `!empty()` checks — a scalar `redirect_uris` would have passed). Five new regression tests in `tests/phpunit/tests/wp-admin/class-test-client-metadata-filter.php` cover the default well-formedness, scalar `redirect_uris` rejection, non-string `client_id` rejection, non-array filter rejection, and the happy-path filter use case.

uninstall.php
```diff
@@ -13,52 +13,112 @@
...
+ *  - atmosphere_last_seen_own_<col> — reaction-sync watermarks per
+ *                                     collection (likes, reposts, posts).
+ *
+ * Direct DB delete is the only way to wildcard-match these. The
+ * `_transient_` / `_transient_timeout_` rows for any object-cache-backed
+ * site fall through to the DB anyway, since uninstall runs without the
+ * plugin loaded.
+ */
+$wpdb->query( "DELETE FROM {$wpdb->options} WHERE option_name LIKE '\_transient\_atmo\_dpop\_nonce\_%' OR option_name LIKE '\_transient\_timeout\_atmo\_dpop\_nonce\_%'" ); // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery,WordPress.DB.DirectDatabaseQuery.NoCaching,WordPress.DB.DirectDatabaseQuery.SchemaChange
+$wpdb->query( "DELETE FROM {$wpdb->options} WHERE option_name LIKE 'atmosphere\_last\_seen\_own\_%'" ); // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery,WordPress.DB.DirectDatabaseQuery.NoCaching,WordPress.DB.DirectDatabaseQuery.SchemaChange
```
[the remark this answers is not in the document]
pfefferle: Fixed in 19016ab. Wildcard sweep now does `SELECT option_name FROM {options} WHERE option_name LIKE ...` first, then loops `delete_transient()` / `delete_option()` per matching row. That path invalidates the object cache (Redis, Memcached, batcache, `alloptions`) alongside the underlying `wp_options` row, so a subsequent reinstall can no longer read stale DPoP nonces or reaction-sync watermarks back from the cache.

pfefferle:
Thanks for the review. Addressed:

- **`class-admin.php`** (`redirect_uris` type-check): fixed in 19016ab with 5 new regression tests.
- **`uninstall.php`** (object-cache invalidation in the wildcard sweep): fixed in 19016ab.
- **`class-test-transform-filter-validation.php`** (Publication constructor): fixed in 80e3a5f.

Dismissed the 7 findings about filter-callback arity (`static fn() => null` will fatal under PHP 8.2). PHP allows extra positional arguments to closures at the language level — verified at runtime: `php -r '$fn = static fn() => 42; echo $fn("extra");'` outputs `42` cleanly. WordPress also `array_slice`s args to the declared `accepted_args` (default 1) inside `WP_Hook::apply_filters`, so a 1-arg `static fn() => ...` callback only ever sees ≤1 argument anyway.

includes/wp-admin/class-admin.php
```diff
@@ -782,9 +821,25 @@ public static function serve_client_metadata(): \WP_REST_Response {
...
-		$metadata = \apply_filters( 'atmosphere_client_metadata', $metadata );
+		$filtered = \apply_filters( 'atmosphere_client_metadata', $metadata );
+
+		if ( \is_array( $filtered )
+			&& isset( $filtered['client_id'] )
+			&& \is_string( $filtered['client_id'] )
+			&& '' !== $filtered['client_id']
+			&& isset( $filtered['redirect_uris'] )
+			&& \is_array( $filtered['redirect_uris'] )
+			&& ! empty( $filtered['redirect_uris'] )
```
kraftbj: **[HIGH] `redirect_uris` is validated as a container, not as content.**

The check requires `is_array && !empty` on the outer array. It accepts:
- `['client_id' => 'x', 'redirect_uris' => ['']]` — empty-string URI, still passes `!empty` because the inner array has one element.
- `['client_id' => 'x', 'redirect_uris' => ['https://evil.example/cb']]` — off-site URI.
- `['client_id' => 'x', 'redirect_uris' => [null]]` or `[['nested']]` — wrong types.
- An associative array that JSON-encodes as an object, not a list.

This matters more than it might look. `Client::redirect_uri()` (ATM-005) gates the inbound filter against `admin_url('')`, but here that gate is bypassed: the public metadata endpoint advertises `token_endpoint_auth_method: 'none'`. With a public client and an attacker-supplied `redirect_uris` entry in the metadata, an attacker initiates an OAuth flow using this site's `client_id` + their own PKCE challenge + the published off-site redirect URI, and the auth server happily redirects the code to them.

**Suggested fix:** apply the same gate to every entry that `Client::redirect_uri()` uses on its filter:

```php
if (
    ! \is_array( $filtered )
    || ! isset( $filtered['client_id'] )
    || ! \is_string( $filtered['client_id'] )
    || $filtered['client_id'] === ''
    || ! isset( $filtered['redirect_uris'] )
    || ! \is_array( $filtered['redirect_uris'] )
    || $filtered['redirect_uris'] === array()
) {
    $filtered = $metadata;
} else {
    $admin_prefix = \admin_url( '' );
    foreach ( $filtered['redirect_uris'] as $uri ) {
        if ( ! \is_string( $uri ) || $uri === '' || ! \str_starts_with( $uri, $admin_prefix ) ) {
            $filtered = $metadata;
            break;
        }
    }
}
```

Also pin `client_id` against `Client::client_id()` (the metadata `client_id` should match what the OAuth flow actually uses). Add tests for: off-site entries, mixed valid/invalid, associative arrays, non-string entries, empty-string entries.

Flagged by codex (HIGH), the multi-persona security reviewer (P3, conf 60), and the adversarial reviewer (conf 95). Triple cross-corroboration.

uninstall.php
```diff
@@ -13,52 +13,138 @@
...
+
+foreach ( $atmosphere_options as $atmosphere_option ) {
+	delete_option( $atmosphere_option );
+}
 
 // Remove scheduled events via the canonical helper.
 clear_scheduled_hooks();
 
-// Remove post meta.
 global $wpdb;
```
kraftbj, line 43: **[HIGH] Network-active uninstall only cleans the current site.**

Every operation in this file runs against the current blog: `delete_option()` is per-site by default, the `$wpdb->postmeta / commentmeta / options` table references resolve to the current blog's tables, and `wp_clear_scheduled_hook()` is per-site. There's no `is_multisite()` / `get_sites()` / `switch_to_blog()` anywhere.

On a network-active install, deleting the plugin leaves every other blog with its full state intact: `atmosphere_connection` (which contains the encrypted access token, refresh token, and DPoP private key), publication options, `_atmosphere_bsky_*` and `_atmosphere_doc_*` post-meta, comment-meta keyed on `protocol = 'atproto'`, and scheduled hooks. The encrypted material is decryptable as long as `AUTH_KEY` survives in `wp-config.php` (which it does — uninstall doesn't touch that).

**Suggested fix:** extract the cleanup body into a `atmosphere_uninstall_site()` function and gate the call:

```php
function atmosphere_uninstall_site() {
    /* ... existing body ... */
}

if ( \is_multisite() ) {
    foreach ( \get_sites( array( 'fields' => 'ids' ) ) as $blog_id ) {
        \switch_to_blog( $blog_id );
        atmosphere_uninstall_site();
        \restore_current_blog();
    }
} else {
    atmosphere_uninstall_site();
}
```

**Important caveat:** the rest of the plugin has no multisite handling either (no `is_multisite` / `switch_to_blog` calls anywhere in `includes/`). If multisite is intentionally unsupported, this finding is reframed as "document the constraint in `readme.txt`." If it's intended to be supported, the uninstall fix alone is the wrong unit of work — the broader audit needs to happen.

Flagged by codex (HIGH) and the adversarial reviewer (medium, conf 95).

includes/transformer/class-facet.php
```diff
@@ -188,17 +198,27 @@ private static function hashtags( string $text ): array {
...
-	 * @return string DID string.
+	 * @return string DID string, or empty string if the handle is malformed.
 	 */
 	private static function resolve_mention( string $handle ): string {
 		$conn = get_connection();
 		if ( ! empty( $conn['handle'] ) && \strtolower( $handle ) === \strtolower( $conn['handle'] ) ) {
 			return $conn['did'];
 		}
 
+		if ( ! self::is_valid_handle( $handle ) ) {
```
kraftbj, line 227: **[HIGH] The defence-in-depth claim only covers malformed handles, not the threat the docblock describes.**

Lines 201-211 describe the threat as "a post containing `@attacker.com.<exfil-data>` produces a slow but reliable side-channel for exfiltrating data via subdomain DNS." The added gate (`is_valid_handle()` on line 218) only rejects syntactically malformed handles. A regular Bluesky-style handle pointing at an attacker-controlled valid domain — `@user.attacker.example`, or `@anything.something-the-attacker-owns.tld` — passes the syntax check and reaches `dns_get_record()` on line 222.

The entry path here matters. `Facet::extract()` is called during publishing of any post AND any comment (`Atmosphere::should_publish_comment()` → `Comment` transformer → `Facet::extract()`). Commenter input is **untrusted by definition**. A commenter typing `@anything.attacker.example` triggers a server-side DNS lookup to attacker-controlled infrastructure on every publish.

Secondary issue: `dns_get_record()` has no PHP-level timeout. The system resolver default (5s × retries) can stall the publishing path under a slow / hostile authoritative DNS server. Per mention, in cron.

**Suggested fix:** decide the threat model first.
- If post-author mentions (admin / editor input) are trusted but commenter mentions aren't, gate `Facet::extract()` differently on the comment path — either skip mention resolution entirely or require an allowlist.
- If all mention resolution is trusted: rewrite the docblock to say so and drop the side-channel framing.
- Either way, set an explicit DNS timeout. PHP's `dns_get_record()` doesn't accept one directly; if it matters, switch to DoH over `wp_safe_remote_get()` with a 2s timeout and cache the result in a transient.

Also: line 237's docblock says "mirroring `Resolver::is_valid_handle()`" — that method is `private static` on Resolver. A future reader will go looking for a shared invariant and find an implicit copy. See the cross-PR coordination note below.

Flagged by codex (MEDIUM, treating the threat model). The multi-persona pass surfaced the same DNS-timeout concern from a reliability angle.

uninstall.php
```diff
@@ -13,52 +13,138 @@
...
+ * underlying `wp_options` row. A plain SQL `DELETE` against the
+ * options table would leave stale values in any persistent cache,
+ * so a subsequent reinstall could read them back.
+ */
+// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery,WordPress.DB.DirectDatabaseQuery.NoCaching
+$atmosphere_transient_rows = $wpdb->get_col(
+	"SELECT option_name FROM {$wpdb->options}
+	 WHERE option_name LIKE '\_transient\_atmo\_dpop\_nonce\_%'
+	    OR option_name LIKE '\_transient\_timeout\_atmo\_dpop\_nonce\_%'"
+);
```
kraftbj, line 155: On sites with a persistent object cache (Redis, Memcached), `set_transient()` writes to the object cache and never touches `wp_options`. The `SELECT option_name FROM {$wpdb->options} WHERE option_name LIKE '_transient_atmo_dpop_nonce_%'` query returns zero rows on those installs, the `foreach` never runs, and the DPoP nonces persist in the object cache after uninstall.

Also missing from the sweep entirely: `atmosphere_profile_<md5(did)>` transients written by `Reaction_Sync` (see `includes/class-reaction-sync.php` around the profile fetcher). No LIKE pattern matches them.

**Suggested fix:** the cleanest path is to track the dynamic transient keys when they're written (a small `set_transient + register-for-cleanup` helper), or accept that object-cache-only transients are residual and call them out in the cleanup docblock. Add `atmosphere_profile_*` to the sweep regardless — that's a flat omission, not a backend-dependent one.

uninstall.php
```diff
@@ -13,52 +13,138 @@
...
+ *
+ * Query for matching option names first, then route deletion through
+ * `delete_transient()` / `delete_option()` so the object cache
+ * (Redis, Memcached, `alloptions`) is invalidated alongside the
+ * underlying `wp_options` row. A plain SQL `DELETE` against the
+ * options table would leave stale values in any persistent cache,
+ * so a subsequent reinstall could read them back.
+ */
+// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery,WordPress.DB.DirectDatabaseQuery.NoCaching
+$atmosphere_transient_rows = $wpdb->get_col(
```
kraftbj, line 149: The wildcard patterns `atmo_dpop_nonce_%` and `atmosphere_last_seen_own_%` are string literals here, but they're already defined as class constants elsewhere (`Nonce_Storage::PREFIX`, `Reaction_Sync::OPTION_LAST_SEEN_OWN_PREFIX`). A future rename of either constant silently leaves orphaned rows that this sweep won't find.

Two options:
1. Bootstrap the autoloader in `uninstall.php` (it normally runs pre-bootstrap) and reference the constants directly. Adds complexity but eliminates the drift class.
2. Leave the literals and add a comment grep-anchored to the constant name, e.g. `// MUST stay in lock-step with Nonce_Storage::PREFIX.` Then anyone changing the constant grepping its name finds this file.

The second is the right call here unless the autoloader bootstrap is trivial.

includes/wp-admin/class-admin.php
```diff
@@ -364,7 +364,42 @@ public static function sanitize_handle( $value ): string {
 			return '';
 		}
 
-		\wp_redirect( $auth_url ); // phpcs:ignore WordPress.Security.SafeRedirect.wp_redirect_wp_redirect
+		/*
```
kraftbj, line 367: The PR body says: > *ATM-004 test coverage — `Admin::sanitize_handle()` redirects + exits, which makes it untestable in PHPUnit without a small refactor. Doable; out of scope for this PR.*

It's actually testable today. `tests/phpunit/tests/class-test-admin-handle.php:155-183` already uses the pattern: filter `wp_redirect` to capture the URL, then catch `WPDieException` when the production code calls `exit`. Same shape works here.

Three tests that would fit:
1. `auth_url` with `http://...` scheme → `settings_error()` is set, no redirect fires.
2. `auth_url` with valid `https://example.com/authorize` → `wp_safe_redirect` is called AND the auth-server host appears in `allowed_redirect_hosts` during the call (verify via `add_filter`).
3. `Client::authorize()` returns `WP_Error` → `settings_error()` is set, no redirect fires.

Not a blocker — flagging because the maintainer note overstates the difficulty. The pattern is already in the test suite.

includes/wp-admin/class-admin.php
```diff
@@ -364,7 +364,42 @@ public static function sanitize_handle( $value ): string {
...
+		if ( empty( $auth_host ) || 'https' !== $auth_scheme ) {
+			\add_settings_error(
+				'atmosphere',
+				'auth_failed',
+				\__( 'Authorization URL is not a safe HTTPS target.', 'atmosphere' )
+			);
+			return '';
+		}
+
+		\add_filter(
```
kraftbj: The anonymous closure is attached to `allowed_redirect_hosts` and never explicitly removed. The block comment above (lines 367-381) says the immediately-following `exit` guarantees no leakage. That's accurate in production. It's also untested.

If a future test (or a `wp_die()` handler in unusual configurations) intercepts the redirect, the closure persists for the rest of the request. The destination is the OAuth auth server, which is benign, but any subsequent `wp_safe_redirect()` in the same request would also accept that host.

Low-cost belt-and-suspenders:

```php
$allow_auth_host = static function ( $hosts ) use ( $auth_host ) {
    $hosts[] = $auth_host;
    return $hosts;
};
\add_filter( 'allowed_redirect_hosts', $allow_auth_host );
\wp_safe_redirect( $auth_url );
\remove_filter( 'allowed_redirect_hosts', $allow_auth_host );
exit;
```

The extra `remove_filter` is dead-code-equivalent in production (because of `exit`), but pins the invariant in a way that survives a future test that suppresses the exit.

includes/transformer/class-post.php
```diff
@@ -187,6 +191,8 @@ public function transform(): array {
 				'is_thread_reply' => false,
 			)
 		);
+
+		return \is_array( $filtered ) ? $filtered : $record;
```
kraftbj: Silent fallback: when the filter returns a non-array, the code reverts to `$record` and proceeds. No log, no `_doing_it_wrong`, no admin notice. The developer who wrote a buggy filter discovers the no-op only when they read back the published AT Protocol record and notice their changes aren't there.

Same pattern repeats in `class-comment.php`, `class-document.php`, `class-publication.php`, and `class-admin.php`'s `serve_client_metadata`. WordPress's idiom for this case is `_doing_it_wrong()` — visible in `WP_DEBUG`, silent otherwise, zero runtime cost. The plugin already uses it elsewhere (`class-post.php:785, 807`).

```php
if ( ! \is_array( $filtered ) ) {
    \_doing_it_wrong( __METHOD__, \esc_html__( 'atmosphere_transform_post must return an array; falling back to the unfiltered record.', 'atmosphere' ), '0.1.0' );
    $filtered = $record;
}
```

One comment per file, leaving the other transformer files to follow the same pattern.

includes/transformer/class-facet.php
```diff
@@ -211,4 +231,24 @@ private static function resolve_mention( string $handle ): string {
...
+	 * RFC 1035-style DNS-name validation, mirroring
+	 * `Resolver::is_valid_handle()`. Rejects empty strings, oversized
+	 * labels, leading/trailing hyphens, single-label hosts, and any
+	 * character outside `[A-Za-z0-9-]` — including percent-encoded
+	 * forms.
+	 *
+	 * @param string $host Handle to validate.
+	 * @return bool
+	 */
+	private static function is_valid_handle( string $host ): bool {
```
kraftbj, line 254: Cross-PR coordination — `class-resolver.php:318` in PR 61 already has a byte-for-byte copy of this regex / length cap / label rule. The PR body flags consolidation as a follow-up.

Landing both PRs together is the right window. Cheapest option: bump `Resolver::is_valid_handle()` and `Resolver::is_safe_https_url()` to `public static` in PR 61, then this method here becomes:

```php
private static function is_valid_handle( string $host ): bool {
    return \Atmosphere\OAuth\Resolver::is_valid_handle( $host );
}
```

No new file, no namespace gymnastics. If either side later wants to tighten the regex (IDN support, lowercased label rule, the reserved-TLD list from PR 61's review), one change covers both. Without consolidation, drift between the mention path and the resolver path is silent — a handle that emits a mention facet but that the resolver rejects, or vice versa.

Also fixes the misleading docblock claim on line 237 — "mirroring `Resolver::is_valid_handle()`" stops being a false reference.

kraftbj (changes requested):
Same review pattern as PR 61 — opinionated multi-persona pass plus an independent codex adversarial run. Three highs land here; the Copilot dismissals stand on their own merits.

First, the things this PR gets right:
- Verified all three "addressed" claims at HEAD: `class-admin.php` client_metadata type-check (lines 831-842 + 5 regression tests), `uninstall.php` cache-aware wildcard sweep (lines 125-150 swapped `DELETE LIKE` for `SELECT` + per-row `delete_transient/delete_option`), Publication constructor (test file line 100). All correct.
- Verified the 7 dismissed Copilot arity findings: the defence is correct. PHP closures silently accept extra positional args, and `WP_Hook::apply_filters` `array_slice`s args to `accepted_args` (default 1) before invocation. Even with the filter being applied with `$record, $post` arguments, a 0-arg `static fn() => ...` callback sees zero arguments. Dismissal is defensible.
- ATM-004 (sanitize_handle redirect), ATM-005 (redirect_uri filter), ATM-006 (mention DNS regex gate against malformed handles) all behave correctly for the threats the audit named.

Three highs inline:
1. **`serve_client_metadata`** — the validator gates container shape only. A filter can return `redirect_uris => ['https://evil.example/cb']` or `redirect_uris => ['']` and the public metadata endpoint serves them. With `token_endpoint_auth_method: 'none'`, that's an unauthenticated public client an attacker can drive with this site's `client_id` and their own redirect URI.
2. **`uninstall.php`** — runs once in the current blog context. On a network-active install, the other blogs keep their `atmosphere_connection` (encrypted access tokens, refresh tokens, DPoP keys), post-meta, comment-meta, and cron hooks indefinitely. This is the credential-residue case the expanded cleanup is meant to address.
3. **`Facet::resolve_mention`** — the new "defence-in-depth side-channel" docstring overclaims. The gate blocks malformed handles. It does not block attacker-controlled valid domains. Any commenter typing `@attacker.example` triggers server-side `dns_get_record('_atproto.attacker.example', DNS_TXT)` every publish, with no allowlist and no timeout. If that's intentional, the docstring needs to say so; if not, the gate needs to grow.

Secondary items in line — object-cache transient leak under Redis/Memcached, the "untestable" claim on `sanitize_handle()` (there's a working pattern at `tests/phpunit/tests/class-test-admin-handle.php:155-183`), wildcard prefixes that should reference class constants, and the cross-PR `is_valid_handle()` consolidation.

For multisite — if multisite is a deliberate non-goal for this plugin today, say so in `readme.txt` / docs and the uninstall gap becomes "document it." Otherwise the fix here lands in a partially-broken state since the rest of the plugin also has no multisite handling.

includes/wp-admin/class-admin.php
pfefferle: Fixed in bef18d3. Validation extracted into `client_metadata_filter_is_valid()`. Container shape unchanged; every `redirect_uris` entry now has to be a non-empty string beginning with `admin_url('')`, matching the gate `Client::redirect_uri()` already applied to the inbound filter. Any failing entry disqualifies the entire filter result.

Four new regression tests in `class-test-client-metadata-filter.php`:
 * Off-site entry rejected.
 * Mixed valid + invalid disqualifies the whole filter return.
 * Empty-string entry rejected (would have passed `!empty` because the parent array had one element).
 * Non-string entries (null, integer, nested array, bool) rejected.

uninstall.php
pfefferle: Resolved via documentation in 1d16da3. As you noted, the rest of the plugin makes no multisite assumptions either — no `is_multisite()` / `switch_to_blog()` anywhere in `includes/`. Fixing only uninstall would land in a partially-broken state.

Took path B: added a "Does ATmosphere support WordPress Multisite?" FAQ entry to `readme.txt` declaring multisite unsupported and calling out the per-site uninstall behaviour explicitly. If/when multisite support becomes a real product goal, the broader audit gets to happen and uninstall can be wrapped in `get_sites()` then. The agent-doc update in this PR codifies the "code OR readme, not neither" rule.

includes/transformer/class-facet.php
pfefferle: Tightened the docblock in d163248. Rewrote it to admit the gate's actual scope: it rejects syntactically malformed handles (percent-encoded segments, control chars from a future regex relaxation), but does NOT block DNS lookups against attacker-controlled valid domains. Any commenter typing `@user.attacker.example` still triggers DNS to attacker-controlled infrastructure — that's by design, not a bug the syntactic gate covers.

Called out what an actual mitigation would look like (commenter-path gating, allowlist, DoH-with-timeout) so future work picks the right layer. No behaviour change. The agent-doc update in this PR codifies "DNS egress via mention resolution is a threat-model question, not a regex question."

uninstall.php
pfefferle: Fixed in d35b174.

 * `atmosphere_profile_<md5(did)>` added to the LIKE query — flat omission, not backend-dependent. Both transient + transient_timeout patterns covered.
 * Persistent-object-cache caveat documented explicitly. Audited the residual: DPoP nonces are server-issued opaque tokens (no secret material), profile cache is public data, TTLs are short (5 min / 1 hr). Declined the write-time registry pattern because the residual is acceptable today, but the docblock spells out the trigger for revisiting ("longer-lived dynamic-key transient with secret material").

uninstall.php
pfefferle: Fixed in 72ec6e0 via option 2 (grep-anchored docblock). Each dynamic-key wildcard now has a "MUST stay in lock-step with `X::Y`" line naming the source constant, so a future rename via grep finds both sites. Chose this over bootstrapping the autoloader because `uninstall.php` runs pre-bootstrap and the autoloader load is non-trivial complexity for one-shot deletion code.

includes/wp-admin/class-admin.php
pfefferle: You're right — the PR body's "untestable without refactor" overstates the difficulty. Added the three tests in 92d6fd7 using the same `WPDieException` + `wp_redirect` filter pattern from `class-test-admin-handle.php:155-183`:

 * `test_http_auth_url_surfaces_error_and_does_not_redirect` — stubbed resolver advertises `http://`, asserts settings error + no redirect.
 * `test_wp_error_from_authorize_surfaces_error_and_does_not_redirect` — well-known 404, asserts settings error + no redirect.
 * `test_valid_auth_url_redirects_with_scoped_allowed_redirect_hosts` — happy path, asserts redirect target + auth-server host present in `allowed_redirect_hosts` at redirect time.

The third one also pins the scoped-filter fix from comment 3256079809 so the closure can't leak past the call.

includes/wp-admin/class-admin.php
pfefferle: Fixed in 14d966f. Closure stored in a local, `remove_filter()` called immediately after `wp_safe_redirect()`. Production behaviour is unchanged (the `exit` made the cleanup dead-code-equivalent), but the new test `test_valid_auth_url_redirects_with_scoped_allowed_redirect_hosts` in 92d6fd7 pins the invariant — if anyone later intercepts the redirect in a test, the filter won't leak into subsequent same-request redirects.

includes/transformer/class-post.php
pfefferle: Fixed in a16dd71 across all five sites: `atmosphere_transform_bsky_post` (both call sites in `class-post.php`), `atmosphere_transform_comment`, `atmosphere_transform_document`, `atmosphere_transform_publication`, and `atmosphere_client_metadata`. Each now calls `\_doing_it_wrong()` with a specific message before falling back.

`serve_client_metadata` gates the notice on `$filtered !== $metadata` to avoid tripping it on every page load when no filter is registered (since `apply_filters` returns the input unchanged in that case). The agent-doc update in this PR codifies "_doing_it_wrong on silent filter fallback" as a review check.

includes/transformer/class-facet.php
pfefferle: Deferring to a follow-up after #61 lands. The `public static` promotion you suggested (and which I committed on #61 — 5ced865) is on `fix/security-audit-highs`, not in `trunk`. Doing the consolidation here would either (a) require rebasing this PR onto the highs branch (couples them operationally) or (b) duplicate the visibility change in this PR (merge conflicts).

Cleanest: merge #61 → trunk first, then a one-line follow-up here flips `Facet::is_valid_handle()` to `return Resolver::is_valid_handle( $host );` and drops the local copy. I'll open that PR the moment #61 merges so the consolidation doesn't drift.

kraftbj (approved):

PR #63, opened 2026-05-18
Fix non-public post federation cleanup

pfefferle:
Fixes #

## Proposed changes:

* Prevent password-protected and otherwise non-public posts from publishing AT Protocol records.
* Delete existing PDS records when a previously public post becomes protected, private, trashed, unsupported, or otherwise non-public.
* Republish posts that become public again after visibility cleanup, and update existing records when stale publish/delete cron events fire after a restore.
* Add defense-in-depth transformer redaction, filter return-shape validation, and local security guidance for future visibility audits.

### Other information:

## Testing instructions:

* Run `composer lint`.
* Run `npm run env-test`.
* Optional manual check: publish a post, confirm AT records are created, add a post password, run scheduled events, and confirm the remote records are deleted.
* Optional manual check: remove the password after cleanup, save the post again, run scheduled events, and confirm fresh AT records are created.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>


includes/class-atmosphere.php
```diff
@@ -297,48 +305,43 @@ public function on_status_change( string $new_status, string $old_status, \WP_Po
...
-			$bsky_tid = \get_post_meta( $post->ID, Transformer\Post::META_TID, true );
-			$doc_tid  = \get_post_meta( $post->ID, Transformer\Document::META_TID, true );
-			if ( ! $bsky_tid && ! $doc_tid ) {
-				// Unpublish of a post that was never synced — nothing to clean up.
-				return;
-			}
-		}
-
 		// Prevent infinite loops from meta updates.
 		if ( \did_action( 'atmosphere_publishing' ) ) {
```
kraftbj: **[P0] Bulk-protect of N posts in one request leaks N-1 of them.**

`did_action('atmosphere_publishing')` returns the total number of times the action has fired in the current PHP request, not whether it fired for *this* post. The flow:

1. Post 1: `did_action` returns 0, handler runs, fires the action at line 332, marks cleanup, schedules `atmosphere_delete_post`.
2. Posts 2..N (same request): `did_action` returns 1+, handler returns at line 329 *before* `mark_visibility_cleanup()` and `wp_schedule_single_event()`.

Triggers: Quick Edit bulk action, WP-CLI `wp post update` loops, privacy export, any plugin calling `wp_update_post` in a loop.

The PHPUnit suite hides this because `reset_publishing_action()` at `tests/phpunit/tests/class-test-atmosphere.php:75-77` does `unset($wp_actions['atmosphere_publishing'])` between every test. Production WordPress never resets the counter within a request.

This is precisely the leak class the PR exists to prevent — flagged by adversarial review at confidence 90.

**Suggested fix:** replace the per-request action counter with a per-post guard:

```php
static $in_progress = array();
if ( isset( $in_progress[ $post->ID ] ) ) {
    return;
}
$in_progress[ $post->ID ] = true;
```

Drop the `do_action('atmosphere_publishing')` line below since nothing else listens for it (verify with grep), or keep it as a public lifecycle hook but stop using it as the re-entry guard.

Add a regression test that calls `on_status_change` on two distinct posts in a single test method *without* `reset_publishing_action()` between them, asserting both schedule their expected cron events. Today's behaviour fails that test.

includes/class-publisher.php
```diff
@@ -94,6 +94,17 @@ public static function delete( \WP_Post|\WP_Comment $object ): array|\WP_Error {
 	 * @return array|\WP_Error applyWrites response(s) or error.
 	 */
 	public static function publish_post( \WP_Post $post ): array|\WP_Error {
+		if ( ! is_post_publishable( $post ) ) {
```
kraftbj, line 97: **[P1] TOCTOU during `applyWrites` in flight.**

The new gate checks visibility once on entry. If a post is password-protected mid-flight (concurrent request landing during the PDS round-trip), the in-flight publish still commits and `store_document_meta` / `mirror_thread_records_meta` persist the record meta.

Meanwhile the concurrent `on_status_change` handler runs against the still-records-less state (records are about to be stored by the in-flight publish). `has_post_records()` returns false, so the cleanup branch is never taken. Protected content stays live on the PDS.

The window is small but real, and cross-corroborated by codex and adversarial review.

**Suggested fix:** re-fetch `get_post( $post->ID )` immediately after `apply_writes_chunked` returns and *before* writing record meta. If the post became non-publishable, fire the cleanup path inline (call `delete_post` on the just-committed records using the API response's TIDs/CIDs directly) and skip the meta-store step. A second option is to keep the in-flight publish and *also* schedule a delete via `wp_schedule_single_event`, but that loses atomicity on a system that already has cleanup-race issues (see the `atmosphere_update_post` finding).

includes/class-atmosphere.php
```diff
@@ -727,8 +790,19 @@ static function ( int $post_id ): void {
 			'atmosphere_update_post',
 			static function ( int $post_id ): void {
 				$post = \get_post( $post_id );
-				if ( $post && 'publish' === $post->post_status && is_supported_post_type( $post->post_type ) ) {
-					Publisher::update_post( $post );
+				if ( ! $post ) {
+					return;
+				}
+				if ( is_post_publishable( $post ) ) {
+					$result = Publisher::update_post( $post );
```
kraftbj: **[P1] Cleanup-race resets state, leaving the post in limbo.**

The `atmosphere_publish_post` handler at line 774-784 routes between `update_post` and `publish_post` via `has_post_records`. The `atmosphere_update_post` handler here does not — it unconditionally calls `Publisher::update_post( $post )`.

Failure path:
1. Admin password-protects publish post → schedules `atmosphere_delete_post` + marks `META_VISIBILITY_CLEANUP`.
2. Cron pops the delete event, `Publisher::delete_post` runs (up to 30s HTTP).
3. Admin removes password mid-flight. `on_status_change` sees `is_publishable=true`, `has_records=true` (meta still present), schedules `atmosphere_update_post`. `wp_clear_scheduled_hook('atmosphere_delete_post')` is a no-op (already popped).
4. Delete completes, `clear_all_record_meta` wipes everything.
5. `atmosphere_update_post` fires here. `Publisher::update_post` sees empty `stored_thread_records` and empty `META_TID`, fires `atmosphere_update_skipped_unsynced_post`, returns `array()`. That's not a `WP_Error`, so line 800 clears `META_VISIBILITY_CLEANUP`.

Result: post is publishable, has no PDS records, has no marker, has no scheduled work. No automatic recovery.

**Suggested fix:** mirror the `has_post_records` branch from `atmosphere_publish_post`:

```php
if ( is_post_publishable( $post ) ) {
    $result = self::has_post_records( $post )
        ? Publisher::update_post( $post )
        : Publisher::publish_post( $post );
    // ...
}
```

Reliability reviewer hit this at confidence 75.

includes/class-atmosphere.php
```diff
@@ -727,8 +790,19 @@ static function ( int $post_id ): void {
...
+					if ( ! \is_wp_error( $result ) ) {
+						self::clear_visibility_cleanup_marker( $post );
+					}
+					return;
+				}
+				if ( self::has_post_records( $post ) ) {
+					self::mark_visibility_cleanup( $post );
+					Publisher::delete_post( $post );
```
kraftbj: **[P1] Silent `WP_Error` swallow on cleanup.**

This call (and the equivalent at line 825) discards `Publisher::delete_post`'s return value. `Publisher::delete_post` returns `WP_Error` on access-token failure, DPoP decrypt failure, network timeout, or PDS 4xx/5xx. Local meta is preserved (good) — but `wp_schedule_single_event` is one-shot, so transient failures leave PDS records live indefinitely with zero operator-visible signal.

Compare to the sibling `atmosphere_delete_records` handler around line 845 which `error_log`s on the same pattern.

Cross-corroborated by correctness, reliability, adversarial, and codex.

**Suggested fix:** mirror the `error_log` pattern from `atmosphere_delete_records`:

```php
$result = Publisher::delete_post( $post );
    \error_log( \sprintf(
        'ATmosphere: visibility-cleanup delete_post failed for post %d: %s',
        $post->ID,
        $result->get_error_message()
    ) );
}
```

A stronger fix also requeues a retry via `wp_schedule_single_event` with backoff, but logging alone closes the silent-failure gap.

includes/class-atmosphere.php
```diff
@@ -738,6 +812,16 @@ static function ( int $post_id ): void {
...
+						$result = Publisher::update_post( $post );
+						if ( ! \is_wp_error( $result ) ) {
+							self::clear_visibility_cleanup_marker( $post );
+						}
+						return;
+					}
+					if ( self::has_post_records( $post ) ) {
+						self::mark_visibility_cleanup( $post );
+					}
 					Publisher::delete_post( $post );
```
kraftbj: **[P1] Silent `WP_Error` swallow on cleanup** — same finding as line 805, in the `atmosphere_delete_post` handler. The fix needs to land at both sites.

tests/phpunit/tests/transformer/class-test-post.php
```diff
@@ -227,6 +227,96 @@ public function test_filter_receives_computed_default_and_post() {
...
+	 * @covers ::transform
+	 */
+	public function test_password_protected_transform_is_redacted() {
+		$post = self::factory()->post->create_and_get(
+			array(
+				'post_status'   => 'publish',
+				'post_title'    => 'CONFIDENTIAL-TITLE',
+				'post_content'  => 'CONFIDENTIAL-BODY',
+				'post_excerpt'  => 'CONFIDENTIAL-EXCERPT',
+				'post_password' => 'secret',
```
kraftbj, line 243: **[P1] No regression test for `post_password = '0'`.**

Commit `c934bfc Harden is_post_redacted password check` was written precisely because `empty('0') === true` would have made a literal-`'0'` password not-redacted. Every test in this file uses `'secret'`. A future revert to `empty()` would silently restore the leak with all tests passing.

Add a parallel case here (and at the equivalent sites in `class-test-document.php` and `class-test-functions.php`):

```php
'post_password' => '0',
```

Assert both `is_post_publishable()` returns false and the transformer output is redacted. The same regression class is the reason this PR exists.

includes/transformer/class-base.php
```diff
@@ -134,6 +134,28 @@ protected function get_excerpt( \WP_Post $post, int $word_limit = 30 ): string {
...
+	 * current request's unlock cookie.
+	 *
+	 * Intentionally narrower than `Atmosphere\is_post_publishable()`:
+	 * direct transformer callers may transform unsupported post types,
+	 * but protected/non-published fields still must not be serialized.
+	 *
+	 * @param \WP_Post $post Post object.
+	 * @return bool
+	 */
+	protected function is_post_redacted( \WP_Post $post ): bool {
```
kraftbj, line 151: **[P1] Non-publish-status branch of `is_post_redacted()` is untested.**

The helper has two branches: non-empty `post_password` (tested) and `'publish' !== get_post_status($post)` (not tested). The docblock claims direct transformer callers (which can hit `draft`/`private`/`trash` statuses) are protected; no test exercises that path.

Add a test that constructs `new Post( $draft_post )` (or `Document( $draft_post )`) directly and asserts the transformer output is redacted, separate from the publishability-gated path Publisher uses.

includes/transformer/class-post.php
```diff
@@ -110,6 +110,8 @@ class Post extends Base {
 	 * @return array app.bsky.feed.post record.
 	 */
 	public function transform(): array {
+		$redacted = $this->is_redacted();
```
kraftbj, line 129: **[P2] `atmosphere_transform_*` filter receives the raw `WP_Post` even on redacted records.**

The redaction logic in `transform()` correctly empties `text` / strips embed fields, then unconditionally calls `apply_filters('atmosphere_transform_bsky_post', $record, $this->object, $context)` at line 192. The filter receives the empty record AND the raw `WP_Post` with protected content intact.

A reasonable third-party filter pattern (`if (empty($record['text'])) $record['text'] = $post->post_excerpt;`) would silently unredact. Compare to `atmosphere_is_short_form_post` at line 138 which is explicitly gated `if (!$redacted)`.

**Suggested fix:** skip the filter on redaction (matching the short-form pattern), or scrub the filtered record after return.

Same pattern repeats in `class-document.php:141`. Cross-corroborated by correctness, adversarial, and codex.

includes/transformer/class-document.php
```diff
@@ -50,9 +50,16 @@ class Document extends Base {
...
+		/*
+		 * Redacted records are defense-in-depth output for authorized
+		 * previews/direct callers. Publisher rejects or deletes
+		 * non-publishable posts before this placeholder reaches the PDS.
+		 */
 		$record = array(
 			'$type'       => 'site.standard.document',
-			'title'       => sanitize_text( \get_the_title( $this->object ) ),
+			'title'       => $redacted ? '' : sanitize_text( \get_the_title( $this->object ) ),
 			'publishedAt' => $this->to_iso8601( $this->object->post_date_gmt ),
```
kraftbj: **[P2] `publishedAt` is emitted on redacted records.**

The redaction branch gates `path` / `description` / `coverImage` / `textContent` / `content` / `tags` / `bskyPostRef` / `updatedAt`, but emits `publishedAt`. For password-protected posts the publish date is already public via WordPress core. For `private` / `future` status, this leaks draft/scheduled timing through `Atmosphere::preview()` and any other direct caller of the redacted record.

Publisher gates above the transformer for the normal flow, so this is defence-in-depth only. If the design intent is "minimal placeholder," `publishedAt` should be redacted too.

kraftbj (changes requested):
Cross-corroborated review: opinionated multi-persona pass (correctness, security, testing, reliability, maintainability, project-standards, adversarial, previous-comments) plus an independent codex adversarial run. Both passes converged on the same blockers.

First, the things this PR gets right:
- `post_password = '0'` is correctly handled at every layer the PR introduces (`functions.php:205`, `class-base.php:151`). Copilot's inline comment is resolved by `c934bfc`.
- `is_post_publishable()` is a single namespaced helper reused across `Atmosphere`, `Publisher`, `Backfill`, and the new cron callbacks. One place to change, one place to mock.
- Cron callbacks re-check visibility at fire time via `get_post()` — a stale event queued before the post was protected does the right thing within a single cron tick.
- Reaction-sync filters out `has_password` posts when resolving inbound reactions.
- No new user-only surfaces — agent-native parity preserved.

One P0 + nine P1s. The P0 negates the security guarantee the PR exists to provide.

## P0 (must fix before merge)

1. `class-atmosphere.php:328` — bulk-protect of N posts in one PHP request leaks N-1 of them. `did_action('atmosphere_publishing')` is a per-request counter, not a per-post guard. After the first transition fires the action, every subsequent transition returns early at line 329, before `mark_visibility_cleanup()` and `wp_schedule_single_event('atmosphere_delete_post')`. Triggers: Quick Edit bulk action, WP-CLI loops, privacy export, any plugin calling `wp_update_post` in a loop. The PHPUnit fixture `reset_publishing_action()` at `class-test-atmosphere.php:75-77` masks this — production WordPress never resets the counter. Details inline.

## P1s anchored inline

2. `class-publisher.php:97` — TOCTOU during `applyWrites` in flight.
3. `class-atmosphere.php:797` — `atmosphere_update_post` handler doesn't branch on `has_post_records`; cleanup-race leaves publishable post with no records and nothing scheduled.
4. `class-atmosphere.php:805` and `class-atmosphere.php:825` — cleanup handlers swallow `Publisher::delete_post` `WP_Error` silently. PDS rate-limit / expired refresh-token / network blip leaves records live forever with no operator-visible signal.
5. `tests/phpunit/tests/transformer/class-test-post.php:243` — no regression test for `post_password = '0'`, the exact value `c934bfc` hardened. A future revert to `empty()` would silently restore the leak.
6. `class-base.php:151` — no test for the non-publish-status branch of `is_post_redacted()`.

## P1s off-diff (folded here)

7. **`class-atmosphere.php:607` — `is_comment_eligible()` does not check parent-post publishability.** Comments on a now-non-public parent post still federate. Between password-add and the queued `atmosphere_delete_post` firing, an approved comment writes a public `app.bsky.feed.post` reply pointing at the soon-to-be-deleted root. Cross-corroborated by three reviewers. Fix: add `is_post_publishable( get_post( (int) $comment->comment_post_ID ) )` to the gate, mirroring the `has_password=false` argument already added to `Reaction_Sync::find_post_by_bsky_uri`.

8. **No historical-leak cleanup.** Sites running ATmosphere before this PR have PDS records for posts that were password-protected pre-fix. The PR only reacts to future `transition_post_status` events. Without an admin-init migration or WP-CLI command, those leaks persist until each post is hand-re-saved. Changelog says "delete existing PDS records when a previously public post becomes protected" — that's only true going forward.

9. **`uninstall.php` does not mirror `_atmosphere_visibility_cleanup`.** Violates the project's own "Three-Way Symmetry" rule (`docs/php-coding-standards.md` → Cron-Specific Rules). Sibling `_atmosphere_publish_attempts` is present at `uninstall.php:80`; this new key is missing. One-line fix.

10. **`class-publisher.php:181-187` — ambiguous `applyWrites` failure orphans the bsky record.** `get_rkey()` persists bare post/doc TIDs before the API call. If PDS commits but WordPress receives an error, later visibility cleanup sees bare TIDs — but `stored_thread_records()` discards the bsky TID when `META_URI` is absent. `delete_post()` deletes only the doc TID and misses the committed bsky record. Pre-existing path, but visibility cleanup is the first feature that exercises it predictably.

## P2s anchored inline

11. `class-post.php:113` — `atmosphere_transform_*` filter receives raw `$this->object` after redaction, so a permissive subscriber can re-inject `post_title`/`post_content`/`post_excerpt`. Compare to `atmosphere_is_short_form_post` at `class-post.php:130-138` which IS explicitly gated `if (!$redacted)`.
12. `class-document.php:63` — `publishedAt` emitted on redacted document for `private`/`future` posts.

## Coverage

- 10 ce-code-review subagents + 1 codex challenge — 0 reviewer failures.
- Cross-reviewer corroboration on findings 1, 4, 7, 8, 10, 11.
- Two codex P1s without an internal twin (findings 2 TOCTOU, 10 orphan bsky) are both verifiable from the diff.

Findings 11+ can land in follow-up commits or a follow-up PR; nothing past P1 blocks.

Fix order: 1 → 7 → 4 → 9 → 5, 6 → 2, 3, 8, 10.

pfefferle:
Thanks for the cross-corroborated pass. All P0/P1/P2 findings addressed in `f887ea8` (P1 6 covered by the earlier `c934bfc`). Per-finding pointers:

**P0 1 — bulk-protect leak (`class-atmosphere.php:328`)**: replaced `did_action('atmosphere_publishing')` with a per-post static guard `self::$publishing_post_ids[ $post->ID ]`. The action still fires (`atmosphere_publishing`) as a public lifecycle hook but is no longer load-bearing for re-entry. New test exercises two posts in one request without `reset_publishing_action()`.

**P1 2 — TOCTOU during `applyWrites` (`class-publisher.php:97`)**: added `reconcile_post_after_write()` — after every successful publish/update path the post is re-fetched via `get_post()`, and if it lost publishability mid-flight the just-committed records are marked + deleted before the result returns.

**P1 3 — `atmosphere_update_post` no `has_post_records` branch (`class-atmosphere.php:797`)**: mirrored the publish-handler branching — `update_post()` when records exist or no cleanup marker, otherwise `publish_post()`. The same branching was applied to the `atmosphere_delete_post` handler's reconcile path.

**P1 4 / 5 — silent `WP_Error` swallow (`class-atmosphere.php:805` and `:825`)**: every `Publisher::delete_post()` call inside the cron handlers now routes through `log_cron_error()`. The helper was generalized from comment-only to take any `$object_id` + operation name.

**P1 6 — `post_password = '0'` regression test**: added cases in `class-test-functions.php`, `class-test-document.php`, and `class-test-post.php`. Hardening commit is `c934bfc`.

**P1 7 — non-publish-status branch untested (`class-base.php:151`)**: added `test_draft_transform_is_redacted` and `test_draft_document_is_redacted_without_published_at` that instantiate the transformers directly against a `draft` post.

**P1 8 — `is_comment_eligible()` parent publishability**: now fetches the parent post and returns false unless `is_post_publishable( $post )`, mirroring the `has_password=false` query argument already in `Reaction_Sync::find_post_by_bsky_uri()`.

**P1 9 — historical-leak cleanup**: added `schedule_historical_visibility_cleanup()` on `admin_init`, gated by the one-shot `atmosphere_visibility_cleanup_migrated` option. Walks any post with Atmosphere record meta, marks visibility cleanup, and queues `atmosphere_delete_post` for ones that are no longer publishable.

**P1 10 — `uninstall.php` missing `_atmosphere_visibility_cleanup`**: added to the meta-keys array, plus the new `atmosphere_visibility_cleanup_migrated` option.

**P1 11 — orphan bsky record on ambiguous `applyWrites` failure**: `delete_post()` now calls `stored_thread_records( …, true )`; the helper synthesizes a bsky `app.bsky.feed.post` URI from a bare reserved TID when `META_URI` is absent. An ambiguous failure that committed on the PDS but errored on the wire is now reachable from the delete path.

**P2 12 — filter receives raw `WP_Post` on redacted records**: `Post::transform()`, `Post::build_long_form_record()`'s two relevant sites, and `Document::transform()` all early-return the pre-filter record when `$redacted`. Matches the existing `atmosphere_is_short_form_post` gate at `class-post.php:130-138`.

**P2 13 — `publishedAt` on redacted documents**: now only emitted in the `! $redacted` branch.

includes/class-publisher.php
```diff
@@ -131,6 +144,33 @@ public static function publish_post( \WP_Post $post ): array|\WP_Error {
...
+	 * @param \WP_Post        $post   Post just written.
+	 * @param array|\WP_Error $result Publisher result.
+	 * @return array|\WP_Error
+	 */
+	private static function reconcile_post_after_write( \WP_Post $post, array|\WP_Error $result ): array|\WP_Error {
+		if ( \is_wp_error( $result ) ) {
+			return $result;
+		}
+
+		$fresh = \get_post( $post->ID );
```
kraftbj, line 170: **[P0] Reconcile reads stale WP object cache; the TOCTOU mitigation is broken for the default WP configuration.**

`\get_post( $post->ID )` here returns whatever the in-process `WP_Object_Cache` holds. On installs without a persistent object cache (Memcached / Redis / drop-in) — which is the WordPress default and the configuration most self-hosted sites run — a concurrent web request running `wp_update_post()` to add a password calls `clean_post_cache()` only in its own process. That invalidation does NOT propagate to other PHP processes. When this line runs, the worker still sees the pre-protect snapshot, `is_post_publishable( $fresh )` returns true on line 161, reconcile bails, and the just-committed records stay live on the PDS with no visibility-cleanup marker and no scheduled delete.

The regression test `test_publish_post_reconciles_when_post_protected_mid_publish` passes because in a single-process PHPUnit run, `wp_update_post()` invalidates the same in-process cache the test then reads. It does not exercise the cross-process scenario the fix exists to close.

Triple-corroborated by security, codex, and adversarial.

**Suggested fix:**

```php
\clean_post_cache( $post->ID );
$fresh = \get_post( $post->ID );
```

The same pattern needs to land on the comment publish path (`is_comment_eligible` reads `\get_post($post_id)` for the parent and is exposed identically — see the separate inline at `class-atmosphere.php:724`).

Add a test that pre-populates the WP_Post cache, mutates `$wpdb->posts`/`postmeta` directly (bypassing `wp_update_post`), and asserts reconcile detects + cleans up. That test would have caught this round trip.

includes/class-atmosphere.php
```diff
@@ -347,6 +365,121 @@ public function on_status_change( string $new_status, string $old_status, \WP_Po
...
+
+	/**
+	 * Schedule cleanup for records leaked before the visibility gate existed.
+	 */
+	public function schedule_historical_visibility_cleanup(): void {
+		if ( ! is_connected() ) {
+			return;
+		}
+
+		if ( \get_option( self::OPTION_VISIBILITY_CLEANUP_MIGRATED ) ) {
```
kraftbj, line 457: **[P1] Historical migration: no capability gate, unbatched query, one-shot option written AFTER the walk.**

Three problems compounding here:

1. **No capability check.** `admin_init` fires for any authenticated user reaching `/wp-admin/*` (including subscribers on self-registration sites, contributors writing drafts, anyone hitting `/wp-admin/profile.php`). `schedule_historical_visibility_cleanup()` checks only `is_connected()` and the migration flag. A subscriber-level user can force the full walk on first login.

2. **One-shot option written AFTER the walk.** The author's per-finding response says "the one-shot option is set BEFORE the walk to prevent re-trigger on parallel admin_init firings." The code does the opposite — the option is set at line 437, after the foreach completes. Two concurrent admin pageloads that arrive before either writes the option both execute the entire `WP_Query` + foreach. `wp_next_scheduled` dedupes per-event scheduling, so the practical impact is doubled CPU/DB load during the migration window plus an idempotent `mark_visibility_cleanup` per post.

3. **Unbatched.** `posts_per_page => -1` + 5-key OR `meta_query` + `post_type => 'any'` + `post_status => array_values( get_post_stati(…) )`. On a site with 10k+ historical posts that have ever had Atmosphere meta, this materializes the full ID list synchronously on an admin pageload, then loops with `get_post()` + 5x `get_post_meta()` per row. If `max_execution_time` kills the request mid-loop, the option flag is never written and the next admin pageload restarts from scratch.

Additionally: all `wp_schedule_single_event` calls use the same `\time()` timestamp, so the next wp_cron pass fires N events at once and hits PDS rate limits.

Flagged by **six** independent reviewers (correctness, security, reliability, adversarial, codex, project-standards).

**Suggested fix:**

```php
if ( ! \current_user_can( 'manage_options' ) ) {
    return;
}
if ( \get_option( self::OPTION_VISIBILITY_CLEANUP_MIGRATED ) ) {
    return;
}
// Claim the migration BEFORE the walk so concurrent admin_inits short-circuit.
\update_option( self::OPTION_VISIBILITY_CLEANUP_MIGRATED, '1', false );
```

Then either batch via paged `get_posts()` with a separate cursor option that persists across runs, or schedule a self-rescheduling `wp_schedule_single_event` that processes ~500 posts per tick. Stagger the scheduled cleanup timestamps so the next cron pass doesn't fire all N at once: `\time() + (int) ( $i / 50 ) * 60`.

includes/class-atmosphere.php
```diff
@@ -588,7 +721,12 @@ private static function is_comment_eligible( \WP_Comment $comment ): bool {
 			return false;
 		}
 
-		$post_id  = (int) $comment->comment_post_ID;
+		$post_id = (int) $comment->comment_post_ID;
```
kraftbj, line 846: **[P1] Comment publish path has the same stale-cache TOCTOU as the post path.**

`is_comment_eligible()` calls `\get_post( $post_id )` here to check the parent's publishability. Same exposure as the reconcile path (`class-publisher.php:159`): on cross-process races on default WP installs, the cached pre-protect parent snapshot is returned, `is_post_publishable()` returns true, and the comment record gets published to PDS as a reply on a now-protected parent.

`reconcile_comment_after_publish` (around line 1171) re-checks but reads the same stale cached parent.

**Compounding:** `includes/transformer/class-comment.php` has no `is_post_redacted` gate. The Post and Document transformers both have defense-in-depth at the transformer level — Comment relies entirely on the cron-handler gate, which is exactly the layer this TOCTOU defeats.

**Suggested fix:**

1. Add `\clean_post_cache( $post_id );` before this `\get_post()` call (and inside `reconcile_comment_after_publish` for symmetry).
2. Add a `Base::is_post_redacted` check to `Comment::transform()` against the parent post — mirror the pattern from `Post::transform()` and `Document::transform()`.

Adversarial caught this; security flagged it as residual risk.

includes/class-publisher.php
```diff
@@ -1585,10 +1630,14 @@ private static function stored_thread_records( int $post_id ): array {
 		// republish failed). Treat that as "nothing published" so the
 		// caller falls back to a fresh publish and the reserved TID is
 		// reused on the next attempt.
-		if ( ! $uri ) {
+		if ( ! $uri && ( ! $include_bare_tid || ! $tid ) ) {
```
kraftbj, line 1704: **[P1] URI synthesis trusts the CURRENT connected DID, not the DID that wrote the record.**

The bare-TID synthesis below (around line 1637) builds `at://{get_did()}/app.bsky.feed.post/{tid}` — but `get_did()` returns whoever is connected NOW, not whoever was connected when the record was written.

Failure mode after a disconnect+reconnect to a different DID (OAuth re-auth, handle change, atproto account migration):

1. Original publish under DID A returns ambiguously (timeout, 5xx). `META_TID` is persisted (via `get_rkey()` before the API call); `META_URI` is not (success branch did not run).
2. Site disconnects, reconnects with DID B.
3. A status transition or the historical migration calls `delete_post()`. `stored_thread_records( $post_id, true )` synthesizes `at://{DID_B}/app.bsky.feed.post/{tid}` — DID B's repo, not DID A's.
4. `applyWrites#delete` against DID B for a TID that lives (if it lives) under DID A. AT Proto returns 200 for missing records (idempotent). `Publisher::delete_post` clears local meta. Operator believes cleanup succeeded.
5. If the original publish under DID A actually committed (which is the entire reason this synthesis exists — the bare-TID-with-no-URI case), that record is **permanently orphaned on DID A's PDS** with no local pointer.

No per-record DID provenance exists in post meta today.

**Suggested fix:** persist the publishing DID alongside `META_TID` (a new `Post::META_DID` / `Document::META_DID` written by `get_rkey()` and `store_document_meta()`). When `stored_thread_records()` is called with `$include_bare_tid=true` and the stored DID differs from `get_did()`, either refuse to synthesize (return empty and log to a `META_ORPHAN_RECORDS` admin-visible field) or build the URI against the stored DID — but the latter would issue a delete on a repo we no longer authenticate to, which will 401.

Adversarial caught this.

includes/class-publisher.php
```diff
@@ -131,6 +144,33 @@ public static function publish_post( \WP_Post $post ): array|\WP_Error {
...
+			return $result;
+		}
+
+		if ( ! $fresh instanceof \WP_Post ) {
+			return $result;
+		}
+
+		Atmosphere::mark_visibility_cleanup( $fresh );
+
+		return self::delete_post( $fresh );
```
kraftbj: **[P2] Reconcile-cleanup failure has no retry and mislabels the operation.**

When reconcile detects a non-publishable post here and `delete_post( $fresh )` returns `WP_Error` (PDS 429, transient network, expired refresh token), three things go wrong:

1. The `WP_Error` overwrites the original publish `$result`. Callers see a "delete failed" error when the publish itself genuinely succeeded — records are live on the PDS, meta references them.
2. `atmosphere_publish_post_result` fires with that cleanup error. Listeners that increment success/failure counters mislabel the publish.
3. `log_cron_error()` writes `op=publish_post` even though the failure is in the reconcile-delete phase.
4. **No retry is scheduled.** `delete_post()` preserves meta on `WP_Error` (good), and the visibility-cleanup marker is set above on line 169 — but recovery requires the user to trigger another status transition. The historical migration won't re-flag the post because `OPTION_VISIBILITY_CLEANUP_MIGRATED` is already set.

Result: PDS records live, no breadcrumb operators can see (on managed hosts where `error_log` is invisible), no automatic recovery.

**Suggested fix:**

```php
if ( ! is_post_publishable( $fresh ) ) {
    Atmosphere::mark_visibility_cleanup( $fresh );
    $delete_result = self::delete_post( $fresh );
    if ( \is_wp_error( $delete_result ) ) {
        // Log the reconcile-cleanup failure with its own op label,
        // schedule a retry, return the original publish success.
        Atmosphere::log_cron_error( 'reconcile_cleanup', $post->ID, $delete_result );
        \wp_schedule_single_event(
            \time() + MINUTE_IN_SECONDS,
            'atmosphere_delete_post',
            array( $post->ID )
        );
    }
    return $result; // Original publish result; success or original error.
}
```

Flagged by correctness, reliability, adversarial, and codex.

includes/class-atmosphere.php
```diff
@@ -297,48 +320,43 @@ public function on_status_change( string $new_status, string $old_status, \WP_Po
...
 
-		if ( $is_unpublish ) {
-			$bsky_tid = \get_post_meta( $post->ID, Transformer\Post::META_TID, true );
-			$doc_tid  = \get_post_meta( $post->ID, Transformer\Document::META_TID, true );
-			if ( ! $bsky_tid && ! $doc_tid ) {
-				// Unpublish of a post that was never synced — nothing to clean up.
-				return;
-			}
-		}
+		self::$publishing_post_ids[ $post->ID ] = true;
```
kraftbj, line 389: **[P2] Per-post static `$publishing_post_ids` guard is not exception-safe.**

Set here on entry, fires `do_action( 'atmosphere_publishing', $post )` immediately after, runs `wp_clear_scheduled_hook` / `wp_schedule_single_event` / `mark_visibility_cleanup`, and `unset()`s the entry near the end of the handler — **no try/finally**. A third-party listener on `atmosphere_publishing` that throws (Sentry SDK transport error, metric hook with a network timeout, etc.) leaves the entry stuck for the rest of the PHP process.

Blast radius:
- Single web request: subsequent transitions of the same post ID in that request silently no-op.
- WP-CLI bulk import: poisoned entries accumulate across the process lifetime, silently skipping any post whose first attempt threw.

Flagged by reliability, security, correctness, adversarial, and codex.

**Suggested fix:** wrap the schedule body in try/finally:

```php
self::$publishing_post_ids[ $post->ID ] = true;
try {
    \do_action( 'atmosphere_publishing', $post );
    // ... existing schedule calls ...
} finally {
    unset( self::$publishing_post_ids[ $post->ID ] );
}
```

includes/class-atmosphere.php
```diff
@@ -347,6 +365,121 @@ public function on_status_change( string $new_status, string $old_status, \WP_Po
...
+			self::mark_visibility_cleanup( $post );
+
+			if ( \wp_next_scheduled( 'atmosphere_delete_post', array( $post->ID ) ) ) {
+				continue;
+			}
+
+			\wp_schedule_single_event( \time(), 'atmosphere_delete_post', array( $post->ID ) );
+		}
+
+		\update_option( self::OPTION_VISIBILITY_CLEANUP_MIGRATED, '1', false );
```
kraftbj: **[P2] Migration completes without verifying that cleanup events were durably scheduled.**

The foreach above (around lines 430-434) ignores `wp_schedule_single_event()`'s return value. Then this line sets the one-shot option regardless. If cron scheduling is blocked or filtered out (a plugin filtering `pre_schedule_event`, DISABLE_WP_CRON misconfiguration, options-table write rejected for any reason), the migration silently marks itself complete and the leaked records are never revisited because the migration is gated to run once.

Flagged by codex.

**Suggested fix:** track the count of `wp_schedule_single_event` calls that returned `true` vs `false`. Only set `OPTION_VISIBILITY_CLEANUP_MIGRATED` after a fully successful walk; on partial success, set a `_progress` option with the remaining post IDs and reschedule the migration via `wp_schedule_single_event` until empty.

This pairs naturally with the option-before-walk fix from the P1 finding on line 380 — once the walk is batched, the migration completion state is a queue rather than a one-shot.

includes/class-atmosphere.php
```diff
@@ -947,11 +1128,11 @@ private static function defer_when_parent_pending( \WP_Comment $comment ): bool
...
 	 *
-	 * @param string $op         One of 'publish_comment' | 'update_comment' | 'delete_comment'.
-	 * @param int    $comment_id Comment ID.
-	 * @param mixed  $result     Publisher call result.
+	 * @param string $op        Operation name.
+	 * @param int    $object_id Post or comment ID.
+	 * @param mixed  $result    Publisher call result.
 	 */
-	private static function log_cron_error( string $op, int $comment_id, $result ): void {
+	private static function log_cron_error( string $op, int $object_id, $result ): void {
```
kraftbj: **[P2] PDS error messages flow into `error_log` unsanitized.**

`$result->get_error_message()` is sprintf'd into the log line on the body of this function (around line 1140) with no escaping. WP_Error messages from `API::apply_writes` can include the PDS response body when JSON parsing fails (PR 64 just hardened that path, so this is more visible now). A malicious or compromised PDS — or a MITM under TLS termination on a misconfigured host — can return a body containing embedded `\n`, ANSI escapes, or fake `[atmosphere]` prefixes.

`error_log()` does not escape newlines. The resulting log entry spans multiple lines, can imitate other plugin log prefixes, and can confuse log-shipping pipelines that parse line-prefix patterns.

Narrow attack model; one-line fix:

```php
$message = \str_replace( array( "\r", "\n" ), ' ', $result->get_error_message() );
```

Applies to this helper and the two inline `\error_log` call sites elsewhere in this file.

Flagged by adversarial.

kraftbj (changes requested):
Fresh multi-persona pass + independent codex adversarial run after the fix commit `f887ea8`. **11 of 13 prior findings cleanly closed.** Two are incomplete in important ways, plus a handful of new issues the fix introduced.

## Cleanly closed (good)

- P0 1 bulk-protect via `did_action` guard — per-post static + new bulk test exercises two posts in one request without `reset_publishing_action()`. Right shape.
- P1 3 update-handler `has_post_records` branching mirrored.
- P1 4/5 cleanup `WP_Error` swallow — `log_cron_error()` covers all three cron handlers. (Log-only-no-retry surfaces as a smaller P2 below.)
- P1 6 `'0'` regression tests landed at all three sites.
- P1 7 non-publish-status redaction test landed for `draft`.
- P1 8 comment-eligibility parent publishability gate added. (Cache-staleness issue surfaces below.)
- P1 10 `uninstall.php` mirror — both keys present.
- P1 11 bare-TID URI synthesis for cleanup works for the documented case. (DID-rotation issue surfaces below.)
- P2 12 filter post-redaction early-return verified at all four transformer sites.
- P2 13 `publishedAt` on redacted documents gated.

## Still blocking (1 P0, 3 P1s)

**P0 1 — `reconcile_post_after_write` reads stale WP object cache.** The headline TOCTOU mitigation does not work on the default WP configuration. Triple-corroborated by security, codex, and adversarial. Details inline at `class-publisher.php:159`. One-line fix: `\clean_post_cache( $post->ID )` before `\get_post( $post->ID )`. The PHPUnit test passes only because in a single-process run, `wp_update_post()` invalidates the same in-process cache the test then reads — it does not exercise the cross-process race.

**P1 2 — Historical migration runs without a capability check, unbatched, with the one-shot option written AFTER the walk** — contradicting the per-finding response's claim it's set before. Triggered by any logged-in user's `admin_init`. Six independent reviewers (correctness, security, reliability, adversarial, codex, project-standards) flagged this. Details inline at `class-atmosphere.php:380`.

**P1 3 — Comment publish path has the same stale-cache TOCTOU AND no transformer-level redaction.** Details inline at `class-atmosphere.php:724`. The Comment transformer at `includes/transformer/class-comment.php` has no `is_post_redacted` gate; the Post and Document transformers both do. Adding the gate is defense-in-depth that catches the leak even when the cron handler's cache is stale.

**P1 4 — `stored_thread_records(true)` synthesizes URIs from the CURRENT connected DID.** After a disconnect+reconnect with a different DID (OAuth re-auth, handle change, atproto account migration), bare-TID cleanups target the new DID's repo while orphaning records on the old DID. Details inline at `class-publisher.php:1633`.

## Worth landing before release (P2s)

5. `class-publisher.php:171` — reconcile-cleanup failure has no retry and mislabels the operation. Inline.
6. `class-atmosphere.php:346` — per-post static guard not exception-safe. Inline.
7. `class-atmosphere.php:437` — migration completes without verifying cron events were durably queued. Inline.
8. `class-atmosphere.php:1135` — PDS error messages flow into `error_log()` unsanitized. Inline.

## Polish / follow-up (P2-P3, body-only)

9. **Cron-handler triplication got worse with the fix.** Three near-identical closures at `class-atmosphere.php:856-925` each gained the same `log_cron_error` line and the update/delete variants both grew an identical `has_post_records || ! has_visibility_cleanup_marker` ternary. Three blocks must be edited in lockstep with subtly different `op` labels. Extract a private dispatcher.
10. **`reconcile_post_after_write` wired inconsistently across writers** — called from `publish_post` (public), `update_single`, and `update_thread_in_place`, but NOT from `update_post` directly, `publish_single`, or `publish_thread`. Whether a write reconciles depends on which private writer the dispatcher picked. Add to `update_post`'s public entry and drop from private writers (or invert).
11. **Historical migration's `meta_query` duplicates `has_post_records()`'s key list** (`class-atmosphere.php:391-413` vs `:449-455`). Extract `private static function post_records_meta_keys(): array` to keep them in lockstep.
12. **Operator visibility gap.** `log_cron_error()` writes only to PHP `error_log`, which is invisible on most managed WP hosts (WordPress.com, WP Engine, Pressable, Kinsta). No admin notice, no Site Health entry, no stored failure list. Consider an admin notice surface or a per-post failure marker for follow-up.
13. **Testing gaps:** no cross-process cache-staleness test (the gap that lets P0 1 slip through), no reconcile-cleanup-failure test, no concurrent admin_init migration test, no try/finally test for the static guard, no DID-rotation test for the URI synthesis.
14. **Changelog `Type: fixed`** for what is a security fix. Companion DPoP entry from PR 64 uses `Type: security`. Worth aligning.

## Coverage

7 ce-code-review subagents + 1 codex challenge — 0 reviewer failures. Cross-reviewer corroboration on findings 1 (3x), 2 (6x), 3 (2x), 4 (2x), 5 (4x), 6 (5x). Codex's only meaningful disagreement was on finding 4 (DID-rotation) where it concluded "no finding" because `applyWrites#delete` is idempotent for missing records — accurate on the local-call side, misses that the original record permanently orphans on the previous DID. The correctness reviewer initially concluded `get_post()` bypasses stale cache via `wp_update_post`'s `clean_post_cache` call — accurate only for the same-process PHPUnit case, which is what the test exercises and is exactly why the test passes despite the cross-process bug.

Fix order: 1 → 2 → 3 → 4 → 5-8. Polish items can land in a follow-up PR.

includes/class-publisher.php
pfefferle: Addressed in dc5f926. `clean_post_cache( $post->ID )` lands immediately before `get_post()`. The cleanup-delete failure now flows through a dedicated `reconcile_cleanup` op label (`Atmosphere::log_reconcile_cleanup_error()`) and preserves the original publish `$result` so `atmosphere_publish_post_result` fires with the publish outcome the caller expects — that also addresses P2 #5 in the same change.

includes/class-atmosphere.php
pfefferle: Addressed in dc5f926. `admin_init` now only *queues* the migration via a single-event cron (`atmosphere_run_historical_visibility_cleanup`); the actual walk runs in cron context.

- **Cap check**: `maybe_queue_historical_visibility_cleanup()` bails when the current user is not `manage_options`-capable, so a subscriber's first `/wp-admin/profile.php` hit no longer schedules.
- **Batched**: `run_historical_visibility_cleanup()` walks `posts_per_page = 200` per fire with a persisted `_offset` and `orderby=ID, order=ASC`, rescheduling itself with a 60s gap until the final empty walk.
- **One-shot option after success**: `OPTION_VISIBILITY_CLEANUP_MIGRATED` is only set after the terminal short batch lands; a fatal mid-walk leaves it unset and the next `admin_init` re-queues from the persisted offset.

Open question: a full atomic-claim via `add_option` to prevent concurrent `maybe_queue` callers from racing the `wp_next_scheduled` dedupe is still worth adding. `wp_next_scheduled` is the only dedupe between admin_init firings, which is best-effort. Happy to fold that in if you want it on this PR rather than a follow-up.

includes/class-atmosphere.php
pfefferle: Both addressed in dc5f926. `is_comment_eligible()` now calls `clean_post_cache( $post_id )` before `get_post()` so a cross-process password-protect on the parent is visible to the cron handler. `Comment::transform()` reads the parent and consults `Base::is_post_redacted()` — if the parent has flipped non-public, `text` is emptied and the filter is skipped, matching the defense-in-depth pattern in the Post and Document transformers.

includes/class-publisher.php
pfefferle: Addressed in dc5f926. New `Post::META_DID` (`_atmosphere_bsky_did`) and `Document::META_DID` (`_atmosphere_doc_did`) are persisted alongside `META_TID` at TID reservation. `stored_thread_records()` synthesizes the URI using the recorded DID (falling back to `get_did()` only when no provenance is stored — pre-existing records). `delete_post()` now aborts with `atmosphere_did_mismatch` when the recorded origin DID is non-empty and does not match the currently-connected DID — instead of silently issuing an idempotent delete against the wrong repo, the operator sees a logged error and the local meta is preserved for manual remediation. `uninstall.php` cleans up the new keys.

includes/class-publisher.php
pfefferle: Op label and result preservation fixed in dc5f926 — `reconcile_post_after_write` now routes cleanup failures through `Atmosphere::log_reconcile_cleanup_error()` (`reconcile_cleanup` op) and returns the original publish `$result` so `atmosphere_publish_post_result` fires with the publish outcome.

Retry-on-failure is not in this commit. The visibility-cleanup marker stays set, so the next status transition or the historical migration revisits the record. A bounded one-shot rescheduled retry would be cleaner — happy to fold in if you want it on this PR rather than a follow-up.

includes/class-atmosphere.php
pfefferle: Intentionally not addressing. WordPress action listeners are not expected to throw, and the project leans away from `try/catch` as flow control. The narrow case for `try/finally` here is process-lifetime state cleanup (a stuck guard silently no-ops subsequent transitions for the rest of the request / CLI process), but the failure mode the audit anticipates — a third-party listener throwing — is the listener's bug, and the cost of paying it forward is loud failure rather than masked staleness.

Keeping the unconditional `unset()` at the end of the handler.

includes/class-atmosphere.php
pfefferle: Partially addressed in dc5f926. The migration is no longer one-shot — `OPTION_VISIBILITY_CLEANUP_MIGRATED` is set only after the terminal short batch lands. If a batch fires partial-walk-then-dies, the persisted `_offset` lets the next admin_init re-queue from the same spot.

Per-event `wp_schedule_single_event` return verification is not in this commit. If a plugin filters `pre_schedule_event` out for an individual post's `atmosphere_delete_post`, that post stays marked `META_VISIBILITY_CLEANUP` and the next status transition picks it up; the migration as a whole completes. Operator-visible per-event scheduling failure is worth adding in a follow-up alongside the broader "operator visibility gap" item (P3 #12).

includes/class-atmosphere.php
pfefferle: Addressed in dc5f926. `log_cron_error()` now strips `\r` and `\n` from `WP_Error::get_error_message()` before sprintf'ing the log line, so a hostile or compromised PDS can't smuggle multi-line content into log-shipping pipelines that parse line prefixes. The two inline `error_log` call sites in this file all route through `log_cron_error()` already (or its public `log_reconcile_cleanup_error()` alias used by the Publisher reconcile path).

includes/class-atmosphere.php
```diff
@@ -347,6 +376,179 @@ public function on_status_change( string $new_status, string $old_status, \WP_Po
...
+			\wp_schedule_single_event( \time(), 'atmosphere_delete_post', array( $post->ID ) );
+		}
+
+		if ( \count( $post_ids ) < self::VISIBILITY_CLEANUP_BATCH_SIZE ) {
+			\delete_option( self::OPTION_VISIBILITY_CLEANUP_MIGRATED . '_offset' );
+			\update_option( self::OPTION_VISIBILITY_CLEANUP_MIGRATED, '1', false );
+			return;
+		}
+
+		\update_option( self::OPTION_VISIBILITY_CLEANUP_MIGRATED . '_offset', $offset + self::VISIBILITY_CLEANUP_BATCH_SIZE, false );
```
kraftbj: **[P1] Offset pagination silently skips ~40-50% of leaked records on any site with >200 leaked records.**

Four independent reviewers (codex, correctness, reliability, adversarial). Worked example with 400 leaked posts:

1. Batch 1: `offset=0` returns IDs 1-200. Schedules `atmosphere_delete_post` for each. Bumps offset to 200.
2. The 200 delete cron events fire (each within 60s reschedule window). Each successful `delete_post` runs `clear_all_record_meta`, which removes META_TID/URI/THREAD_RECORDS — the exact keys the migration's `meta_query` selects on. Result set is now 200 posts (originally 201-400).
3. Batch 2: `offset=200` reads against the 200-row result set → empty → `count < BATCH_SIZE` terminator at line 499 → `OPTION_VISIBILITY_CLEANUP_MIGRATED` flipped. Migration marked complete.
4. **Posts 201-400 never cleaned. Records orphan permanently** because the migration is one-shot.

Same failure shape from concurrent deletes shifting the cursor mid-walk (adversarial finding adv-4). The `count < BATCH_SIZE` terminator at line 499 also can't distinguish 'walk exhausted' from 'transient empty result from DB hiccup or pre_get_posts filter' (reliability finding) — a transient empty mid-walk flips the option early.

**Suggested fix:** keyset paging on post ID.

```php
$post_ids = \get_posts(
    array(
        'fields'         => 'ids',
        'posts_per_page' => self::VISIBILITY_CLEANUP_BATCH_SIZE,
        'post_status'    => \array_values( \get_post_stati() ),
        'post_type'      => 'any',
        'orderby'        => 'ID',
        'order'          => 'ASC',
        'no_found_rows'  => true,
        'meta_query'     => array(
            'relation' => 'AND',
            array(
                'key'     => '_atmosphere_visibility_cleanup_last_id',
                'compare' => 'NOT EXISTS',
            ),
            // ... existing 5-key OR
        ),
    )
);
```

Or the simpler shape: persist `$last_seen_id`, query `WHERE ID > $last_seen_id ORDER BY ID ASC LIMIT 200`. Deletes don't shift positions when keyed by ID.

**Two other reliability concerns folded into this same finding** since they're all in the same function:

- **No PDS rate-limit backoff** on the 200-deletes-per-minute burst. Bluesky's write budget is ~5000/hr; 200/min = 12000/hr. PDS 429s log via `log_cron_error` but don't reschedule. Either widen the batch interval to 5 min, or retry-with-jitter on 429.
- **Mid-batch fatal strands the migration** until a `manage_options` user hits admin_init. The +60s reschedule queues only AFTER the foreach completes — if the worker dies mid-foreach (OOM, third-party listener fatal), the next event is never queued. Low-traffic sites with no admin sessions stall indefinitely. Queue the next event BEFORE the foreach so a crash leaves a recovery breadcrumb.

includes/transformer/class-post.php
```diff
@@ -220,6 +247,7 @@ public function get_rkey(): string {
 		if ( empty( $rkey ) ) {
 			$rkey = TID::generate();
 			\update_post_meta( $this->object->ID, self::META_TID, $rkey );
+			\update_post_meta( $this->object->ID, self::META_DID, \Atmosphere\get_did() );
```
kraftbj: **[P1] META_DID is frozen at TID reservation AND written non-atomically with META_TID.**

Two issues in one block — both defeat the DID guard the round-2 fix introduced.

### Issue A — never refreshed on republish (correctness, adversarial)

The `if ( empty( $rkey ) )` guard means META_DID is written ONLY on first TID mint. On any subsequent publish under a different DID — exactly the disconnect→reconnect-to-different-account flow the docblock at lines 33-46 documents as the threat model — `get_rkey()` short-circuits, returns the existing TID, and never touches META_DID. Concrete failure path:

1. Connected to DID A. Publish post 42. `get_rkey()` writes `META_TID=X, META_DID=A`. `apply_writes` fails on network blip. META_URI never written.
2. Reconnect to DID B. `atmosphere_publish_post` retries → `update_post` sees no URI → `publish_post` → `get_rkey()` short-circuits (META_TID present), does NOT touch META_DID. `apply_writes` succeeds against B. `mirror_thread_records_meta` writes `META_URI=at://did:plc:B/.../X`.
3. State: `META_TID=X, META_DID=A (stale), META_URI=B`.
4. Post password-protected. `delete_post` compares stored META_DID=A vs current=B → `atmosphere_did_mismatch` → bails. **Record on DID B orphans forever, blocked by the very guard meant to protect it.**

### Issue B — non-atomic write (codex)

META_TID is written before META_DID in two separate `update_post_meta` calls. If the second call fails (DB blip, plugin filter rejection) or the request dies between them, cleanup later sees TID + empty DID, the mismatch guard at `class-publisher.php:1070-1073` skips because it requires non-empty origin, and `stored_thread_records()` falls back to `get_did()`. Reopens the "delete against current account" bypass.

### Suggested fix

Write META_DID unconditionally on every `get_rkey()` call, and write it BEFORE META_TID so a partial failure leaves the row in 'no TID' state (cleanup ignores) rather than 'TID + no DID' state (bypass):

```php
public function get_rkey(): string {
    $rkey = \get_post_meta( $this->object->ID, self::META_TID, true );

    if ( empty( $rkey ) ) {
        $rkey = TID::generate();
        \update_post_meta( $this->object->ID, self::META_DID, \Atmosphere\get_did() );
        \update_post_meta( $this->object->ID, self::META_TID, $rkey );
    } else {
        // Refresh DID on every reservation lookup — idempotent when DID is
        // unchanged, repairs stale DID after a reconnect-then-republish.
        \update_post_meta( $this->object->ID, self::META_DID, \Atmosphere\get_did() );
    }

    return $rkey;
}
```

A stronger fix that survives the disconnect→reconnect-then-different-republish path: persist DID per-record inside `META_THREAD_RECORDS` (where the per-TID provenance actually belongs) and key the mismatch check off the stored per-TID DID. That also makes thread rewrites correct under DID change. Out of scope as a one-liner but worth considering.

The same fix applies at `class-document.php:190` — separate inline there.

includes/transformer/class-document.php
```diff
@@ -159,6 +187,7 @@ public function get_rkey(): string {
 		if ( empty( $rkey ) ) {
 			$rkey = TID::generate();
 			\update_post_meta( $this->object->ID, self::META_TID, $rkey );
+			\update_post_meta( $this->object->ID, self::META_DID, \Atmosphere\get_did() );
```
kraftbj: **[P1] Same META_DID issues as `Post::get_rkey()`** — frozen at first reservation, written non-atomically.

Apply the same fix here: write META_DID unconditionally on every `get_rkey()` call, ordered before META_TID. See full reasoning at the parallel inline on `includes/transformer/class-post.php:250`.

Additional Document-specific concern: `output_document_link()` (`includes/class-atmosphere.php:181-200`, the wp_head verification-link renderer) ALSO lazily mints Document::META_TID without touching META_DID. Because the `if ( empty( $rkey ) )` guard here means META_DID is only set on first reservation, any post whose verification link is rendered before its first publish ends up in the "legacy fallback" state forever — even on installs that shipped this PR from day 1. The fix above closes this too (every `get_rkey()` writes META_DID), but it also means the lazy-mint in `output_document_link` should call `(new Document( $post ))->get_rkey()` instead of inlining `\get_post_meta` + `TID::generate` + `\update_post_meta`, so the META_DID write is structurally tied to the TID write.

includes/transformer/class-comment.php
```diff
@@ -65,6 +79,10 @@ public function transform(): array {
 			'reply'     => $this->build_reply_ref( $comment ),
 		);
 
+		if ( $redacted ) {
```
kraftbj: **[P2] Comment transformer redaction emits an invalid strongRef** (correctness finding).

When `$redacted=true` (parent post is null or non-publishable), this block correctly empties `text` and skips the filter — good. But line 79 still calls `$this->build_reply_ref( $comment )`, which reads `Post::META_URI` and `Post::META_CID` from `comment_post_ID`. After a parent post's `on_before_delete` cleanup cascade or after the parent is fully deleted, those keys are gone, so `root.uri=''` and `root.cid=''` — an invalid strongRef.

`is_comment_eligible()` upstream rejects when post_uri/post_cid are empty, so this is unreachable via cron — BUT it IS reachable via direct callers: `Atmosphere::preview()` (the JSON preview admin can view per-comment) and any third-party listener of `atmosphere_transform_comment` that calls `Comment::transform()` outside the cron path. Those receive a malformed record rather than a meaningful 'do not publish' signal.

**Suggested fix:** when `$redacted`, return an early stub before building the reply ref, or omit the `reply` key entirely:

```php
if ( $redacted ) {
    return array(
        '$type'     => 'app.bsky.feed.post',
        'text'      => '',
        'createdAt' => $this->to_iso8601( $comment->comment_date_gmt ),
        'langs'     => $this->get_langs(),
    );
}
```

includes/class-atmosphere.php
```diff
@@ -297,48 +331,43 @@ public function on_status_change( string $new_status, string $old_status, \WP_Po
...
 
-		if ( $is_unpublish ) {
-			$bsky_tid = \get_post_meta( $post->ID, Transformer\Post::META_TID, true );
-			$doc_tid  = \get_post_meta( $post->ID, Transformer\Document::META_TID, true );
-			if ( ! $bsky_tid && ! $doc_tid ) {
-				// Unpublish of a post that was never synced — nothing to clean up.
-				return;
-			}
-		}
+		self::$publishing_post_ids[ $post->ID ] = true;
```
kraftbj, line 389: **[P2] Pushing back on the declined try/finally.** The reasoning ("WordPress action listeners are not expected to throw, project leans away from try/catch as flow control") is sound at face value but doesn't hold under PHP 8 throwing patterns. Adversarial constructed a concrete failure path:

WP-CLI `wp post import` or `wp atmosphere backfill` invokes `on_status_change` for thousands of posts in a single PHP process. A listener on `atmosphere_publishing` (any third-party plugin adding metrics, webhook delivery, sync to another service) calls:

```php
$payload = \json_decode( $webhook_response, true, 512, JSON_THROW_ON_ERROR );
```

Webhook returns malformed JSON. Listener throws `JsonException`. Line 380's `unset()` never runs. `$publishing_post_ids[ $post->ID ]` stays set for the rest of the PHP process. Same post ID later revisited in the process: line 353 early-returns. Publish/cleanup events never get scheduled.

This isn't "listeners shouldn't throw" — it's "modern PHP code routinely throws on transient input failures." Other patterns that throw in WP listeners: `JSON_THROW_ON_ERROR`, throw-mode PCRE, DI containers with strict typing, any code path using `wp_remote_post` + `\Throwable` handling, anything that loads symfony/console.

**Cost of try/finally is one line of code, zero runtime impact on the happy path:**

```php
self::$publishing_post_ids[ $post->ID ] = true;
try {
    \do_action( 'atmosphere_publishing', $post );
    // ... existing schedule calls ...
} finally {
    unset( self::$publishing_post_ids[ $post->ID ] );
}
```

Not a blocker if you'd rather keep the current shape, but worth reconsidering — the "loud failure rather than masked staleness" rationale you cited actually *favors* try/finally here: without it, the failure is silent (subsequent transitions no-op for the rest of the request), not loud.

kraftbj (changes requested):
Round-3 multi-persona pass + independent codex adversarial run after `dc5f9266`. **11 of 13 prior items addressed cleanly; 2 incompletely; 1 declined with reasoning.** The new code in `dc5f9266` introduces new exposures, mostly around the DID provenance work and the batched migration. The P0 cache-staleness fix is solid where it landed but missed a symmetric path.

## Cleanly addressed (thank you)

- P0 1 reconcile cache — `clean_post_cache` lands at both `class-publisher.php:169` and `class-atmosphere.php:803` ahead of `get_post`. Verified.
- P1 3 update-handler `has_post_records` branching mirrored to update + delete handlers.
- P1 4/5 log_cron_error covers all three cron sites.
- P1 6 `'0'` regression tests.
- P1 7 non-publish-status redaction test.
- P1 8 comment-eligibility parent publishability + `clean_post_cache`.
- P1 10 `uninstall.php` mirrors both new keys.
- P2 5 reconcile op label + original-result preservation.
- P2 6 try/finally — declined; reasoning has merit but a concrete PHP 8 counter-example exists (see inline finding 11).
- P2 8 log sanitization for CRLF.
- P2 12 filter post-redaction.
- P2 13 publishedAt on redacted documents.

## Still blocking (P1)

**P1 finding 1 — Migration's offset pagination silently skips ~40-50% of leaked records on any non-trivial site.** Four independent reviewers (codex, correctness, reliability, adversarial). The batched walk uses `offset` pagination on a `meta_query` whose selection criteria are exactly the meta keys that `atmosphere_delete_post` then clears. After batch 1 deletes posts 1-200, those rows drop out of the result set; batch 2's `offset=200` then reads against a now-shrunken window and silently skips posts 201-N. The migration is the centerpiece of the fix and as written can't complete on installs >200 leaked records. Details inline at `class-atmosphere.php:505`. Fix: keyset paging.

**P1 finding 2 + 3 — META_DID is frozen at TID reservation AND written non-atomically with META_TID.** Two reviewers (correctness, adversarial); plus codex on atomicity. The docblock at `class-post.php:33-46` documents the disconnect→reconnect-to-different-DID flow as the threat model META_DID defends against — but in that exact flow META_DID has the wrong value because `get_rkey()` short-circuits when META_TID is already present. Details inline at `class-post.php:250` and `class-document.php:190` (parallel site).

**P1 finding 4 — `reconcile_comment_after_publish` (`class-atmosphere.php:1275`) is missing the symmetric `clean_comment_cache()` before `\get_comment()`.** Round-2 added the cache-flush to `is_comment_eligible` and the post-side reconcile but not this path. Cross-process moderator-unapprove during in-flight comment publish leaves a Bluesky reply with META_URI set but no cleanup event scheduled — exact race the round-2 fix exists to close. Line 1275 isn't in a diff hunk so this is body-only; fix is one line: `\clean_comment_cache( $comment_id );` before line 1275's `\get_comment( $comment_id )`.

## Should fix before release (P2)

Four anchored inline (findings 9, 10, 11, plus the migration finding's compound issues); four body-only because their lines aren't in diff hunks:

- **`output_document_link()` (`class-atmosphere.php:181-200`) writes META_TID lazily without META_DID.** The wp_head renderer mints a Document TID on any singular pageview. Because `Document::get_rkey()` only writes META_DID inside `if ( empty( $rkey ) )`, a post viewed before its first publish is stuck in the "legacy fallback" state forever, even on installs that ship this PR from day 1. Defeats the DID guard for any post a logged-in visitor renders before publish.
- **Comment cleanup has no DID-mismatch guard at all** (correctness + adversarial). `Comment::META_DID` doesn't exist. `Publisher::delete_comment` (`class-publisher.php:1449`) and `delete_comment_by_tid` (`:1494`) issue `applyWrites#delete` against currently-connected DID with no origin check. Symmetric exposure to the post path the round-2 fix just closed; add `Comment::META_DID = '_atmosphere_bsky_comment_did'`, persist at TID-reserve, gate `delete_comment` and `delete_comment_by_tid`.
- **`delete_post_by_tids` (`class-publisher.php:1199-1240`) has no DID-mismatch guard either.** The hard-delete path via `on_before_delete` → `atmosphere_delete_records` → `delete_post_by_tids` is uncovered by the round-2 guard. Same threat model, uncovered path. Capture META_DID in `on_before_delete`, pass through the cron args, gate in the worker.
- **No PDS rate-limit backoff on the 200-deletes-per-minute migration burst** (reliability). Bluesky's typical write budget is ~5000/hr; 200/min = 12000/hr. PDS 429s log but don't reschedule — records orphan until a status transition (which rarely happens for non-public posts). Either widen the batch interval to 5 min, or retry-with-jitter on 429.

## P3 polish (non-blocking — please fold into a follow-up unless trivial)

- **`atmosphere_run_historical_visibility_cleanup` missing from `get_cron_hooks()`** (`includes/functions.php:148-165`). Three independent confirms (codex, security, correctness). Disconnect/deactivate/uninstall don't clear it. One-line fix.
- **Self-reschedule at `class-atmosphere.php:506` missing `wp_next_scheduled` dedup** (the queue-side has it but the handler-side doesn't). Concurrent admin_init + handler-in-flight can double-schedule, double-advancing the offset.
- **All-or-nothing skip in `delete_post` (`class-publisher.php:1070-1073`) when bsky and doc origin DIDs cross-mismatch.** Per-collection gating would let the side that DOES match clean up.
- **`clear_all_record_meta` doesn't delete META_DID.** Stale across cleanup cycles; self-heals on next publish (after finding 2 is fixed). Mismatched with the `uninstall.php` enumerator that DOES clear it.
- **Legacy posts (pre-PR) unbounded fallback exposure.** Posts with META_TID but no META_DID silently delete against current DID. One-shot backfill alongside the migration walk closes this.
- **Log sanitization incomplete** — strips CRLF but not NUL / ANSI escapes. Defense-in-depth only.
- **Operator visibility gap** — `error_log` invisible on managed hosts (WordPress.com, WP Engine, Pressable). Site Health entry or admin notice for failed cleanups.
- **META_DID not registered with `register_post_meta` + `auth_callback`** — authors with `edit_post_meta` cap via ACF/Pods/custom block fields could tamper to block or force cleanup.
- **Pfefferle's open question A (atomic-claim via `add_option`)**: correctness reviewer judges it defensible to leave as-is (`wp_schedule_single_event` has its own 10-min hash dedup); not a blocker. Follow-up if a race is observed.
- **Pfefferle's open question B (bounded reconcile-cleanup retry)**: not a blocker — the visibility-cleanup marker is set so the next status transition retries — but for non-public posts that may never happen. Worth a follow-up alongside the operator-visibility surface.
- **Terminal short-batch detection can't distinguish 'walk exhausted' from 'transient empty result'** (reliability). DB hiccup mid-walk flips `OPTION_VISIBILITY_CLEANUP_MIGRATED` prematurely. Validate `offset == 0` before treating empty as terminal, or compare to a counted total.

## Testing debt (6 P1 gaps on new round-2 code)

`dc5f9266` adds ~225 production lines vs ~2 test lines. Missing coverage on:
- Cross-process cache-staleness race (the existing test passes via `wp_update_post`'s side-effect cache invalidation — masks whether `clean_post_cache` is doing anything).
- `atmosphere_did_mismatch` path end-to-end.
- META_DID write at TID reservation.
- `stored_thread_records` origin-DID-vs-fallback fork.
- Multi-batch migration (only happy path with 2 posts tested; offset-persists-and-reschedules untested).
- Reconcile-cleanup-failure original-result preservation.

## Coverage

5 ce-code-review subagents + codex challenge. Adversarial stalled mid-write and was unstuck via SendMessage — confirmed alive afterwards. Cross-reviewer corroboration on findings 1 (4x), 2 (2x), 3 (1x but worked example), 4 (1x — the only reviewer who looked at the comment reconcile path).

Fix order: 1 → 2 + 3 → 4 → 7 → 8 → 5 → 6 → 9. The P3 polish list is genuinely non-blocking — none of those items are exploitable on their own, but several would close minor edge cases worth picking up in a follow-up PR.

includes/class-atmosphere.php
pfefferle: Addressed in 67cf38f. Switched to keyset paging via raw `$wpdb` query (`WHERE p.ID > %d ORDER BY p.ID ASC LIMIT %d`) so concurrent deletes can't shift the cursor. Next-batch cron is queued BEFORE the foreach so a mid-batch fatal leaves a recovery breadcrumb. Migrated flag only on confirmed empty batch — `count < BATCH_SIZE` was conflating "walk exhausted" with "transient empty," as you flagged. Test updated to require two invocations (terminal-walk contract).

Not addressed in this commit: PDS rate-limit backoff on the 200-deletes-per-minute burst. Happy to fold in a follow-up. The actual fire rate is bounded by WP-Cron dispatch (one-event-per-tick on most installs), not the queue rate — but the worst-case burst concern is fair on installs with system cron firing aggressively.

includes/transformer/class-post.php
pfefferle: Addressed in 67cf38f. `Post::get_rkey()` now writes `META_DID = get_did()` on every call, BEFORE `META_TID`. Compare-before-write guards the DB write so steady-state callers (like the `wp_head` document-link renderer that now routes through this method) only fire an UPDATE on actual DID transitions.

The ordering DID→TID means a partial failure leaves "DID set, no TID" (cleanup gates skip cleanly) rather than "TID set, no DID" (mismatch guard falls through to `get_did()` and re-opens the wrong-repo-delete bypass). The per-record DID-inside-META_THREAD_RECORDS suggestion is the stronger fix for thread rewrites under DID change — tracked as follow-up, not folded in to keep this PR scoped to the round-3 blockers.

includes/transformer/class-document.php
pfefferle: Addressed in 67cf38f. Same fix in `Document::get_rkey()`. Additionally, `Atmosphere::output_document_link()` now calls `( new Document( $post ) )->get_rkey()` instead of inlining `get_post_meta` + `TID::generate` + `update_post_meta` — the lazy mint at `wp_head` time is now structurally tied to the `META_DID` write, closing the "legacy fallback" state the inline path used to produce on installs that shipped this PR from day one.

includes/transformer/class-comment.php
pfefferle: Addressed in 67cf38f. Redacted branch now returns an early stub with no `reply` key at all — direct callers (preview, third-party listeners) see a "do not publish" placeholder rather than `{ root: { uri:'', cid:'' } }`. The cron path is unaffected because `is_comment_eligible()` still rejects on empty post URI/CID before the transformer runs.

includes/class-atmosphere.php
pfefferle: Folded in 05b3402. `on_status_change` now wraps the schedule body in `try { } finally { unset( ... ) }` so a throwing listener on `atmosphere_publishing` can't strand the per-post guard. WP-CLI bulk imports where the same PHP process handles thousands of posts were the deciding case — without this, one fatal early in the run silently no-ops every subsequent transition of the same post ID for the rest of the process.

kraftbj (approved):
Approving with the cache-staleness one-liner pushed in `d1c8aff`. The headline non-public-post leak class is solidly closed, the DID provenance defense is structurally correct, the migration rewrite is durable (keyset paging, recovery-before-foreach, empty-batch terminator), and the try/finally reversal closes the WP-CLI stuck-guard exposure cleanly.

Two follow-up-eligible items worth filing as separate issues before the release tag — both require a specific account-rotation sequence to exploit and are reasonable as separate PRs, but I'd want them tracked so they don't drift:

1. **`Comment::META_DID` doesn't exist.** Symmetric to the post-side DID guard that landed in round 2. `Publisher::delete_comment` and `Publisher::delete_comment_by_tid` issue `applyWrites#delete` against the currently-connected DID with no origin check, so the disconnect → reconnect-to-different-account flow that's now blocked for posts is still wide-open for reply records. Suggested shape: add `Comment::META_DID = '_atmosphere_bsky_comment_did'`, persist at TID reservation in `Comment::get_rkey()` mirroring the Post/Document fix, gate both cleanup entry points.

2. **`delete_post_by_tids` (the hard-delete path via `on_before_delete` → `atmosphere_delete_records`) has no DID guard.** Same threat model as the post-side, but uncovered because the cron event carries TIDs only (no post meta to read at fire time). Capture `META_DID` in `on_before_delete`, pass through the cron event args, gate in the worker.

The P3 polish list from my round-3 review (handler-side `wp_next_scheduled` dedup on the self-reschedule, all-or-nothing skip when bsky and doc origin DIDs cross-mismatch, `clear_all_record_meta` doesn't delete `META_DID`, legacy-post backfill, log sanitization beyond CRLF, operator visibility for `error_log` on managed hosts, `register_post_meta` + `auth_callback` for the new DID keys) can land in a cleanup PR after merge — none of those is exploitable on its own and a sweep makes more sense than per-item PRs.

Three rounds of rigorous fixes that each addressed nearly every flagged blocker, plus the test debt was caught up in round 4. Good work.

PR #65, opened 2026-05-18
Harden OAuth lifecycle, TID generation, and reaction sync

pfefferle:
## Proposed changes

Addresses several medium-severity findings from the latest security audit. Behaviour is preserved on the happy path; each change is defensive hardening.

- **REVOKE-1** — `OAuth\Client::disconnect()` now best-effort POSTs the refresh token to the auth server's `revocation_endpoint` (RFC 7009) before clearing local credentials, so a leaked refresh token stops working server-side too. The resolver captures and validates `revocation_endpoint` when the auth-server metadata advertises one; malformed values drop without failing connect.
- **REFRESH-1** — concurrent `OAuth\Client::refresh()` calls are serialised with an `add_option`-backed atomic lock. Stale locks are reclaimed after 30s; the lock-winner re-reads the connection under the lock so a previous holder's commit isn't burned. Inner logic extracted to `do_refresh()`.
- **RATE-1** — per-user transient counter on `Client::authorize()` and `handle_callback()` (10 attempts / 15 min). Defence-in-depth so a chain that survives the nonce check can't drive unbounded outbound requests to an attacker-chosen auth server.
- **TID-1** — `Transformer\TID::generate()` now persists `last_ts` to an `atmosphere_tid_last_ts` option so cross-worker mints share the monotonic floor. `clock_id` switched from `wp_rand()` to `random_int()` (with a `wp_rand` fallback if no CSPRNG source is available) for stronger per-worker entropy.
- **NONCE-1** — `OAuth\Nonce_Storage` hash key now includes the connected DID so a future multi-account variant can't silently share nonces between accounts.
- **Reaction-sync moderation** — `Reaction_Sync::insert_reaction()` skips the import when comments are closed on the target post, and honours the global `comment_moderation` setting instead of unconditionally writing `comment_approved => 1`.
- **uninstall.php** — cleans up the new `atmosphere_tid_last_ts` and `atmosphere_refresh_lock` options; covers `atmosphere_oauth_rate_<user>` transients via an `atmosphere_oauth_` wildcard.

### Other information

Test scaffolding updated: `tear_down()` in `class-test-client-refresh.php` now also clears the refresh lock. Dedicated tests for the lock contention, rate limit, and revoke paths are deferred to a follow-up.

## Testing instructions

1. `composer lint` — clean.
2. `npm run env-test` — full suite (348 tests, 1050 assertions) passes.
3. Manually: connect/disconnect a Bluesky account against a PDS whose metadata advertises a `revocation_endpoint`; confirm a `POST` to that endpoint appears in the server log on disconnect, with the refresh token in the body.
4. Manually: trigger ten consecutive `Client::authorize()` calls as the same admin user within 15 minutes; the eleventh returns `atmosphere_rate_limited`.

### Changelog entry

Committed at `.github/changelog/harden-oauth-and-reaction-sync`.


includes/transformer/class-tid.php
```diff
@@ -53,14 +73,23 @@ class TID {
...
+
+		if ( $ts <= $floor ) {
+			$ts = $floor + 1;
 		}
 
 		self::$last_ts = $ts;
 
+		if ( $ts > $persisted ) {
+			\update_option( self::OPTION_LAST_TS, $ts, false );
+		}
```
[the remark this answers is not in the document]
pfefferle: Real race, but the practical-collision risk it could cause is already covered by `clock_id`. Two workers minting at the same `$ts` collide only if their `clock_id` values also collide (1/1024 per pair; `clock_id` is now seeded via `random_int`). The persisted floor regressing just means subsequent workers may briefly mint values close to recent ones — they still produce distinct TIDs as long as `microtime` advances or the `clock_id` differs, which it does in 1023/1024 cases. The atomic `GREATEST` SQL update is the correct full fix but it would have to bypass `update_option`'s caching layer; deferring rather than adding raw `$wpdb` for a defence-in-depth refinement. Tracking for follow-up.

includes/transformer/class-tid.php
```diff
@@ -53,14 +73,23 @@ class TID {
...
+
 		if ( null === self::$clock_id ) {
-			self::$clock_id = \wp_rand( 0, 1023 );
+			self::$clock_id = \random_int( 0, 1023 );
```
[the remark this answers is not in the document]
pfefferle: Good catch — addressed in 5e3a8a1. `random_int` now sits in a `try / catch (\Throwable)` block that falls back to `wp_rand( 0, 1023 )`. The fallback is non-cryptographic but the collision space stays at 1024.

includes/oauth/class-client.php
```diff
@@ -612,12 +712,197 @@ public static function access_token(): string|\WP_Error {
...
+		self::revoke_refresh_token();
 		\delete_option( 'atmosphere_connection' );
 		clear_scheduled_hooks();
 	}
 
+	/**
+	 * Best-effort revoke the stored refresh token at the auth server.
+	 *
+	 * Reads the live `atmosphere_connection` option, decrypts the
+	 * refresh token + DPoP JWK, and POSTs the token to the auth
```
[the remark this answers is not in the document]
pfefferle: Addressed in 5e3a8a1. `revoke_refresh_token()` now re-runs the endpoint through `Resolver::is_safe_https_url()` at call time and bails if it fails, in addition to the resolver-time check that happens at connect.

includes/oauth/class-client.php
```diff
@@ -612,12 +712,197 @@ public static function access_token(): string|\WP_Error {
...
+			if ( false === $dpop_proof ) {
+				return;
+			}
+
+			$response = \wp_safe_remote_post(
+				$revocation_endpoint,
+				array(
+					'headers' => array(
+						'Content-Type' => 'application/x-www-form-urlencoded',
+						'DPoP'         => $dpop_proof,
```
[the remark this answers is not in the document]
pfefferle: Addressed in 5e3a8a1. The retry path now reads `dpop-nonce` off the second response and calls `DPoP::persist_nonce()`, matching the pattern in `handle_callback()` and `do_refresh()`.

includes/oauth/class-client.php
```diff
@@ -462,6 +493,75 @@ public static function refresh(): true|\WP_Error {
...
+		}
+
+		/*
+		 * `add_option` is atomic at the SQL layer (INSERT IGNORE).
+		 * Fourth arg `'no'` disables autoload so the lock value does
+		 * not bloat the always-loaded options cache. When another
+		 * worker holds the lock the call returns false and we fall
+		 * through to the re-read path: if that worker already updated
+		 * `atmosphere_connection` with a fresh token, the caller's
+		 * `access_token()` will see it without us issuing a second
```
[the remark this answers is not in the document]
pfefferle: Acknowledged — the code-review agent flagged the same gap. Adding lock-contention and stale-reclaim tests for `refresh()` is deferred to a follow-up so this security fix can land first. Tracking.

includes/oauth/class-client.php
```diff
@@ -612,12 +712,197 @@ public static function access_token(): string|\WP_Error {
...
+	 * Throttle OAuth start and callback per WordPress user.
+	 *
+	 * The admin-only capability gate keeps the surface narrow, but a
+	 * compromised admin or a chain that survives the nonce check (a
+	 * settings-screen XSS in another plugin, for example) could
+	 * otherwise drive unbounded outbound requests to an attacker-chosen
+	 * auth server via the handle field. The counter is cheap, scoped to
+	 * the current user (or a shared `0` bucket for unauthenticated
+	 * callers, which the admin nonce normally rejects anyway), and
+	 * naturally resets when the window expires.
```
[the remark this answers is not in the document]
pfefferle: Same as the refresh-lock case — adding success + 11th-attempt-blocked tests for `rate_limit_check()` is deferred to a follow-up. Tracking.

includes/oauth/class-client.php
```diff
@@ -434,25 +444,46 @@ public static function handle_callback( string $code, string $state ): true|\WP_
...
+	/**
+	 * Maximum lifetime (seconds) of the refresh lock before it is
+	 * presumed stale and reclaimed. One refresh roundtrip is typically
+	 * sub-second; the buffer covers the slowest reasonable network
+	 * timeout (15s) plus a margin for the request that set the lock to
+	 * actually finish.
+	 *
+	 * @var int
+	 */
+	private const REFRESH_LOCK_TTL = 30;
```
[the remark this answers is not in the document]
pfefferle: Good catch — bumped to 90s in cffb218. The math: `do_refresh()` can issue two sequential HTTP POSTs on the `use_dpop_nonce` retry path, each with a 15-second `wp_safe_remote_post` timeout. 90s leaves a comfortable margin over the 30-second worst-case HTTP cost plus encryption / option I/O overhead. Below the new TTL, a legitimate mid-flight refresh could be reclaimed by a second worker, reintroducing the very race this lock exists to prevent.

includes/oauth/class-client.php
```diff
@@ -612,12 +712,214 @@ public static function access_token(): string|\WP_Error {
...
+				$revocation_endpoint,
+				array(
+					'headers' => array(
+						'Content-Type' => 'application/x-www-form-urlencoded',
+						'DPoP'         => $dpop_proof,
+					),
+					'body'    => $body,
+					'timeout' => 10,
+				)
+			);
```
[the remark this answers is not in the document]
pfefferle: Acknowledged. The revoke path has five branches worth pinning (no endpoint, unsafe endpoint, happy POST, `use_dpop_nonce` retry, `WP_Error` logging). Adding a focused test that stubs `pre_http_request` and asserts the request shape + retry semantics is deferred to a follow-up alongside the lock-contention and rate-limit test gaps so this security batch can land. Tracking.

includes/class-reaction-sync.php
```diff
@@ -525,7 +544,7 @@ private static function insert_reaction(
 			'comment_date'         => \get_date_from_gmt( $gm_date ),
 			'comment_date_gmt'     => $gm_date,
 			'comment_type'         => $comment_type,
-			'comment_approved'     => 1,
+			'comment_approved'     => $comment_approved,
 			'comment_agent'        => 'ATmosphere/' . ATMOSPHERE_VERSION,
```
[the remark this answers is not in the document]
pfefferle: Acknowledged. The two new gates (`comments_open` → `false`, `comment_moderation` → `comment_approved=0`) are easy to pin with the existing reaction-sync test fixtures. Same follow-up bucket as the other test gaps from this round. Tracking.

includes/oauth/class-client.php
```diff
@@ -111,6 +111,11 @@ public static function redirect_uri(): string {
 	 * @return string|\WP_Error Authorization URL or error.
 	 */
 	public static function authorize( string $handle ): string|\WP_Error {
+		$rate_limit = self::rate_limit_check();
```
kraftbj: **P2 — Rate limiter charges the bucket on resolver failures.** `rate_limit_check()` fires before `Resolver::resolve()` (line 119). A handle typo (`nslookup` fail, `_atproto` TXT missing, malformed DID document — none of which actually hit the auth server) still increments the per-user counter. Ten consecutive typos in 15 minutes lock the user out of OAuth without any abuse occurring.

Fix: either move `rate_limit_check()` to fire after a successful `resolve()`, or split into `precheck()`/`charge()` so the network-bound path is the only one that consumes budget.

includes/oauth/class-client.php
```diff
@@ -549,7 +576,9 @@ public static function refresh(): true|\WP_Error {
 
 	/**
 	 * Inner refresh routine. Runs under the refresh lock acquired in
-	 * `refresh()`. Do not call directly.
+	 * `refresh()`. Do not call directly — concurrent callers will
```
kraftbj, line 622: Two related findings on the `refresh_locked()` lifecycle:

**P1 — Disconnect race vs in-flight refresh.** `refresh_locked()` writes `atmosphere_connection` (line 708) *after* the network call completes. If `Admin::handle_disconnect()` deletes the connection options while this method is mid-flight, the worker resurrects them on return. Failure chain: cron starts `Client::refresh()` → admin clicks disconnect → local options deleted → refresh worker returns → `update_option( 'atmosphere_connection', $conn )` writes fresh tokens back, and `get_identity()` lazily repopulates identity. If the admin reconnected to a different DID in that window, the resurrected row points at the wrong account.

Fix options: (a) acquire the same `Client::lock()` in `disconnect()` so the two cannot interleave; (b) capture the option's value at lock-acquisition time and only `update_option` if the row hasn't been deleted (compare-and-swap on the connection row, not just the lock); (c) check `! empty( get_option('atmosphere_connection') )` before the post-refresh write.

**P2 — DPoP nonce from the retry response is not persisted.** The use_dpop_nonce retry block (around line 643–668) extracts `$status` and `$data` but does not call `DPoP::persist_nonce( $dpop_jwk, $token_endpoint, $nonce_header_of_retry )`. Same gap exists in `handle_callback()` (the token-exchange retry around line 444). Copilot review thread 3262107801 already flagged this; the fix in `5e3a8a1` only added the call in `revoke_refresh_token()`. Without persisting the second response's nonce, the *next* refresh / token exchange starts from a known-stale nonce and pays another `use_dpop_nonce` round-trip.

includes/oauth/class-client.php
```diff
@@ -923,14 +952,216 @@ public static function access_token(): string|\WP_Error {
...
 	 * Mirrors the cleanup performed on plugin deactivate / uninstall.
+	 *
+	 * Best-effort revokes the refresh token at the auth server before
+	 * deleting local state (RFC 7009). A leaked refresh token therefore
+	 * stops working server-side even though disconnect already removed
+	 * it locally. Revocation failures are logged but do not block the
+	 * local cleanup.
 	 */
 	public static function disconnect(): void {
+		self::revoke_refresh_token();
```
kraftbj: **P1 — `disconnect()` blocks for up to ~20 seconds on a slow or unreachable auth server.** `revoke_refresh_token()` issues up to two sequential `wp_safe_remote_post` calls at the default 10-second timeout each, then `disconnect()` proceeds with the local cleanup. The docblock above explicitly says revocation must be best-effort and not gate disconnect, but the synchronous call makes it so.

Fix: push revocation to `wp_schedule_single_event( time(), 'atmosphere_revoke_refresh', [ $refresh_token_ciphertext, $dpop_jwk_ciphertext, $revocation_endpoint ] )` so the admin click returns immediately, *or* drop each request timeout to ~3 seconds so worst case is ~6s. The async-cron approach is cleaner but requires a cron handler that's allowed to read the credentials after the option is deleted (so it has to receive them as args or read from a one-shot transient).

includes/oauth/class-client.php
```diff
@@ -923,14 +952,216 @@ public static function access_token(): string|\WP_Error {
...
+	 * reachable.
+	 */
+	private static function revoke_refresh_token(): void {
+		$conn = \get_option( 'atmosphere_connection', array() );
+
+		if ( ! \is_array( $conn ) ) {
+			return;
+		}
+
+		$revocation_endpoint = isset( $conn['revocation_endpoint'] ) ? (string) $conn['revocation_endpoint'] : '';
```
kraftbj: **P2 — `revocation_endpoint` is re-validated via `is_safe_https_url()` but not bound to the auth-server's issuer origin.** The `Resolver` validates it at connect time, but the stored option is not trusted at use time (correct design), and the second-pass check only verifies it's *some* safe HTTPS URL — not that it shares an origin with `$conn['auth_server']`. Attack chain: option tampered between connect and disconnect (backup restore, misbehaving sibling plugin, compromised admin) → `revocation_endpoint` swapped for `https://attacker.example/collect` → on disconnect we POST the decrypted refresh token + a DPoP proof keyed to the hostile URL to the attacker.

Fix: add a same-origin check, e.g.
```php
$auth_host = \wp_parse_url( $conn['auth_server'] ?? '', PHP_URL_HOST );
$rev_host  = \wp_parse_url( $revocation_endpoint, PHP_URL_HOST );
if ( '' === $auth_host || $auth_host !== $rev_host ) { return; }
```

includes/oauth/class-client.php
```diff
@@ -923,14 +952,216 @@ public static function access_token(): string|\WP_Error {
...
+	 * settings-screen XSS in another plugin, for example) could
+	 * otherwise drive unbounded outbound requests to an attacker-chosen
+	 * auth server via the handle field. The counter is cheap, scoped to
+	 * the current user (or a shared `0` bucket for unauthenticated
+	 * callers, which the admin nonce normally rejects anyway), and
+	 * naturally resets when the window expires.
+	 *
+	 * @return true|\WP_Error
+	 */
+	private static function rate_limit_check(): true|\WP_Error {
```
kraftbj, line 1360: **P2 — `rate_limit_check()` is a non-atomic read-then-write.** Two concurrent OAuth callbacks at attempts=9 both `get_transient` → both see 9 → both pass the `>=` check → both `set_transient(10)`. The cap silently doubles under contention. On persistent object cache (Redis/Memcached) this is fixable with `wp_cache_add` + `wp_cache_incr`; on DB transients, the same `INSERT IGNORE` + atomic `UPDATE` pattern used by `lock()` (line 738) closes it.

Four reviewers independently flagged this. Even if the boundary-doubling is acceptable as defence-in-depth, it should be documented in the docblock so a future maintainer doesn't assume the cap is hard.

includes/transformer/class-tid.php
```diff
@@ -53,14 +73,34 @@ class TID {
 	public static function generate(): string {
 		$ts = (int) ( \microtime( true ) * 1_000_000 );
 
-		if ( $ts <= self::$last_ts ) {
-			$ts = self::$last_ts + 1;
+		// Cross-request floor: another worker may have minted a TID
+		// later than this process's static counter.
+		$persisted = (int) \get_option( self::OPTION_LAST_TS, 0 );
```
kraftbj, line 109: **P2 — The new persistent monotonic floor is not actually atomic across workers.** Three reviewers (codex, reliability, adversarial) corroborated independently. Workers A and B both `get_option( OPTION_LAST_TS, 0 )` → both compute `$ts = max( microtime, persisted+1 )` → both `update_option(...)`. The result is that the floor does not strictly advance, and the rkey collision-avoidance falls back entirely to the 10-bit `clock_id` (`random_int` per-process). That's better than nothing but it's not the cross-worker guarantee the docblock claims; `Publisher::publish_single()` will surface duplicates as `record already exists` `applyWrites` errors.

Fix: replace the read-compare-write with a compare-and-swap `$wpdb->query` of the form `UPDATE {$wpdb->options} SET option_value = GREATEST(option_value, %s) WHERE option_name = %s`, then re-read inside the same critical section. Alternatively, since the value never decreases, an `INSERT ... ON DUPLICATE KEY UPDATE option_value = GREATEST(option_value, VALUES(option_value))` is a single SQL statement that does the right thing.

If the decision is to accept the residual risk and rely on `clock_id`, update the docblock to be explicit so the next reader doesn't assume the option-write guarantees monotonicity.

includes/class-reaction-sync.php
```diff
@@ -504,6 +504,15 @@ private static function insert_reaction(
 		array $notification,
 		array $profile
 	): int|false {
+		/*
+		 * Respect the post-level comments-open policy. If the site
+		 * admin closed comments on a post, an inbound Bluesky like /
+		 * repost / reply should not appear as a fresh comment row.
+		 */
+		if ( ! \comments_open( $post_id ) ) {
```
kraftbj: **P2 — `comments_open()=false` silently drops the reaction *and* advances the watermark past it.** Combined effect: reactions targeting a post with comments closed are dropped at this `return false`, but `paginate()` (around line 233) still treats the first URI seen as the new `newest` pointer regardless of whether `insert_reaction()` accepted it. When comments are later re-opened, every reaction below the watermark is permanently invisible to future syncs — they're behind the `WATERMARK_GRACE = 10` window after a single cron run.

Fix options: (a) skip the `comments_open()` short-circuit and let `wp_allow_comment()` handle policy (see the separate comment at line 547); (b) drop a tombstone meta on the post (`_atmosphere_reactions_held = at://...`) so the watermark can be safely advanced; (c) propagate the reason for skip back to `paginate()` so it can distinguish 'policy drop' from 'duplicate'.

includes/class-reaction-sync.php
```diff
@@ -525,7 +544,7 @@ private static function insert_reaction(
 			'comment_date'         => \get_date_from_gmt( $gm_date ),
 			'comment_date_gmt'     => $gm_date,
 			'comment_type'         => $comment_type,
-			'comment_approved'     => 1,
+			'comment_approved'     => $comment_approved,
```
kraftbj: **P1 — Reactions bypass the rest of the WordPress moderation pipeline.** Respecting `comment_moderation` is half the job. The actual WordPress comment policy lives in `wp_allow_comment( $commentdata )`, which evaluates: previously-approved-author bypass, moderation keys, disallowed keys (the spam blacklist), max links threshold, and the Akismet pre-check (via `pre_comment_approved`/`comment_post` actions). `wp_insert_comment()` skips that entire pipeline.

The changelog entry says the PR "respect[s] WordPress comment moderation when importing Bluesky reactions" — that's misleading. On a site whose moderation policy depends on Akismet or disallowed-keys (the common case), this PR doesn't actually change the threat model: spam from Bluesky still lands published.

Fix: replace the direct `wp_insert_comment()` with `wp_new_comment( $commentdata, true )`, which runs `wp_allow_comment()` and the surrounding filter chain. The `true` flag preserves IP / user-agent metadata. The `comment_agent` stamp already prevents the publish cron from picking it back up. Either do this or update the changelog wording so it doesn't oversell what was fixed.

kraftbj (commented):
## Multi-agent review summary

Ran `/ce-code-review` (8 reviewer personas + CE always-on agents) and the OpenAI Codex CLI in `challenge` mode against the post-merge tip (`848c0d8`) vs `origin/trunk` (`9495939`). Diff is ~360 changed lines across OAuth, TID, reaction-sync, nonce-storage, resolver, and uninstall.

**Verdict: Not ready to merge.** Strong cross-reviewer corroboration on race conditions in the new lock/refresh/disconnect lifecycle and on a moderation-bypass in reaction-sync that the changelog implies is fixed but isn't (only `comment_moderation` is honoured; Akismet / disallowed-keys / link thresholds are still skipped). Codex agreed independently.

### Headline findings (cross-reviewer)

- **P1 — Disconnect can be undone by an in-flight refresh.** `refresh_locked()` writes `atmosphere_connection` after the network call; if an admin clicks disconnect while a cron refresh is in-flight, the worker resurrects the deleted connection (potentially with a different DID's tokens if the admin reconnected in that window). See inline on `class-client.php:579`.
- **P1 — `disconnect()` synchronously waits up to ~20s on a slow/unreachable auth server.** Two sequential `wp_safe_remote_post` calls at 10s each before any local cleanup. Code comment explicitly says "must not be gated on the auth server being reachable." Inline on `class-client.php:963`.
- **P1 — Reaction-sync still bypasses the WordPress moderation pipeline.** Honouring `comment_moderation` is the easy half; the hard half is `wp_allow_comment()` (Akismet, disallowed-keys, max-links, previously-approved-author). Inline on `class-reaction-sync.php:547`.
- **P1 — Cron one-shots silently lose work on `atmosphere_refresh_locked`.** `access_token()` waits ~5s for the holder, refresh worst-case is 2×15s. `publish_comment` / `delete_records` / etc. log the soft error and do not reschedule. (Codex finding.)
- **P2 — TID cross-worker monotonic floor is racy.** `get_option` → compare → `update_option` is not atomic; two workers can both read the same floor and overwrite each other. Inline on `class-tid.php:78`. Flagged by Codex, reliability, adversarial, and correctness reviewers.
- **P2 — DPoP nonce from the use_dpop_nonce retry response is not persisted in `refresh_locked()` or `handle_callback()`.** Copilot already flagged this; commit `5e3a8a1` only added the call in `revoke_refresh_token()`. Inline on `class-client.php:579`.
- **P2 — Rate-limit counter is non-atomic and consumes budget on resolver failures.** A typo in the handle field charges the user's bucket; concurrent calls at the boundary can both pass. Inline on `class-client.php:114` and `:1149`.
- **P2 — `revocation_endpoint` not bound to the auth-server origin.** A tampered `atmosphere_connection` row could exfiltrate the decrypted refresh token to a hostile URL on disconnect. Inline on `class-client.php:989`.
- **P2 — Reaction-sync watermark advances past `comments_open()=false` drops.** Reactions targeting a post with comments closed are silently dropped *and* the `newest` pointer moves past them; re-opening comments later won't recover them. Inline on `class-reaction-sync.php:512`.
- **P2 — Changelog entry contains developer jargon.** Violates the AGENTS.md rule about end-user-friendly changelog wording. Inline on `.github/changelog/harden-oauth-and-reaction-sync:4`.

### Test coverage (from the testing reviewer)

The PR body acknowledges "dedicated tests for the lock contention, rate limit, and revoke paths are deferred to a follow-up." The testing reviewer flagged this as three separate P1 gaps; previous-comments reviewer confirmed Copilot already requested them and the author explicitly deferred. Concretely missing:

- `rate_limit_check()` — no test for the 11th attempt returning `atmosphere_rate_limited`, none for per-user bucket isolation.
- `revoke_refresh_token()` — entire 136-line method has zero coverage. The use_dpop_nonce retry, the no-endpoint early return, the is_safe_https_url rejection — none exercised.
- `Nonce_Storage::hash()` DID-scope — no test that the same URL under different DIDs produces different storage keys.
- `insert_reaction()` — no test for `comments_open=false` short-circuit, none for `comment_moderation=1` producing `comment_approved=0`.
- `TID::generate()` cross-worker — `test_generate_monotonic` only covers within-process (the static `$last_ts` handles that without the option).

### Lower-severity / observational

- `atmosphere_oauth_rate_0` is a shared bucket for all callers with `user_id=0`; mitigated by the upstream nonce + capability gate but a latent fragility.
- TTL-based stale-lock theft in `lock()` is ownerless: `unlock()` blindly deletes the row. Worker A stalls >90s → B steals → A's `finally` deletes B's lock → C acquires while B is still mid-flight.
- `REFRESH_LOCK_TTL` docblock references `refresh_locked()` by name (will drift on rename).
- `atmosphere_tid_last_ts` hardcoded in `uninstall.php` has no sync comment pointing at `TID::OPTION_LAST_TS` (the adjacent `_atmosphere_refresh_lock` has one).
- `insert_reaction()` docblock claims callers handle policy checks but the method itself gates on `comments_open()`.
- `revoke_refresh_token()`'s use_dpop_nonce retry does not re-parse `$data` (cosmetic — only `$status` is consulted downstream).
- `refresh()` doesn't re-check `expires_at` after acquiring the lock; could save a token round-trip when a sibling worker just rotated.

### Pre-existing / out-of-scope

- `wait_for_token_refresh()` waiter on default WP installs (no persistent object cache) may not observe the rotated `atmosphere_connection` because `get_option` returns a cached pre-lock value. This race pre-dates the PR but the new lock makes it the dominant failure mode for contended publishes.
- `paginate()` watermark advance on mid-stream `WP_Error` (correctness, pre-existing).

### Coverage notes

- 9 reviewers + Codex returned results. No failures or timeouts.
- Some reviewers reported `/tmp/compound-engineering/ce-code-review/<run>/` write was denied; full findings still flowed back via the return channel.
- No requirements plan was discovered (no `plan:` arg; no plan referenced in PR body).
- Untracked files were not in scope.

Run artifacts in `/tmp/compound-engineering/ce-code-review/20260519-103730-34a9132b/`. Codex raw output in `/tmp/atmosphere-codex/`.

.github/changelog/harden-oauth-and-reaction-sync
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: changed
+
+Tighten OAuth token lifecycle, harden TID generation against cross-worker collisions, and respect WordPress comment moderation when importing Bluesky reactions.
```
kraftbj: **P2 — Changelog uses developer jargon.** AGENTS.md Common Pitfalls section makes this a hard rule: changelog entries are end-user-visible on the WordPress update screen and must avoid jargon, class names, and method names. "OAuth token lifecycle", "TID generation", and "cross-worker collisions" all fail that test.

Suggested rewrite:
```
Improved Bluesky connection reliability, fixed a rare duplicate-record issue when publishing simultaneously, and now respects your comment moderation settings when importing Bluesky reactions and replies.
```

Also consider splitting into multiple entries (significance: patch each) so each user-visible change reads on its own line — there are at least three independent improvements bundled here.

includes/oauth/class-client.php
pfefferle: Fixed in 22d06b6. `rate_limit_check()` now fires AFTER a successful `Resolver::resolve()` in `authorize()`. Resolver's own DNS/HTTP timeouts bound the pre-resolve flood surface, so handle typos no longer burn the user's bucket.

includes/oauth/class-client.php
pfefferle: Fixed in 983e445. `refresh_locked()` re-reads `atmosphere_connection` immediately before both write paths (happy + `needs_reauth`). If the row is gone (disconnect won the race), the worker bails with `atmosphere_disconnected_mid_refresh` and the rotated tokens are discarded instead of resurrecting the connection.

includes/oauth/class-client.php
pfefferle: Fixed in 983e445. `disconnect()` now captures the refresh-token + DPoP-JWK ciphertexts and the revocation endpoint *before* wiping local options, then schedules a single-shot `atmosphere_revoke_refresh_token` cron event *after* `clear_scheduled_hooks()` runs (so the helper doesn't clear the just-queued event). The admin click returns immediately; the cron worker decrypts and POSTs in the background. `revoke_refresh_token()` is now public and takes its inputs as args (now four — see the origin-binding thread).

includes/oauth/class-client.php
pfefferle: Fixed in 22d06b6. `revoke_refresh_token()` now takes the auth-server issuer URL as a fourth arg (`disconnect()` captures it before wiping). After `is_safe_https_url()`, the worker checks the revocation endpoint's host matches the issuer's host (lower-case, host-only) and bails if not. A tampered row pointing `revocation_endpoint` at an attacker host can no longer exfiltrate the decrypted refresh token.

includes/oauth/class-client.php
pfefferle: Fixed in 22d06b6 using the same `INSERT IGNORE` + compare-and-swap pattern that `Client::lock()` already uses (introduced in #68, line 802). Same reasoning applies: `set_transient` / `get_transient` is the get-then-set race you flagged, `add_option` is preflight-SELECT + `INSERT … ON DUPLICATE KEY UPDATE` (silent overwrite, not actually atomic for our purpose), and `wp_cache_add` is per-request on default WP (cannot see a concurrent worker's bucket). `INSERT IGNORE` on the UNIQUE `option_name` index is atomic at the SQL layer; combined with a bounded `UPDATE … WHERE option_value = $previous` CAS-loop, the read-decide-write cycle becomes lock-free and cross-process correct. Packed value is `attempts|expires_at` so the bucket is a single CAS-replaceable string and window expiry is encoded in the value rather than relying on transient TTL. Inline comment in `rate_limit_check()` walks through the alternative primitives and why each is unsuitable.

includes/transformer/class-tid.php
pfefferle: Fixed in 22d06b6 using the same direct-SQL atomicity pattern as `Client::lock()` and the rate limiter above. `update_option` has no conditional-write path, so the read-modify-write shape (`get → max → update`) is intrinsically racy and a slower worker regresses the floor by overwriting a faster worker's larger value. Replaced with `UPDATE … WHERE CAST(option_value AS UNSIGNED) < $ts` — single atomic statement, WHERE clause defeats the regression race. The per-process `self::$last_ts` static keeps the hot path cheap (no UPDATE when the candidate is already below it). First-write still goes through `update_option` so the row lazily exists; the CAS path takes over from the second write onward. Inline comment in `generate()` walks through the regression scenario step-by-step.

includes/class-reaction-sync.php
pfefferle: Fixed in 22d06b6 by removing the `comments_open()` gate from `insert_reaction()` — your option (a). Reasoning baked into the diff comment: federated reactions are an audit record of activity that happened on Bluesky, not a comment-form submission; saving them to `wp_comments` preserves the federation history even on closed-comments posts. The standard `comments_open` UI gate still controls rendering on the front end, and the moderation pipeline (P1 fix below) decides the approval state. The watermark can now advance cleanly because every paginated item reaches `wp_insert_comment`.

includes/class-reaction-sync.php
pfefferle: Fixed in 4c26b61. `insert_reaction()` now runs `wp_allow_comment( $commentdata, true )` — moderation keys, disallowed-keys (hard-rejects via `WP_Error`), max-links, previously-approved-author bypass, AND the `pre_comment_approved` filter chain that Akismet hooks into. `wp_is_comment_flood` is short-circuited to false for that single call because federated reactions are server-to-server traffic without an IP — WP's IP/email-based 15-second flood heuristic models user-submitted comments, not federation imports; upstream relay rate-limits cover the abuse vector. The filter is removed immediately after the call so it can't affect any subsequent user-submitted comment in the same request. Changelog rewritten accordingly (no longer claims just "comment moderation").

.github/changelog/harden-oauth-and-reaction-sync
pfefferle: Fixed in 22d06b6. Adopted your suggested rewrite, slightly extended to also call out the disconnect-speed improvement from the P1 fix: *"Improved Bluesky connection reliability and disconnect speed, fixed a rare duplicate-record issue when publishing simultaneously from multiple workers, and now respects your comment moderation and spam filter settings when importing Bluesky reactions and replies."*

kraftbj (commented):
## Re-review of fix commits (`983e445`, `4c26b61`, `22d06b6`)

Ran a focused audit of the new commits plus an independent Codex (`challenge` mode) pass. Both agreed on the major findings below.

**Verdict: still not ready to merge.** Most fixes landed correctly, but one of the P2 fixes is *silently disabled* by a hook-registration mismatch, and the P1 disconnect-race fix has a second-order hole that survives a disconnect-then-reconnect cycle.

### Verified correctly fixed ✅

- DPoP nonce persistence on `use_dpop_nonce` retry — added at both `handle_callback()` (line 464) and `refresh_locked()` (line 696), mirroring the existing pattern in `revoke_refresh_token()`.
- Rate-limit moved AFTER `Resolver::resolve()` in `authorize()` (typos no longer charge the bucket).
- Rate-limit atomicity: `INSERT IGNORE` + bounded 5-try CAS loop. Failure-mode is fail-closed (acceptable trade-off). New `_atmosphere_oauth_rate_*` keys covered by uninstall wildcard.
- Reaction-sync moderation: `wp_allow_comment()` now runs the full pipeline (Akismet, disallowed-keys, comment_whitelist, max-links). `wp_is_comment_flood` is short-circuited only for the duration of one call.
- Changelog rewritten in end-user terms.
- Test suite grew from 354 → 383 tests, all passing.

### Still open / regressed 🔴

- **P1 — `add_action` accepts only 3 args; revocation origin-binding check never fires.** `disconnect()` schedules with 4 args (refresh_token, dpop_jwk, revocation_endpoint, auth_server) but `class-atmosphere.php:170` declares `accepted_args = 3`. WordPress trims the 4th, so `$auth_server_issuer` defaults to `''`, and the `if ( '' !== $auth_server_issuer )` guard at `class-client.php:1137` short-circuits — the host check the previous P2 fix added is silently dead code. A tampered `revocation_endpoint` that survives `is_safe_https_url()` (any well-formed HTTPS host on a non-private IP) still receives the decrypted refresh token. Found independently by both me and Codex. One-character fix.
- **P1 — Disconnect race not fully closed for the disconnect-then-reconnect window.** `refresh_locked()` now re-reads `atmosphere_connection` and bails if the row is gone (good), but it only checks for *any* `refresh_token` — not the same refresh token. Failure chain: admin clicks Disconnect mid-refresh → row deleted → admin immediately reconnects to a different account → new connection row exists with new tokens → in-flight refresh returns, the guard at `class-client.php:758` passes because `refresh_token` is non-empty (it's the NEW one), and `update_option` at line 773 overwrites the new account's `access_token`/`expires_at` with the OLD account's rotated values. End state: the new connection's encrypted access token belongs to the old account. Codex finding.
- **P1 — Async-revocation queue lives in `wp_options['cron']` even when WP-Cron is disabled or the plugin is deactivated.** The new `atmosphere_revoke_refresh_token` hook is deliberately omitted from `get_cron_hooks()` so `clear_scheduled_hooks()` doesn't wipe the queued event during `disconnect()`. The trade-off is that *every* exit path that calls `clear_scheduled_hooks()` (deactivate, uninstall) leaves the queued event behind with encrypted refresh-token + DPoP-key ciphertexts in `wp_options['cron']` until WP-Cron eventually fires it — which never happens on cron-disabled sites or after deactivation (the callback is no longer registered). Acceptable in steady-state, but the docblock at `functions.php:249-252` is incorrect: WP doesn't auto-drop cron rows for deregistered hooks. Codex finding. Suggested mitigation: also clear `atmosphere_revoke_refresh_token` from `wp_options['cron']` at deactivate/uninstall (after the revocation worker has fired or as a fail-safe cleanup at that point), or accept the residual and update the docblock.
- **P1 (carry-over from prior review) — One-shot cron handlers don't reschedule on `atmosphere_refresh_locked`.** Not addressed in these commits. `access_token()` waits ~5s for the lock holder; refresh worst-case is 2×15s. `publish_comment` / `delete_records` / `delete_comment_record` etc. log the soft error and silently drop the operation. Codex flagged this in the previous review; it survived this round.

### Lower-severity

- **P2 — TID first-write path bypasses the atomic CAS.** When `$persisted === 0`, `class-tid.php:128` falls through to `update_option()` instead of the SQL CAS at line 142. Two first publishers can both observe `persisted = 0`, both call `update_option()` with their respective `$ts`, and the later-arriving (smaller-microtime) write wins, regressing the floor. Subsequent calls work because they take the CAS branch — but the first-write window is precisely the bootstrap moment when concurrency probability is highest (e.g., a backfill kicking off N parallel publishers).
- **P2 — `comments_open` UI gate claim in `class-reaction-sync.php:520` is overstated.** The docblock says "rely on the standard `comments_open` gate in the template / REST layer to control whether those rows render on the front end." In default themes that gate exists (`comments.php` typically returns early when `! comments_open()`), but the plugin doesn't actually register any rendering filter to ensure that's true on themes that bypass `comments_template()` or for the REST `/wp/v2/comments` endpoint (which returns comments regardless of `comment_status`). On those surfaces, imported reactions WILL appear on closed-comments posts. Either register an explicit display-side filter, or update the docblock to acknowledge it's the theme's job.
- **P2 — Revocation origin check is host-only and ignores port.** `https://auth.example.com:8443/token` and `https://auth.example.com/revoke` are treated as same-origin. Probably acceptable for typical OAuth deployments, but worth either an explicit decision in the docblock or a tightening to scheme+host+port.

### Residual / unchanged from prior review

- Stale-lock theft in `lock()` remains ownerless (`unlock()` blindly deletes the row). Not addressed; probably acceptable given the 90s TTL and the surrounding `needs_reauth` recovery path.
- `wait_for_token_refresh()` waiter visibility on default WP installs without persistent object cache — still pre-existing.
- `paginate()` watermark advance on mid-stream `WP_Error` — still pre-existing.

### Test coverage added 🟢

Author added 29 new tests covering the lock, rate-limit, revoke, and reaction-moderation paths. The deferred-test gaps from the prior review are largely closed.

---

Run artifacts: `/tmp/atmosphere-codex/output2.jsonl`, `extracted2.md`. Lint clean; full PHPUnit suite green (383 / 383).

includes/class-atmosphere.php
```diff
@@ -157,6 +157,19 @@ public function init(): void {
...
+		 * Async refresh-token revocation, scheduled by
+		 * `Client::disconnect()`. The callback is registered for every
+		 * request so the worker fires correctly when WP-Cron picks up
+		 * the queued event even though the local connection is gone.
+		 */
+		\add_action(
+			'atmosphere_revoke_refresh_token',
+			array( Client::class, 'revoke_refresh_token' ),
+			10,
+			3
```
kraftbj: **P1 — `accepted_args = 3` silently disables the new revocation origin-binding check.**

`Client::disconnect()` schedules `atmosphere_revoke_refresh_token` with FOUR args:
```php
$revoke_args = array(
    (string) $conn['refresh_token'],
    (string) $conn['dpop_jwk'],
    (string) $conn['revocation_endpoint'],
    (string) $conn['auth_server'],  // ← this one
);
```
With `accepted_args = 3` here, WordPress's filter system trims to the first 3 when invoking the callback. `Client::revoke_refresh_token()`'s 4th parameter `$auth_server_issuer = ''` then falls back to its default. The guard at `class-client.php:1137` (`if ( '' !== $auth_server_issuer )`) is consequently always FALSE, and the host-only origin check below it never fires.

Net effect: the P2 fix this commit batch added (revocation endpoint origin-binding to defend against a tampered `atmosphere_connection`) is dead code in production.

Fix: change the `3` to `4`. Found independently by Codex.

includes/oauth/class-client.php
```diff
@@ -659,24 +723,54 @@ private static function refresh_locked( array $conn ): true|\WP_Error {
...
+		$current['expires_at']   = \time() + ( $data['expires_in'] ?? 3600 );
+		$current['needs_reauth'] = false;
 
 		if ( ! empty( $data['refresh_token'] ) ) {
-			$conn['refresh_token'] = Encryption::encrypt( $data['refresh_token'] );
+			$current['refresh_token'] = Encryption::encrypt( $data['refresh_token'] );
 		}
 
-		\update_option( 'atmosphere_connection', $conn );
+		\update_option( 'atmosphere_connection', $current, false );
```
kraftbj, line 819: **P1 — Disconnect-then-reconnect window still grafts old credentials onto the new connection.**

The re-read guard at line 756–763 correctly handles the disconnect case (row gone → bail). But it only checks for `! empty( $current['refresh_token'] )` — not that the refresh_token is the same one this refresh started with.

Failure chain:
1. Cron starts `Client::refresh()` with refresh token RT-A for account A.
2. Admin clicks Disconnect (row deleted).
3. Admin immediately reconnects to a different account B (new row, refresh token RT-B).
4. The original cron worker returns from its network call. The re-read at line 756 sees a row with `refresh_token` set (it's RT-B encrypted) → guard passes.
5. Line 773 `update_option` overwrites account B's access_token / expires_at with the freshly-rotated value from A's refresh response.

Result: account B's stored access_token is actually account A's. The next publish either fails (DPoP key mismatch) or — worse, if A's refresh response also returned an updated refresh_token at line 770 — writes A's refresh_token over B's, silently re-binding the connection to the wrong account.

Fix: snapshot the `refresh_token` ciphertext at the start of `refresh_locked()` and verify it still matches `$current['refresh_token']` before writing. If it differs, the connection was rotated under us — bail with the existing `atmosphere_disconnected_mid_refresh` error. Codex finding.

includes/oauth/class-client.php
```diff
@@ -923,12 +1017,397 @@ public static function access_token(): string|\WP_Error {
...
+			 * decryption step.
+			 */
+			$revoke_args = array(
+				(string) $conn['refresh_token'],
+				(string) $conn['dpop_jwk'],
+				(string) $conn['revocation_endpoint'],
+				(string) $conn['auth_server'],
+			);
+		}
+
```
kraftbj, line 1125: **P1 — Async revocation queue leaks ciphertexts on cron-less / deactivated installs.**

The deliberate omission of `atmosphere_revoke_refresh_token` from `get_cron_hooks()` (per the comment in `functions.php:244-252`) means `clear_scheduled_hooks()` won't wipe the event we just queued — good. But it also means the event row in `wp_options['cron']` survives every cleanup path:

- **Cron disabled** (`DISABLE_WP_CRON=true` and no external trigger): event never fires; ciphertext sits in `wp_options` indefinitely.
- **Plugin deactivated before fire**: callback no longer registered. When cron eventually fires, `do_action_ref_array` finds no callback and silently no-ops, but the event row is *consumed* (WP marks it run). On long-disabled-cron installs that gap may be hours/days.
- **Plugin uninstalled before fire**: `uninstall.php` runs `clear_scheduled_hooks()` but doesn't touch this hook either. The event survives the uninstall.

The ciphertexts are encrypted with `AUTH_KEY` so the risk is bounded — but the docblock at `functions.php:249-252` claims "if the plugin is deactivated before it fires the event is dropped by WordPress automatically." That's not how WP-Cron works; the event row persists.

Fix options:
- (a) Add an explicit `wp_clear_scheduled_hook( 'atmosphere_revoke_refresh_token' )` in both deactivate and uninstall paths (after a brief attempt to run the worker inline), accepting the trade-off that pending revocations are skipped.
- (b) Inline-fire the worker synchronously on deactivate/uninstall (the fast-disconnect rationale doesn't apply during plugin teardown).
- (c) Accept the leak and update the docblock to be accurate — encrypted ciphertexts at rest in `wp_options['cron']` is the same threat model as in `wp_options['atmosphere_connection']`, both keyed by `AUTH_KEY`.

Codex finding.

includes/transformer/class-tid.php
```diff
@@ -51,16 +71,97 @@ class TID {
...
+			if ( 0 === $persisted ) {
+				/*
+				 * First-write: there is no row to UPDATE against yet,
+				 * so the standard options API creates it. Subsequent
+				 * advances go through the atomic UPDATE branch below.
+				 * `autoload=false` keeps the row out of the
+				 * always-loaded options cache; the per-process static
+				 * is what makes the hot path tight.
+				 */
+				\update_option( self::OPTION_LAST_TS, (string) $ts, false );
```
kraftbj: **P2 — First-write path is not atomic.**

The `$persisted === 0` branch falls through to `update_option()` instead of the SQL CAS used in the `else` branch at line 142. Failure chain: two backfill workers both call `generate()` for the first time on a fresh install (or after the option was deleted) → both observe `$persisted = 0` → both compute their respective `$ts` from microtime → both call `update_option(self::OPTION_LAST_TS, (string) $ts, false)`. The later-arriving write wins, which means the smaller `$ts` can clobber a larger one — exactly the regression the CAS branch defends against.

The window is narrow (only the first cross-worker call after a clean install) but it's the bootstrap moment, which is precisely when concurrency probability is highest (e.g., the backfill cron firing across multiple PHP-FPM workers).

Fix: use the same SQL CAS idiom unconditionally. `INSERT IGNORE` handles the row-doesn't-exist case atomically:
```sql
INSERT IGNORE INTO {$wpdb->options} (option_name, option_value, autoload) VALUES (%s, %s, %s)
```
If that succeeds (1 row), we're done. Otherwise fall through to the existing `UPDATE ... WHERE CAST(...) < $ts` branch. This is the same pattern `Client::lock()` uses. Codex finding.

includes/oauth/class-client.php
```diff
@@ -923,12 +1017,397 @@ public static function access_token(): string|\WP_Error {
...
+		 * to its origin.
+		 */
+		if ( '' !== $auth_server_issuer ) {
+			$revoke_host = \wp_parse_url( $revocation_endpoint, PHP_URL_HOST );
+			$issuer_host = \wp_parse_url( $auth_server_issuer, PHP_URL_HOST );
+
+			if ( ! \is_string( $revoke_host )
+				|| ! \is_string( $issuer_host )
+				|| '' === $revoke_host
+				|| \strtolower( $revoke_host ) !== \strtolower( $issuer_host )
```
kraftbj, line 1210: **P2 — Origin check is host-only; port-only differences pass.**

`wp_parse_url( $url, PHP_URL_HOST )` strips the port. So `https://auth.example.com:8443/token` (issuer) and `https://auth.example.com:4443/revoke` (tampered endpoint) compare equal here.

For typical OAuth deployments the same host = same operator assumption holds. For a multi-tenant auth server on the same domain with per-tenant ports (unusual but exists), this is exploitable.

Fix: compare scheme + host + port. `wp_parse_url` accepts `PHP_URL_PORT` separately, or use the array form and compare both fields:
```php
$revoke  = \wp_parse_url( $revocation_endpoint );
$issuer  = \wp_parse_url( $auth_server_issuer );
if ( strtolower( $revoke['host'] ?? '' ) !== strtolower( $issuer['host'] ?? '' )
    || ( $revoke['port'] ?? null ) !== ( $issuer['port'] ?? null )
) { return; }
```

includes/class-reaction-sync.php
```diff
@@ -504,6 +504,24 @@ private static function insert_reaction(
...
+		 * closed-comments posts at this gate, the watermark would
+		 * still move past them and a subsequent reopen of comments
+		 * could not recover the missed imports — the WATERMARK_GRACE
+		 * window is only ten items.
+		 *
+		 * Federated reactions are an audit record of activity that
+		 * happened on Bluesky, not a comment-form submission. Insert
+		 * the row, let the moderation pipeline below decide the
+		 * approval state, and rely on the standard `comments_open`
+		 * gate in the template / REST layer to control whether
```
kraftbj: **P2 — "Standard `comments_open` gate in the template / REST layer" claim isn't actually backed by code in this plugin.**

The docblock implies a downstream rendering gate hides imported reactions on closed-comments posts. In WordPress that gate only exists if the theme calls `comments_template()` and the template checks `! comments_open()` before rendering. Many themes do (the default Twenty-* themes), but:

- Block themes that render via `core/comments` block use the block's own rules — `core/comments` renders the comments query regardless of `comment_status`; the *form* respects `comments_open()`, but existing rows still appear.
- The REST endpoint `/wp/v2/comments` returns comments by `post` parameter regardless of `comment_status`. Anything querying that endpoint (mobile apps, feed readers, headless setups) sees imported reactions on closed-comments posts.

So on those surfaces, closed-comments posts will accumulate imported Bluesky reactions and they WILL render — contradicting the docblock's implication.

Fix options: (a) register an explicit display-side filter (e.g., `comments_array` returning empty when `! comments_open( $post_id )` for `atproto`-stamped comments); (b) skip the import when comments are closed and store a tombstone meta on the post so the watermark can still advance safely; (c) update the docblock to acknowledge the gate is the theme's job and accept that headless/REST consumers see them.

Codex finding.

includes/class-atmosphere.php
pfefferle: Fixed in 2dd0b22 — flipped `accepted_args` to 4 in `register_async_hooks`. The issuer arg now reaches `revoke_refresh_token()` instead of being trimmed by WordPress, so the host check actually runs.

includes/oauth/class-client.php
pfefferle: Fixed in 2dd0b22. The re-read in `refresh_locked()` now compares the refresh-token ciphertext against the snapshot taken at lock-acquisition via `hash_equals()`. libsodium uses a fresh nonce per `encrypt()`, so any delete/recreate (reconnect to same OR different account) or re-encryption fails the equality and the worker bails with `atmosphere_disconnected_mid_refresh`. Same compare lives in the `needs_reauth` write branch above.

Same commit also addresses the P1 #4 carry-over from your re-review body (no inline thread): `wait_for_token_refresh()` deadline bumped from 5s to `REFRESH_LOCK_TTL` (90s) so it covers the realistic worst case of two sequential 15-second `wp_safe_remote_post` calls on the `use_dpop_nonce` retry path. Cron one-shots no longer drop content under that path — the waiter either observes the holder's result or concludes the holder crashed.

includes/oauth/class-client.php
pfefferle: Fixed in 2dd0b22 by going with your "also clear at deactivate/uninstall" suggestion. Split into two helpers in `functions.php`: `clear_scheduled_hooks()` keeps the disconnect contract (omits the revoke event so `Client::disconnect()` can schedule it right after); `clear_scheduled_hooks_all()` is the new variant that *also* clears `atmosphere_revoke_refresh_token`, wired into `deactivate()` and `uninstall.php`. The orphan-ciphertext window on cron-disabled / deactivated installs is closed. Docblocks rewritten to stop claiming WP auto-drops cron rows for deregistered hooks.

includes/transformer/class-tid.php
pfefferle: Fixed in 2dd0b22. Every write now takes the CAS path first (`UPDATE … WHERE CAST(option_value AS UNSIGNED) < $ts`); if the UPDATE affects zero rows (no row exists yet OR our `$ts` isn't strictly larger), the `INSERT IGNORE` directly below it creates the row. Both statements are atomic on the UNIQUE index, so the bootstrap window is no longer worse than steady state. The `0 === $persisted` branch is gone.

includes/oauth/class-client.php
pfefferle: Fixed in 2dd0b22. Comparison is now scheme + host + port (RFC 6454 origin semantics). Missing ports normalise to 443 so `auth.example.com` and `auth.example.com:443` compare equal; `:8443` issuer correctly mismatches a `:443` revocation endpoint.

includes/class-reaction-sync.php
pfefferle: Fixed in 2dd0b22. The overclaim is gone — the comment block now acknowledges that themes bypassing `comments_template()` and the REST `/wp/v2/comments` endpoint will surface imported reactions on closed-comments posts regardless. Sites that need stricter rendering control are pointed at `the_comments` / `rest_prepare_comment` filters. The existence of the row is the federation-history record this method is responsible for; render-side policy is left to the site.

kraftbj (approved):
## Final pass — ready to merge

Verified every fix in `2dd0b22` against the corresponding finding from my prior re-review. All 6 inline P1/P2 threads and the carry-over P1 (`atmosphere_refresh_locked` retry deadline) are addressed:

| Finding | Fix | Verified |
|---|---|---|
| `accepted_args = 3` silently disabled origin check | `class-atmosphere.php:170` flipped to `4` | ✅ |
| Disconnect-then-reconnect graft on refresh | `refresh_locked()` now `hash_equals` ciphertext snapshot against current row (both happy-path + needs_reauth) | ✅ |
| Async revocation queue leaks on cron-disabled/deactivated | Split into `clear_scheduled_hooks()` (disconnect) and `clear_scheduled_hooks_all()` (deactivate + uninstall, includes revoke hook); docblock corrected | ✅ |
| Cron one-shots silently drop on `atmosphere_refresh_locked` | `wait_for_token_refresh()` deadline 5s → 90s (`REFRESH_LOCK_TTL`) | ✅ |
| TID first-write bypassed CAS | Unconditional UPDATE-CAS first, INSERT IGNORE fallback for bootstrap (both atomic on UNIQUE index) | ✅ |
| Origin check host-only | Scheme + host + port; missing port normalizes to 443 (RFC 6454) | ✅ |
| `comments_open` UI-gate overclaim | Docblock rewritten to acknowledge `comments_template()`-bypassing themes and `/wp/v2/comments` REST surface; points readers at `the_comments` / `rest_prepare_comment` for stricter control | ✅ |

The `hash_equals` ciphertext-snapshot approach is stronger than the variant I suggested — `Encryption::encrypt()` calls `random_bytes(SODIUM_CRYPTO_SECRETBOX_NONCEBYTES)` per invocation, so the comparison also catches re-encryption (e.g., from `sync_connection_handle()`), not just delete-then-recreate.

The trunk merge brought in PR #66 (Force HTTPS for OAuth callback + `atmosphere_connection` no longer autoloaded). The `autoload=false` flip is particularly nice: encrypted tokens are no longer in the always-loaded `alloptions` payload, AND it incidentally closes the pre-existing waiter-visibility race on default WP installs (each `wait_for_token_refresh()` poll now hits SQL fresh rather than reading a stale request-cached alloptions value).

### Coverage

- All 24 review threads on this PR resolved.
- CI green: phpcs, phpunit on PHP 8.2/8.3/8.5 across WP 6.2 / latest / trunk, changelog, repo gardening.
- 383 tests / 1195 assertions in the latest local run (29 new tests added across this batch).
- Lint clean.

### Acceptable residuals (not blockers)

- Stale-lock theft in `lock()` is still ownerless — `unlock()` blindly deletes. The 90s TTL plus `needs_reauth` recovery make the worst case bounded, and the realistic refresh worst case is ~30s, so the steal window is mostly theoretical. Worth a follow-up but not a release blocker.
- TID cross-worker monotonic guarantee for *literally simultaneous* microsecond-perfect calls still falls back to the 10-bit `clock_id` (1/1024 collision). The docblock is honest about this.
- `paginate()` watermark advance on mid-stream `WP_Error` — pre-existing, scope creep to fix here.

LGTM 🚀

PR #72, opened 2026-05-19
Add atmosphere_post_embed filter and generalize image blob upload

kraftbj:
Refs [DOTCOM-17143](https://linear.app/a8c/issue/DOTCOM-17143).

## Proposed changes:

Open a focused seam for swapping the embed attached to a Bluesky post record. The new `atmosphere_post_embed` filter fires for every composition strategy — including short-form, where the default is `null` so consumers can attach an embed that today's transformer never builds.

Use cases this unlocks downstream:

- `app.bsky.embed.images` for photo-shaped WordPress posts (the immediate driver — federating WP photo posts to Flashes / Pinksky as native image posts instead of link cards).
- `app.bsky.embed.video` for video posts.
- `app.bsky.embed.record` quote embeds.
- Suppressing the default external card on specific post types.

Before this filter, the only way to swap an embed was the wider `atmosphere_transform_bsky_post`, which forces the consumer to be aware of the whole record shape (text, facets, langs, tags, reply refs) and fires per-record in teaser threads. `atmosphere_post_embed` lets a listener do the embed swap as a one-liner.

### Filter contract

Signature: `apply_filters( 'atmosphere_post_embed', ?array \$embed, WP_Post \$post, string \$strategy )`.

- `\$embed` is the default for this strategy: `null` for short-form, an `app.bsky.embed.external` card for link-card and teaser-thread.
- `\$strategy` is one of `short-form`, `link-card`, `teaser-thread`.
- Return null to suppress, an array with a non-empty string `$type` key to replace, the default to no-op.
- Returns that don't match that contract — non-array, empty array, or array missing/with a non-string `$type` — are rejected with `_doing_it_wrong` and the pre-filter value is restored. Same posture as `atmosphere_transform_bsky_post`, and keeps every composition call site (`transform()`, `record_for_link_card()`, the teaser-thread entries) treating half-formed returns the same way.
- Does not fire on redacted (password-protected / non-public) transforms — matches `atmosphere_is_short_form_post` and `atmosphere_transform_bsky_post`.

Wired through every existing embed callsite: short-form path in `transform()`, link-card default in `transform()`, the link-card branch of `build_long_form_records()`, the collapsed single-record teaser thread, and the terminal CTA entry of multi-entry teaser threads.

### Related cleanups

- `Post::upload_thumbnail()` becomes a backward-compatible alias for `Post::upload_image_blob()`. The body was already attachment-agnostic; the rename clarifies it for the new caller shape (uploading non-featured image attachments for `embed.images`). Existing callers in Publication and Document keep working.
- New `Post::get_attachment_aspect_ratio( int \$attachment_id ): ?array` returns `[ 'width' => int, 'height' => int ]` from `wp_get_attachment_metadata()` — the dict shape `app.bsky.embed.images` expects. Centralizing it here keeps every downstream image-embed consumer on the same metadata path.

### Other information:

Adds 13 new tests to `Test_Post`:
- filter fires for long-form with the external card as default;
- filter fires for short-form with `null` as default (the new capability);
- filter return replaces the embed;
- filter can attach an embed to a short-form post (the new capability);
- filter returning null suppresses the default;
- non-array, non-null returns are rejected;
- filter does not fire on redacted transforms;
- filter fires for the link-card branch of `build_long_form_records()`;
- filter fires for the terminal CTA of a teaser-thread;
- `upload_thumbnail()` is identical to `upload_image_blob()`;
- `get_attachment_aspect_ratio()` happy path, missing metadata, zero-dimension guard.

## Testing instructions:

* `composer lint` — clean.
* `composer test` (or `composer test:wp-env`) — full suite, including the 13 new cases.
* Manual: with this branch active, install a tiny plugin that does

  ```php
  add_filter( 'atmosphere_post_embed', function ( \$embed, \$post, \$strategy ) {
      if ( 'short-form' === \$strategy ) {
          return null; // no embed
      }
      // replace default external card with a record-quote stub
      return [
          '\$type' => 'app.bsky.embed.record',
          'record' => [ 'uri' => 'at://did:plc:test/app.bsky.feed.post/123', 'cid' => 'bafyfake' ],
      ];
  }, 10, 3 );
  ```

  Publish a titled post and confirm the outgoing record's `embed.\$type` is `app.bsky.embed.record` (long-form path), not `external`. Toggle the filter off and confirm the default external card returns.

### Changelog entry

Changelog entry committed in `.github/changelog/add-post-embed-filter`.


pfefferle:
@kraftbj I did some small changes based on the feedback of co-pilot, let me know what you think!

kraftbj:
Thanks @pfefferle — both changes look great to me. The empty-array/`$type` gate fixes a real inconsistency between the three composition call sites (the truthy `if ( $embed )` paths would have silently dropped an empty-array return while the thread-entry `null !== $embed` path would have attached it), and the `is_numeric()` guard on `get_attachment_aspect_ratio()` lines the implementation up with what the docblock claimed.

I pushed one follow-up commit (5b8667c) that just brings the helper docblock, the filter docblock, and the changelog entry in sync with the widened rejection contract — they all still described the old "non-array only" gate. No behavior change.

Let's ship it! 🚀

pfefferle (approved):

PR #85, opened 2026-05-21
Preserve atmosphere_identity across OAuth disconnect

kraftbj:
## Proposed changes

`OAuth\Client::disconnect()` was deleting `atmosphere_identity` alongside the OAuth session. That option is the bidirectional-verification anchor read by the `/.well-known/atproto-did` route and the publication link tag. Wiping it on disconnect creates a chicken/egg trap for sites that have adopted a custom domain handle:

1. User connects with `alice.bsky.social`, then applies `example.com` as their handle via the "use your domain" panel.
2. User disconnects later.
3. `atmosphere_identity` gets deleted, so `https://example.com/.well-known/atproto-did` now 404s.
4. User tries to reconnect with `example.com`. `Resolver::handle_to_did()` checks DNS TXT (not set) and the HTTPS well-known (now 404) and returns `Could not resolve handle example.com to a DID.`
5. The user is locked out of reconnecting with their own handle. The PDS-side handle is still `example.com`, but no public mechanism publishes the DID anymore.

The identity/connection split was introduced specifically so token failures (refresh rejected, decrypt failed) wouldn't break the verification headers — see the docstring on `get_identity()`. Explicit disconnect was violating that same invariant.

This PR drops only the OAuth session on disconnect. Identity is preserved; a subsequent reconnect refreshes it in place, and sites that genuinely want to forget the DID can clear `atmosphere_identity` separately.

### Other information

## Testing instructions

1. On a fresh install, complete a Bluesky OAuth connect (any handle).
2. `wp option get atmosphere_identity` — confirm it contains `did`, `handle`, `pds_endpoint`.
3. Click Disconnect from the Atmosphere settings page.
4. `wp option get atmosphere_identity` — **before this PR:** option is missing. **After this PR:** the option still contains the same `did`, `handle`, `pds_endpoint`.
5. `curl https://your-site.tld/.well-known/atproto-did` — should return your `did:plc:...`, both before and after disconnect.
6. Reconnect with the original handle. OAuth completes and overwrites identity in place (same shape).

Also covered by the new `test_disconnect_preserves_identity_for_handle_resolution` regression test, alongside existing `test_disconnect_clears_all_cron_hooks` and `test_disconnect_sweeps_stale_publication_uri_option` which continue to pass.

### Changelog entry

A `.github/changelog/preserve-identity-on-disconnect` entry is included in the diff.

<details>

#### Significance

- [x] Patch

#### Type

- [x] Fixed

#### Message

Preserve the persisted AT Protocol identity when disconnecting so the `.well-known/atproto-did` route keeps serving. Sites that adopted a custom domain handle (`example.com` instead of `alice.bsky.social`) could not reconnect after a disconnect because the wiped identity left the well-known route 404'ing, and the handle resolver had nothing to find. Disconnect now drops only the OAuth session; reconnecting refreshes identity in place.

</details>


pfefferle (commented):
Findings:

- **Medium:** `Admin::handle_disconnect()` still calls `Handle::maybe_revert_on_disconnect()` before `Client::disconnect()`, and a successful revert changes the PDS handle back to the previous handle before identity preservation matters. That means this PR does not fully cover the documented "use your domain, disconnect, reconnect with domain" path when the revert succeeds. Either the admin disconnect flow needs to stop reverting in this case, or the PR scope/test should clarify it only fixes manual/failed-revert custom-domain cases.

- **Low:** `needs_reauth()` returns true whenever identity exists but credentials are gone. With this PR, an intentional disconnect now leaves identity behind, so the global "session has expired / reconnect" warning will appear after a deliberate disconnect. That state probably needs distinct copy or an explicit-disconnect marker.

Checks run:

- `npm run env-test -- --filter=disconnect` passed: 7 tests, 20 assertions.
- `composer lint` passed.
- `npm run env-test` passed: 416 tests, 1253 assertions.

pfefferle (approved):

PR #96, opened 2026-05-27
Fix: short-form posts now include images when published to Bluesky

jeherve:
Fixes [CM-796](https://linear.app/a8c/issue/CM-796/short-form-posts-drop-images-when-distributed-to-bluesky)

## Proposed changes:

Short-form Bluesky posts (aside, status, quote — any format where the post body is shipped verbatim as text) were silently dropping their images. The link-card embed path only ran for the long-form composition, so a short-form post with images shipped with no embed at all.

This PR makes short-form posts build an `app.bsky.embed.images` record from:

1. `core/image` blocks parsed from `post_content` — recursive into `innerBlocks`, deduped, document order, capped at 4 per the AT Protocol limit.
2. The featured image (`get_post_thumbnail_id`) as a fallback when no in-body images are found.

Redacted (password-protected) posts continue to ship no embed. The `atmosphere_post_embed` filter contract is preserved — third-party listeners can still suppress, replace, or augment the default.

### Hardening from review:

* The walker that collects in-body image IDs is capped at 32 entries so an adversarial `post_content` packed with thousands of `core/image` blocks can't grow the working array linearly.
* Regression tests added for the `> 0` / `isset` ID guards and for the always-present `alt` field (Lexicon requirement).
* `atmosphere_post_embed` filter docblock corrected to reflect the new short-form default (was `null`; is now an `app.bsky.embed.images` record when the post has images).

### Out of scope:

* Classic-editor `` tags (block-editor scope only — documented in code).
* The pre-existing 1 MB blob cap in `upload_image_blob()` (AT Protocol Lexicon now allows 2 MB — separate uplift opportunity).

### Other information:

## Testing instructions:

1. Connect ATmosphere to a Bluesky test account.
2. In the WordPress editor, create a new post with the **Aside** post format (or leave the title empty for any post type).
3. Add one or more `core/image` blocks to the post body — uploaded from the media library so they have attachment IDs. Publish.
4. Open the resulting post on Bluesky. Confirm the post body is shipped as text and the images are attached as a native image embed (up to 4, in document order).
5. Repeat with a post that has **no** in-body images but **does** have a featured image. Confirm the featured image is attached instead.
6. Repeat with a password-protected aside that has a featured image. Confirm Bluesky receives the redacted record with no embed and no leaked image.
7. Optional: drop a filter on `atmosphere_post_embed` that returns `null` and confirm the embed is suppressed; one that returns an array with a non-empty `$type` and confirm it replaces the default.

### Changelog entry

Already committed in `.github/changelog/fix-short-form-image-embed`.

### DIFF
--- a/.github/changelog/fix-short-form-image-embed
+++ b/.github/changelog/fix-short-form-image-embed
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Aside, status, and other short-form posts now include their images when published to Bluesky.
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -410,6 +410,13 @@ class Post extends Base {
 	 * out of scope; consumers needing those can wire them in via the
 	 * `atmosphere_post_embed` filter.
 	 *
+	 * The walker stops collecting once 32 IDs have been gathered — well
+	 * above the 4-image AT Protocol cap, but enough headroom that dedupe
+	 * still preserves document order on realistic posts. Bounds the
+	 * memory profile so an attacker-controlled `post_content` packed with
+	 * thousands of `core/image` blocks can't grow the working array
+	 * past a constant ceiling.
+	 *
 	 * @return int[]
 	 */
 	private function collect_image_attachment_ids(): array {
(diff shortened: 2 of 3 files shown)


pfefferle:
Pushed `f7a1ae5` to address the breadth/depth and test-coverage observations from my review pass:

- Added a `$max_depth = 16` cap to the walker in `collect_image_attachment_ids()`. The existing 32-ID breadth cap only bounds wide-and-shallow inputs; a tree of hundreds of nested `core/group` wrappers with no images never accumulates IDs and never trips the breadth guard. 16 leaves ample headroom over realistic theme nesting (cover → group → columns → column → group → image is six) while keeping the worst-case stack bounded against adversarial `post_content`.
- Added `test_short_form_image_collector_stops_at_breadth_cap` (34 images, tail at position 35 must not appear in the embed) and `test_short_form_image_collector_stops_at_depth_cap` (image inside 30 nested groups must be dropped cleanly, no fatal).

Copilot's two findings are unaddressed and worth a separate look from you before merging:

1. Image-only short-form posts (aside with only a `core/image` block, no paragraph) still fall through to the long-form link-card branch because `build_short_form_text()` returns empty for image-only blocks. The fix would touch the `transform()` branching at `class-post.php:159-166`.
2. Attachment alt text isn't truncated to AT Protocol's `maxGraphemes: 1000` / `maxLength: 2000`. A super-long WordPress alt would fail Lexicon validation and reject the record at the PDS. One-liner with `truncate_text()`.

pfefferle (approved):

PR #105, opened 2026-06-01
Add publication strongRef to link-card associatedRefs

pfefferle:
## Proposed changes:

Posts ATmosphere publishes to Bluesky build their `app.bsky.embed.external` link card by hand rather than letting Bluesky's share UI fetch the URL. The Lexicon defines `external.associatedRefs` as an array of strongRefs of the records that backed the embed, and Bluesky's manual-share UI already populates it for any URL whose HTML carries our `<link rel=\"site.standard.publication\">` / `<link rel=\"site.standard.document\">` tags. A post we publish ourselves should advertise the same field — otherwise the bsky-side UI loses the navigable link back to the standard.site source.

Compare what Bluesky's share UI ships for the [same URL](https://bsky.app/profile/jeherve.com/post/3mn6layvcfs2e) vs. what [our publisher](https://bsky.app/profile/did:plc:4i6hvdii3km3kbnj3losmwnt/post/3mn6kp4mh4oum) ships — the manual share has `associatedRefs` with strongRefs for both the document and the publication; the Atmosphere post has neither.

This PR handles the **publication** half. The publication is written once at connect (and re-synced on theme + site-option changes), so its CID is observable from the existing `putRecord` response — we just need to capture it.

* `Publication::OPTION_CID` (new option `atmosphere_publication_cid`) stores the CID the PDS returns on each successful `Publisher::sync_publication()`. The CID rotates on every successful putRecord (the publication's content hash changes whenever the record's content does), so the option is overwritten every call — not just on first write.
* `Publication::get_strong_ref()` builds the `com.atproto.repo.strongRef` from the stored TID + CID + connected DID, or returns `null` if any of those are missing.
* `Post::build_embed()` appends the ref as the sole entry in `external.associatedRefs`. Missing state → field is simply omitted; the embed is still valid.
* `uninstall.php` sweeps the new option alongside the existing publication state.

### Out of scope

The **document** strongRef (the per-post companion) needs its CID before the atomic `applyWrites`, which is the chicken-and-egg case I called out separately. Two paths there — a follow-up `applyWrites#update`, or a client-side DAG-CBOR encoder so we can compute the document CID locally — both are bigger changes that warrant their own review. Splitting them out so this user-visible win can land now.

### Existing installs

`get_strong_ref()` returns `null` until the next `sync_publication` trigger captures the CID. The plugin already re-syncs on theme changes, site URL updates, etc., so most installs pick it up on the first post-merge admin save. Until then, posts publish without `associatedRefs` (same as today). No backfill needed.

### Other information:

Three new test groups:

* `Test_Publication`: `get_strong_ref()` returns the right shape with full state, null when CID is missing, null without identity.
* `Test_Post`: long-form embed includes the ref when state is present, omits it when CID is missing, never leaks onto short-form posts (which use `app.bsky.embed.images` and don't have `associatedRefs` in their lexicon).
* `Test_Publisher`: `sync_publication()` captures the CID into the option and overwrites it on each call.

## Testing instructions:

1. Connect a site to ATmosphere.
2. Trigger a publication sync (Settings → General → save anything, or change the active theme — either fires the existing `schedule_publication_sync` hook).
3. \`wp option get atmosphere_publication_cid\` — confirm a `bafyrei…` value is now stored.
4. Publish a new titled blog post and inspect the resulting bsky record via `pdsls.dev` or `curl -s https://public.api.bsky.app/xrpc/com.atproto.repo.getRecord?repo=<did>&collection=app.bsky.feed.post&rkey=<rkey>`.
5. Confirm `value.embed.external.associatedRefs[0]` is a `com.atproto.repo.strongRef` whose `uri` matches `at://<did>/site.standard.publication/<your-publication-tid>` and whose `cid` matches the captured option.
6. Delete the cached CID (\`wp option delete atmosphere_publication_cid\`), publish another post, confirm the new record's `external` has no `associatedRefs` field — graceful degradation rather than a malformed strongRef.

Automated:
* `composer lint`
* `npm run env-test`

### Changelog entry

Already committed at `.github/changelog/add-publication-strong-ref` (Patch / Added).

### DIFF
--- a/.github/changelog/add-publication-strong-ref
+++ b/.github/changelog/add-publication-strong-ref
@@ -1,0 +1,4 @@
+Significance: patch
+Type: added
+
+Posts shared to Bluesky now link back to the site's publication record, so Bluesky can show your site's source alongside the link preview.
--- a/.github/changelog/add-publication-strong-ref
+++ b/.github/changelog/add-publication-strong-ref
@@ -1,4 +1,4 @@
 Significance: patch
 Type: added
 
-Posts shared to Bluesky now link back to the site's publication record, so Bluesky can show your site's source alongside the link preview.
+Posts shared to Bluesky now link back to both the site's publication record and the per-post document record, so Bluesky can show your site's source and richer document metadata alongside the link preview.
--- a/includes/class-publisher.php
+++ b/includes/class-publisher.php
@@ -1346,7 +1346,7 @@ class Publisher {
 		 * reconnects, so the record always lands at the same address
 		 * for the active owner.
 		 */
-		return API::post(
+		$result = API::post(
 			'/xrpc/com.atproto.repo.putRecord',
 			array(
 				'repo'       => $did,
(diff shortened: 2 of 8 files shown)


pfefferle:
Closing — investigation upstream landed: PDS has both strongRefs after the follow-up `applyWrites#update`, but Bluesky's AppView serves an older indexed version of the record and never picks up the update (see `bafyreid4dwo6mr...` on the PDS vs `bafyreid2itlvpd...` on `public.api.bsky.app` for post `3mn7u77lp4dns`). `source` / `associatedProfiles` are only emitted from the initial-create indexing, so the document strongRef has to be in the original atomic write — which needs client-side CID computation to know the document's CID before that write.

Re-opening as a fresh PR once the DAG-CBOR encoder + CID computation lands.

PR #128, opened 2026-06-11
Never log exception messages from DPoP proof generation

pfefferle:
## Proposed changes:
* Stop logging `$e->getMessage()` from the catch block in `DPoP::create_proof()`. The exception is thrown with the private-key JWK in scope, and OpenSSL error strings are not guaranteed to be free of key material; the log line is visible under `WP_DEBUG` and on stderr.
* Log a safe summary instead — exception class plus `basename(file):line` of the throw site — built by a new `describe_throwable()` helper whose docblock forbids reintroducing the message.
* Cover the catch path with a test that throws from the nonce-transient lookup, and pin the no-message property directly on the helper (the wp-env tests environment runs with `WP_DEBUG` off, so the logged line is not observable end-to-end there).

Addresses finding ATM-M-05 from the in-house security audit. This is hardening — there is no known case of key material in these messages today.

### Other information:

## Testing instructions:

* Run `npm run env-test -- --filter=Test_DPoP` — all 7 tests pass, including the two new ones.
* Read the diff of `includes/oauth/class-dpop.php` and confirm the catch block no longer references `getMessage()`.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>

### DIFF
--- a/.github/changelog/fix-dpop-exception-message-logging
+++ b/.github/changelog/fix-dpop-exception-message-logging
@@ -1,0 +1,4 @@
+Significance: patch
+Type: security
+
+Hardened error logging so details from cryptographic failures are never written to the log.
--- a/includes/oauth/class-dpop.php
+++ b/includes/oauth/class-dpop.php
@@ -135,23 +135,24 @@ class DPoP {
 
 			return self::sign_es256( $header, $payload, $jwk );
 		} catch ( \Throwable $e ) {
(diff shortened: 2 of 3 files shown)


jeherve (approved):
LGTM! 🚢

PR #129, opened 2026-06-11
Detach the scoped allowed_redirect_hosts filter even when the redirect throws

pfefferle:
## Proposed changes:
* Wrap the `wp_safe_redirect()` call in `Sanitize::handle()` in `try`/`finally` so the scoped `allowed_redirect_hosts` filter is detached even when a `wp_redirect` filter throws instead of returning (a test interceptor or a custom `wp_die` handler). Previously the exception propagated past `remove_filter()`, leaving the auth-server host allowed for any later redirect in the same request.
* Add a regression test that throws from the `wp_redirect` filter and asserts the host is no longer in `allowed_redirect_hosts` afterwards — it fails on trunk and passes with the fix.

Addresses finding ATM-M-07 from the in-house security audit. Production behavior on the happy path is unchanged (`exit` still ends the request).

### Other information:

## Testing instructions:

* Run `npm run env-test -- --filter=Test_Sanitize_Handle` — all 5 tests pass, including the new one.
* Optionally revert the `includes/class-sanitize.php` hunk and re-run: the new test fails with `auth.example.com` still present in `allowed_redirect_hosts`.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>

### DIFF
--- a/.github/changelog/fix-scope-redirect-hosts-filter
+++ b/.github/changelog/fix-scope-redirect-hosts-filter
@@ -1,0 +1,4 @@
+Significance: patch
+Type: security
+
+Hardened the Bluesky connect flow so a redirect interruption cannot loosen redirect safety checks for the rest of the request.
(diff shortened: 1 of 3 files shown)


jeherve (approved):
This should be good to merge! 🚢

PR #134, opened 2026-06-12
Resolve link facets when importing Bluesky replies as comments

### LINKED ISSUE #132: Imported Bluesky comments lose link URLs (facets not applied on inbound)
## Problem

When a Bluesky reply containing a link is imported as a WordPress comment, the link is reduced to Bluesky's truncated, non-clickable display string (e.g. \`bsky.app/profile/jere...\`) and the real URL is lost.

## Root cause

This is **not** the plugin stripping content — it's how AT Protocol represents links. The post record's \`text\` field stores only the *display* string (Bluesky shortens long URLs for readability), while the real URL lives in a separate \`facets\` array that maps a byte range in \`text\` to the full \`uri\`.

Example record (\`app.bsky.feed.post/3mnzucbkwfs2h\`):

\`\`\`json
"text": "...just copied / pasted into SkyPress :)\n\nbsky.app/profile/jere...",
"facets": [{
  "index": { "byteStart": 112, "byteEnd": 136 },
  "features": [{
    "$type": "app.bsky.richtext.facet#link",
    "uri": "https://bsky.app/profile/jeremy.herve.bzh/post/3mnzu7nvcss2e"
  }]
}]
\`\`\`

The inbound reaction-sync path (\`includes/class-reaction-sync.php:478\`) takes \`$record['text']\` verbatim and never reads \`$record['facets']\`, so the comment stores the lossy display string with no way to reach the real link.

The plugin already has a \`Transformer\Facet\` class, but it's **outbound only** (\`extract()\` builds facets from WP text when *publishing*). There is no inverse that *applies* facets when *importing*.

## Proposed fix

Add the inverse of \`Facet::extract()\` — a \`Facet::apply( $text, $facets )\` that walks facets in byte order and rebuilds the text:

- **link** facets → replace the byte range with \`<a href="{uri}">{display}</a>\`
- **mention** facets → link to the profile
- **tag** facets → optionally link the hashtag, or leave as-is

Call it at \`class-reaction-sync.php:478\` before passing \`$text\` into \`insert_reaction()\`. \`wp_kses_post()\` already runs on the content downstream, so emitting \`<a>\` tags is safe.

### Implementation notes

- **Byte offsets, not character offsets.** Facet indexes are UTF-8 byte ranges — reassembly must use byte-based \`substr()\`, never \`mb_substr()\`.
- **Sort by \`byteStart\` and splice in order**, copying the untouched gaps between ranges (mirroring how \`extract()\` sorts on the way out).
- Add a regression test using the record above as a fixture.

jeherve:
Fixes #132

## Proposed changes:

* Bluesky stores a reply's `text` as a *display* string — long URLs are truncated to e.g. `bsky.app/profile/jere...` while the real target lives only in the record's `facets` array — so `Reaction_Sync` was importing the lossy, unclickable preview.
* Adds `Facet::apply()` (the inverse of `Facet::extract()`), which splices the facet byte ranges back into the text as anchors for links, mentions, and hashtags, and calls it on the inbound reply path before the text becomes comment content.
* Reassembly is byte-based to match AT Protocol's UTF-8 facet offsets, and malformed or out-of-bounds ranges are skipped rather than corrupting surrounding text; only generated `href`s are escaped, with the result still flowing through the existing `wp_kses_post()` gate.

### Other information:

## Testing instructions:

* From Bluesky, reply to that post with a message containing a long link (long enough that Bluesky truncates it to `…`), e.g. paste another post's URL.
* Run the reaction sync (`wp atmosphere sync-reactions`, or wait for the scheduled run) and open the imported comment.
* **Before:** the comment ends with a dead `bsky.app/profile/jere...` string. **After:** it shows a working link to the full URL.
* Existing comments imported before this change keep their old text — the fix applies to new imports.

### Changelog entry

A changelog entry is already committed on this branch (`.github/changelog/fix-comment-link-facets`), so no auto-entry is needed.

### DIFF
--- a/includes/class-reaction-sync.php
+++ b/includes/class-reaction-sync.php
@@ -488,8 +488,13 @@ class Reaction_Sync {
 		 * target lives only in `facets`. Resolve those byte ranges back
 		 * into anchors before the text becomes comment content; otherwise
 		 * imported replies keep the lossy, unclickable display string.
+		 *
+		 * `facets` is untrusted PDS JSON: guard the type here so a
+		 * present-but-non-array value can't fatal the typed `apply()` call
+		 * and silently kill this notification's cron sync.
 		 */
-		$text = Facet::apply( $text, $record['facets'] ?? array() );
+		$facets = $record['facets'] ?? array();
+		$text   = Facet::apply( $text, \is_array( $facets ) ? $facets : array() );
 
 		/**
 		 * Filters whether a reply should be synced as a WordPress comment.
--- a/includes/class-reaction-sync.php
+++ b/includes/class-reaction-sync.php
@@ -11,6 +11,7 @@ namespace Atmosphere;
 \defined( 'ABSPATH' ) || exit;
 
 use Atmosphere\OAuth\Client;
+use Atmosphere\Transformer\Facet;
 use Atmosphere\Transformer\Post as BskyPost;
 
 /**
@@ -480,6 +481,15 @@ class Reaction_Sync {
 		if ( '' === $text ) {
 			return false;
 		}
+
+		/*
+		 * Bluesky stores `text` as the display string — long URLs are
+		 * truncated to e.g. `bsky.app/profile/jere...` and the real
+		 * target lives only in `facets`. Resolve those byte ranges back
+		 * into anchors before the text becomes comment content; otherwise
+		 * imported replies keep the lossy, unclickable display string.
+		 */
+		$text = Facet::apply( $text, $record['facets'] ?? array() );
 
 		/**
 		 * Filters whether a reply should be synced as a WordPress comment.
--- a/includes/transformer/class-facet.php
+++ b/includes/transformer/class-facet.php
@@ -77,6 +77,103 @@ class Facet {
 	}
 
 	/**
+	 * Reassemble rich text by applying facets back onto a plain-text string.
+	 *
+	 * The inverse of {@see Facet::extract()}. AT Protocol stores a post's
+	 * `text` as the *display* string — Bluesky truncates long URLs (e.g.
+	 * `bsky.app/profile/jere...`) — while the real target lives in the
+	 * `facets` array, which maps a byte range in `text` to a feature
+	 * (link `uri`, mention `did`, or `tag`). Without this step an imported
+	 * comment keeps only the lossy display string.
+	 *
+	 * Link and mention features become anchors; tags become hashtag-search
+	 * links; the byte ranges between facets are copied through untouched.
+	 * The result is an HTML fragment intended to be passed through
+	 * `wp_kses_post()` by the caller (as the reaction-sync path does), so
+	 * only the generated `href` attributes are escaped here.
+	 *
+	 * @since unreleased
+	 *
+	 * @param string $text   Plain-text display string from the record.
+	 * @param array  $facets Facet array from the record, as stored on the PDS.
+	 * @return string Text with facets resolved to anchors.
+	 */
+	public static function apply( string $text, array $facets ): string {
+		if ( empty( $facets ) ) {
+			return $text;
+		}
+
+		\usort(
+			$facets,
+			static fn( $a, $b ) => ( $a['index']['byteStart'] ?? 0 ) <=> ( $b['index']['byteStart'] ?? 0 )
+		);
+
+		// Facet indexes are UTF-8 byte offsets, so splice in byte space.
+		$length = \strlen( $text );
+		$cursor = 0;
+		$result = '';
+
+		foreach ( $facets as $facet ) {
+			$start = $facet['index']['byteStart'] ?? null;
+			$end   = $facet['index']['byteEnd'] ?? null;
+
+			/*
+			 * Skip ranges that are malformed, empty, out of bounds, or
+			 * that overlap a facet we've already consumed. Dropping a bad
+			 * range leaves its display text in place rather than corrupting
+			 * the surrounding bytes.
+			 */
+			if ( ! \is_int( $start ) || ! \is_int( $end ) || $start < $cursor || $start >= $end || $end > $length ) {
+				continue;
+			}
+
+			$result .= \substr( $text, $cursor, $start - $cursor );
+			$result .= self::render_feature( $facet['features'][0] ?? array(), \substr( $text, $start, $end - $start ) );
+			$cursor  = $end;
+		}
+
+		return $result . \substr( $text, $cursor );
+	}
+
+	/**
+	 * Render a single facet feature around its display text.
+	 *
+	 * Unknown feature types fall back to the display text unchanged, so a
+	 * facet type we don't handle never drops the text it annotated.
+	 *
+	 * @param array  $feature Facet feature (first entry of a facet's `features`).
+	 * @param string $display Display text the facet covers.
+	 * @return string HTML fragment, or the display text unchanged.
+	 */
+	private static function render_feature( array $feature, string $display ): string {
+		switch ( $feature['$type'] ?? '' ) {
+			case 'app.bsky.richtext.facet#link':
+				$uri = $feature['uri'] ?? '';
+				if ( '' === $uri ) {
+					return $display;
+				}
+				return '<a href="' . \esc_url( $uri ) . '">' . $display . '</a>';
+
+			case 'app.bsky.richtext.facet#mention':
+				$did = $feature['did'] ?? '';
+				if ( '' === $did ) {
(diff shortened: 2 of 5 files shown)


pfefferle:
Nice one!

includes/transformer/class-facet.php
```diff
@@ -76,6 +76,103 @@ public static function for_urls( string $text, array $urls ): array {
...
+	 * @return string HTML fragment, or the display text unchanged.
+	 */
+	private static function render_feature( array $feature, string $display ): string {
+		switch ( $feature['$type'] ?? '' ) {
+			case 'app.bsky.richtext.facet#link':
+				$uri = $feature['uri'] ?? '';
+				if ( '' === $uri ) {
+					return $display;
+				}
+				return '<a href="' . \esc_url( $uri ) . '">' . $display . '</a>';
```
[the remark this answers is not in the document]
jeherve: Good catch — fixed in 41e8cdf. `render_feature()` now runs the display substring through `\esc_html()` before embedding it in the anchor, so a crafted record `text` (e.g. containing `</a>`) can't break out of the link or inject markup, independent of the downstream `wp_kses_post()`. Covered by a new `test_apply_escapes_anchor_display_text` regression test.

includes/transformer/class-facet.php
```diff
@@ -76,6 +76,103 @@ public static function for_urls( string $text, array $urls ): array {
...
+
+			case 'app.bsky.richtext.facet#mention':
+				$did = $feature['did'] ?? '';
+				if ( '' === $did ) {
+					return $display;
+				}
+				return '<a href="' . \esc_url( 'https://bsky.app/profile/' . $did ) . '">' . $display . '</a>';
```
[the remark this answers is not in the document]
jeherve: Good catch — fixed in 41e8cdf. `render_feature()` now runs the display substring through `\esc_html()` before embedding it in the anchor, so a crafted record `text` (e.g. containing `</a>`) can't break out of the link or inject markup, independent of the downstream `wp_kses_post()`. Covered by a new `test_apply_escapes_anchor_display_text` regression test.

includes/transformer/class-facet.php
```diff
@@ -76,6 +76,103 @@ public static function for_urls( string $text, array $urls ): array {
...
+
+			case 'app.bsky.richtext.facet#tag':
+				$tag = $feature['tag'] ?? '';
+				if ( '' === $tag ) {
+					return $display;
+				}
+				return '<a href="' . \esc_url( 'https://bsky.app/hashtag/' . \rawurlencode( $tag ) ) . '">' . $display . '</a>';
```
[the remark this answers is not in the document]
jeherve: Good catch — fixed in 41e8cdf. `render_feature()` now runs the display substring through `\esc_html()` before embedding it in the anchor, so a crafted record `text` (e.g. containing `</a>`) can't break out of the link or inject markup, independent of the downstream `wp_kses_post()`. Covered by a new `test_apply_escapes_anchor_display_text` regression test.

pfefferle (commented):
Reviewed for security, robustness, correctness, and tests. The XSS surface is solid — `esc_html()` on display text before tag assembly, `esc_url()` on every generated href, and `wp_kses_post()` as a second gate — and the byte-splice logic is correct.

Two blocking issues: `TypeError` fatals when a remote record carries `facets` (or a facet's `features`) as a present-but-non-array value — see inline comments. Both crash the cron sync silently for the affected notification, and both are one-line `is_array()` guards.

Non-blocking:
- Low-severity UX: a scheme `esc_url()` rejects yields `<a href="">display</a>` (links to the current page) rather than falling back to bare text.
- Mention links use the DID (`bsky.app/profile/{did}`) while the rest of `Reaction_Sync` uses the handle — correct here since the facet only carries the DID, but worth an inline comment.
- `render_feature` reads only `features[0]`; fine in practice (Bluesky emits single-feature facets), worth a one-line note.
- Test gaps: hashtag `#tag` branch, adjacent facets (`start == cursor`), unknown `$type` fallback, and the two type-confusion fatals above.

includes/class-reaction-sync.php
```diff
@@ -481,6 +482,15 @@ private static function process_reply( array $notification ): int|false {
...
 		}
 
+		/*
+		 * Bluesky stores `text` as the display string — long URLs are
+		 * truncated to e.g. `bsky.app/profile/jere...` and the real
+		 * target lives only in `facets`. Resolve those byte ranges back
+		 * into anchors before the text becomes comment content; otherwise
+		 * imported replies keep the lossy, unclickable display string.
+		 */
+		$text = Facet::apply( $text, $record['facets'] ?? array() );
```
pfefferle: **Fatal on malformed PDS data.** `?? array()` only catches absent/null. If the record carries `"facets"` as a string/scalar, it's passed to `Facet::apply( string, array $facets )` → `TypeError`, which kills the cron sync for this notification (WP-Cron swallows it). The PDS response is untrusted JSON. Guard with `\is_array()` here, or as the first line of `apply()`:

```php
if ( ! \is_array( $facets ) ) {
	return $text;
}
```

includes/transformer/class-facet.php
```diff
@@ -76,6 +76,111 @@ public static function for_urls( string $text, array $urls ): array {
...
+			 * that overlap a facet we've already consumed. Dropping a bad
+			 * range leaves its display text in place rather than corrupting
+			 * the surrounding bytes.
+			 */
+			if ( ! \is_int( $start ) || ! \is_int( $end ) || $start < $cursor || $start >= $end || $end > $length ) {
+				continue;
+			}
+
+			$result .= \substr( $text, $cursor, $start - $cursor );
+			$result .= self::render_feature( $facet['features'][0] ?? array(), \substr( $text, $start, $end - $start ) );
```
pfefferle: **Fatal on malformed PDS data.** If a facet's `features` is a string rather than an array, `$facet['features'][0]` returns a one-char string (not null), so `?? array()` doesn't fire and `render_feature( array $feature )` throws `TypeError`. (The sibling `$facet['index']['byteStart'] ?? null` is safe — it returns null cleanly.) Suggest:

```php
$features = $facet['features'] ?? array();
$feature  = \is_array( $features ) ? ( $features[0] ?? array() ) : array();
```

includes/class-reaction-sync.php
jeherve: Fixed in 5342213. The call site now guards the type before the typed `apply()` call:
```php
$facets = $record['facets'] ?? array();
$text   = Facet::apply( $text, \is_array( $facets ) ? $facets : array() );
```
Covered by `test_process_reply_tolerates_malformed_facets` (record with `facets` as a string still imports as a plain comment, no fatal).

includes/transformer/class-facet.php
jeherve: Fixed in 5342213. `apply()` now defends every nested shape from untrusted JSON, not just `features`: non-array facet entries are dropped via `array_filter`, a non-array `index` is skipped, and `features` is checked before `[0]`:
```php
$features = $facet['features'] ?? null;
$feature  = \is_array( $features ) && \is_array( $features[0] ?? null ) ? $features[0] : array();
```
The sort key also reads through a null-safe `byte_start()` helper so a malformed `index` can't fatal `usort`. New tests: `test_apply_tolerates_non_array_features`, `test_apply_tolerates_non_array_facet_entry`, `test_apply_tolerates_non_array_index`.

jeherve:
Thanks for the thorough review @pfefferle — all points addressed in 5342213.

**Blocking (both fixed, replies inline):**
- `facets` non-array at the call site → `is_array()` guard before `apply()`.
- `features` (and, while I was at it, facet entries and `index`) non-array inside `apply()` → guarded against the array typehints; `usort` now sorts via a null-safe `byte_start()` helper.

**Non-blocking:**
- **esc_url-empty UX** — fixed. `render_feature()` now falls back to the bare (escaped) display text whenever there's no usable target — unknown type, missing value, or a scheme `esc_url()` rejected — instead of emitting `href=""`. Covered by `test_apply_unsafe_scheme_falls_back_to_text`.
- **Mention DID vs handle** — added an inline comment noting the facet only carries the DID and that `bsky.app/profile/{did}` resolves the same as the handle form.
- **`features[0]` only** — added an inline comment explaining Bluesky emits single-feature facets.
- **Test gaps** — all filled: hashtag `#tag` branch, adjacent facets (`start == cursor`), unknown `$type` fallback, and the two type-confusion fatals.

Full suite green (668 tests), PHPCS clean.

pfefferle (approved):

PR #138, opened 2026-06-15
Decouple post + document deletes from the comment-reply cascade

### LINKED ISSUE #48: Chunk and decouple post-delete applyWrites batch
`Publisher::delete_post()` (trash path) and `delete_post_by_tids()` (permanent-delete path, since #32) submit one atomic `applyWrites` batch containing the post, the document, and one delete per outbound comment reply. The batch size scales with thread length, has no chunking, and no fallback. On a popular post:

- PDS write-count or payload limits may be exceeded.
- Atomic failure under load means no partial-progress recovery.
- Root cleanup is blocked by a single oversized comment cascade.

Matters more on permanent-delete than trash, since trash can be retried by re-trashing.

### Suggested shape

- Chunk comment deletions to a bounded batch size (e.g., 25–50 writes) and submit sequentially.
- Decouple post + document deletes from the comment-reply deletes so root cleanup is not blocked by an oversized comment cascade.
- If the target PDS documents a write-count limit, enforce it and add a test for the overflow path.

### Origin

Codex adversarial review; deferred from #32.

pfefferle:
Fixes #48

## Proposed changes:

`Publisher::delete_post()` (trash) and `delete_post_by_tids()` (permanent delete) used to bundle the post, the `site.standard.document`, and one delete per outbound comment reply into a single `applyWrites` batch. On a popular post that batch scales with thread length, and an atomic failure (or a write-count overflow) in the comment cascade would block cleanup of the post itself — leaving the root records orphaned on Bluesky while the post looks deleted locally.

This PR decouples the two:

* The post + document deletes and the comment-reply deletes are now submitted as **two independent, individually-chunked** `applyWrites` batches via a new `delete_in_decoupled_batches()` helper. Chunking (introduced in #32) is unchanged; it now runs per-batch.
* The **root batch goes first**. On `delete_post()`, the post/document meta is cleared as soon as that batch succeeds — independent of the comment batch — so a failing or oversized reply cascade can no longer strand the post in a published-looking state. If the comment batch then fails, only the reply meta is kept, so a re-trash retries just the replies.
* If the root batch itself fails, nothing is removed remotely, all meta is left intact, and the comment batch is skipped so a retry re-runs the whole cascade cleanly.

This addresses the "decouple" half of #48; the chunking half already landed in #32.

### Other information:

## Testing instructions:

* Reply to that post from Bluesky and run the reaction sync (`wp atmosphere sync-reactions`) so the post has at least one synced outbound comment reply.
* Trash the post. Confirm the post, its document, and the reply records are all removed from Bluesky.
* To exercise the decoupling guarantee directly, run the suite:
  ```
  npm run env-test -- --filter=delete
  ```
  The new tests assert that a failing comment batch still clears the post/document meta (`test_delete_post_clears_root_meta_when_comment_batch_fails`), that a failing root batch leaves everything intact and skips the comment batch (`test_delete_post_root_batch_failure_leaves_all_meta_intact`), and that root and comment deletes are always separate batches.

### Changelog entry

A changelog entry is already committed on this branch (`.github/changelog/fix-decouple-post-delete`), so no auto-entry is needed.

### DIFF
--- a/.github/changelog/fix-decouple-post-delete
+++ b/.github/changelog/fix-decouple-post-delete
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Deleting a post now reliably removes it from Bluesky even when it has a large number of replies to clean up.
--- a/includes/class-publisher.php
+++ b/includes/class-publisher.php
@@ -1124,9 +1124,11 @@ class Publisher {
 	 * cascade semantics, so they have to be enumerated alongside the
 	 * post records or they orphan on Bluesky.
 	 *
-	 * Writes are chunked into bounded `applyWrites` calls (the lexicon
-	 * caps a single batch at 200), so a high-traffic post with a long
-	 * reply tail still cleans up cleanly.
+	 * The post + document deletes and the outbound comment-reply deletes
+	 * are submitted as two independent, individually-chunked `applyWrites`
+	 * batches (the lexicon caps a single batch at 200). The root batch
+	 * goes first: a long reply tail can neither inflate the root batch
+	 * past the cap nor, when it fails, block cleanup of the post itself.
 	 *
 	 * @param \WP_Post $post WordPress post.
 	 * @return array|\WP_Error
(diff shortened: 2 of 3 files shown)


jeherve (approved):
This looks good to me. 🚢

PR #139, opened 2026-06-15
Block editor integration: pre-publish panel + per-post share toggle

### LINKED ISSUE #50: Add pre-publish federation sidebar panel (parity with bundled ActivityPub's)
## Problem

When a user is in the WordPress block editor about to publish a post, there's no visibility into what Atmosphere is going to do with it. Will it publish to Bluesky? As a link card, a short-form note, or a teaser-thread? Will it hit the 300-grapheme limit? The user finds out by publishing and looking at the result, which is poor UX for any post that isn't a quick test.

The bundled ActivityPub plugin has solved the equivalent problem for Mastodon: `bundled/activitypub/build/pre-publish-panel/plugin.js` registers a `PluginPrePublishPanel` that surfaces post-format suggestions ("change format to Status for short posts so they render as Notes on Mastodon"), media-type recommendations (Gallery / Video / Image / Audio), and federation-readiness info before publish.

Atmosphere has no equivalent today. Adding one closes the symmetry gap and gives Atmosphere users the same pre-publish confidence Mastodon-bound posts already get.

## Proposal

A new pre-publish federation sidebar panel in Atmosphere that surfaces, before the user clicks Publish:

- **Will this publish to Bluesky?** Based on auto-publish setting + post format + connection state. If no, a one-line reason (e.g. "Bluesky disconnected" or "auto-publish off").
- **What strategy will run?** `link-card-fallback` / `short-form-note` / `long-form-teaser-thread`. Tied to the existing strategy projector (`atmosphere_long_form_composition` filter etc.).
- **Bluesky character count** for the post body, with the 300-grapheme limit and a clear over-limit indicator. For teaser-thread strategy, show per-post counts.
- **Optionally**: a small preview of how the post will appear — link card thumbnail + title for the link-card path, first thread post for teaser-thread, etc.

## Why this is independent of any one consumer

This is pure Atmosphere UX. Any Atmosphere user — standalone install, FOSSE-bundled, wp.com Simple — benefits. It is not a FOSSE-driven request; FOSSE just noticed the gap while spec'ing its own posting UI and decided not to build a parallel surface.

## Implementation reference

Look at `wordpress-activitypub`'s `bundled/activitypub/build/pre-publish-panel/plugin.js` for the shape. Key pieces:

- `registerPlugin('activitypub-pre-publish', { render: ... })`
- `PluginPrePublishPanel` from `@wordpress/editor`
- `useSelect` on the editor store to pull blocks + post format
- Per-block analysis (image count, video count, text length) to recommend a post format

Atmosphere's equivalent would pull from its own publish-strategy logic and connection state instead.

## Out of scope for this proposal

- Outbound-post-as-reply support (separate concern, see Atmosphere's lack of `?in_reply_to=<bsky-url>` handling).
- The composer / quick-note UX itself (that's a downstream consumer concern; this proposal is just about the sidebar panel in the standard editor).

## Tracking

FOSSE-side tracking issue (so the FOSSE team is notified if/when this work moves): linked from FOSSE's posting-ui SDD at `sdd/posting-ui/spec.md` and as a Linear issue under the FOSSE project. We may end up contributing the work upstream ourselves if the use case sharpens.

pfefferle:
Fixes #50

## Proposed changes:

Adds a block-editor **pre-publish federation panel** (parity with the bundled ActivityPub plugin's). Before the author clicks Publish, the panel shows:

* **Will this be shared to Bluesky?** — reflecting the *intended* editor visibility (connection state, auto-publish setting, supported post type, and crucially **private / password-protected** posts are shown as *not* shared), with a plain-language reason when it won't.
* **What strategy will run** — Short note / Link card / Text with link / Teaser thread.
* **Bluesky character count** vs the 300-grapheme limit, with an over-limit warning (per-post for teaser threads).

### Architecture (mirrors `wordpress-activitypub`)

* `Atmosphere\Transformer\Post::project()` — a side-effect-free *projection mode* that reuses the real `transform()` / `build_long_form_records()` logic so the panel never drifts from publish. It skips blob uploads and the DNS-resolving facet extraction (no network, no meta writes).
* `Atmosphere\REST\Admin\Pre_Publish_Controller` (`extends WP_REST_Controller`) — admin endpoint on the **`atmosphere/1.0`** namespace (`admin/pre-publish-preview`, `show_in_index => false`, `edit_post` permission). Kept separate from the public `atmosphere/v1` OAuth namespace. Projects the *unsaved* editor content.
* `Atmosphere\Block_Editor` — enqueues the built panel script only (no REST logic).
* `src/pre-publish-panel/{block.json,plugin.js,utils.js,__tests__}` → built to `build/pre-publish-panel/` via `@wordpress/scripts`.

This adds the `@wordpress/scripts` build pipeline (the plugin had no editor JS before).

### Other information:

- [x] Have you written new tests for your changes, if applicable? (PHP: projector + REST controller; JS: `utils` unit tests.)

## Testing instructions:

* Connect the plugin to a Bluesky account with auto-publish on.
* Start a new post and open the **Bluesky** panel in the pre-publish sidebar (click Publish). Confirm it shows the strategy and a live character count as you type; type past 300 chars and confirm the over-limit warning.
* Set the post to **Private**, or add a **password** → the panel should say the post won't be shared, with the reason.
* Switch to an unsupported post type / turn off auto-publish / disconnect → each shows its own reason.
* Run the suites:
  ```
  npm run env-test -- --filter='test_project|Test_Pre_Publish_Controller'
  npm run test:unit:js
  ```

### Follow-ups (out of scope)

* Optional visual preview (link-card thumbnail / first thread post) — deferred per the issue.
* A per-user/content-hash cache or rate limit on the projection endpoint (the panel debounces client-side at 400ms; same exposure as the publish path, auth-gated to `edit_post`).

### Changelog entry

A changelog entry is committed on this branch (`.github/changelog/add-pre-publish-panel`).

---

## Update: per-post share toggle + sidebar migration

Following review feedback, this PR now also **replaces the legacy PHP meta box** with a JS document-sidebar panel and adds a **per-post share control**:

* **"Share this post to Bluesky" toggle** in a new **Bluesky** document panel (`src/editor-plugin/`, mirroring ActivityPub's `editor-plugin`), bound to a REST-exposed `atmosphere_disabled` post meta.
* The toggle folds into `is_post_publishable()`, so switching sharing **off after a post was shared removes it from Bluesky** via the existing cleanup path (verified by `test_disabling_share_on_synced_post_schedules_delete`).
* A **"View on Bluesky"** link appears once the post is shared (read-only `atmosphere_url` REST field).
* The **pre-publish panel respects the toggle** (sends the unsaved `disabled` state) and correctly reflects **private / password-protected** visibility.
* Old `add_meta_box` / `templates/meta-box.php` removed.

**Verification:** 701 PHP tests + 9 JS tests pass; PHPCS + ESLint clean.

### Testing (toggle)

* In the editor, open the **Bluesky** panel in the document sidebar; toggle "Share this post to Bluesky" off and publish → the post is not shared.
* Toggle it off on an already-shared post and update → the post is removed from Bluesky.


kraftbj (dismissed):
The latest two commits clear the Copilot review. I resolved the two threads that 56cb80e actually fixed (the projection HTTP block and the code-point count). Nice work single-sourcing the route and meta key through `config.js` too.

I went through the rest carefully and there's nothing that blocks merge. A few things worth a look first, roughly in priority order.

**The un-share path leans on `transition_post_status`, and I think that's a little fragile.** Walking it through: on a toggle-off save, REST writes `atmosphere_disabled` *after* `wp_update_post()` fires the transition, so `on_status_change` still sees the post as publishable and schedules `atmosphere_update_post` instead of a delete. In normal cron that self-heals — the update handler re-fetches and calls `delete_post()` once the meta is committed, and I confirmed all three handlers re-check `is_post_publishable()` at fire time, which is the right design. So the common path is fine. Two gaps, though:

- A meta-only save (writing `atmosphere_disabled` without a post update firing the transition) schedules nothing at all. Gutenberg's normal save does fire the transition, so the editor's covered, but we're one client-behavior change away from a silent miss.
- If a cron worker fires in the window after we schedule but before the meta write commits, it acts on the stale (still-enabled) meta and no delete ever gets scheduled — live record left on Bluesky.

Both close cleanly if we also react to the meta change itself (`updated_post_meta` / `added_post_meta` on the disabled key, or `rest_after_insert_post`) and re-run the cleanup decision once meta is persisted. Bonus: that's the hook the test should exercise. Right now `test_disabling_share_on_synced_post_schedules_delete` sets the meta *before* calling `on_status_change`, which is the reverse of the real REST order, so it's green without testing the case that actually matters.

**A failed un-share is silent.** If the delete cron fails (PDS down, etc.), we log under `WP_DEBUG` and stop — no retry, and the panel still shows "View on Bluesky." The cleanup marker sticks around so a later save retries, but the user gets no signal and no reason to re-save. Worth a delayed retry and/or surfacing the pending state in the panel.

Two smaller decisions, not blockers:

- `Atmosphere\REST\Admin` under `includes/rest/admin/` isn't in our documented structure (`docs/php-class-structure.md` points REST handlers at `includes/wp-admin/` → `Atmosphere\WP_Admin`), and the route namespace is `atmosphere/1.0` while the existing one is `atmosphere/v1`. Both work; I just want us to pick on purpose and update the docs so the next person isn't guessing.
- The panel only enqueues on supported post types, so the "won't be shared" reason the controller returns for unsupported types never actually shows. Copilot flagged this too (I left that thread open). Intended, or do we want the panel to explain itself there?

FWIW, two edge cases I'll just note and wouldn't hold the PR on: the projection HTTP block covers the WP HTTP API but not a shortcode doing raw cURL/sockets, and a filter on `atmosphere_transform_bsky_post` that appends text after `truncate_text()` could desync the count from what publishes. Both need a third party to do something unusual.

The tl;dr: solid, no blockers, the un-share self-heals on the normal path. I'd like the meta-change hook plus the failed-delete handling before we call it done; the namespace and post-type bits are quick calls.

pfefferle:
Thanks for the review — all three Copilot points addressed:

- **oEmbed/HTTP during projection** (`class-post.php`): the preview endpoint now blocks all outbound HTTP for the duration of the projection (`56cb80e`), so it can't fetch oEmbeds or be used as an egress amplifier. The `no-network` guarantee now actually holds.
- **Grapheme vs code-point mismatch**: the preview count now uses `mb_strlen` (code points) — the same unit the short-form publish path clamps on — so the panel never reports "within limit" for text that would be shortened. The spec-correct fix (make *publish* grapheme-aware, then switch the panel back to graphemes) is tracked in #140.
- **Panel on unsupported post types**: intended — we don't surface a Bluesky panel on a post type that isn't wired to Bluesky, so the controller's defensive `unsupported` reason simply isn't reachable from the UI. Corrected the misleading testing note.

Also addressed @kraftbj's review: the un-share path now reconciles off the committed meta change (not just `transition_post_status`), the editor panel surfaces a pending-removal notice, and the client-metadata route moved into a proper `Atmosphere\Rest` controller with the docs updated. Automatic retry of a failed un-share is tracked separately in #141.

kraftbj (approved):

PR #145, opened 2026-06-17
Treat overflowing titleless posts as long-form

pfefferle:
---

### Notes for integrators / reviewers

- **Filter default flip:** the `atmosphere_is_short_form_post` default argument now flips from `true` to `false` for a **text-only** titleless / post-format post whose rendered body exceeds 300 characters. Callbacks that observe and pass the default through will see the new value. Posts with in-body images are unaffected (they stay short-form). The filter can still force either outcome.
- **Record shape on republish:** an affected post previously published as short-form will, on its next republish/update, change from a plain-text record to a link-card (`app.bsky.embed.external`) record. Under a non-default `teaser-thread` composition this can route through `rewrite_thread()`; see the inline thread on the republish hazard (tracked as a follow-up).
- **Scope narrowed after review:** per @kraftbj's review, conversion is limited to text-only overflow so a photo post with a long caption keeps its native image gallery. Deferred follow-ups (memoization of the verdict, grapheme-aware cap, fallback card title + rendered teaser, republish-transition guard) are noted in the inline threads.


includes/transformer/class-post.php
```diff
@@ -304,30 +304,13 @@ public function set_document_strong_ref( array $ref ): void {
...
-		if ( ! $redacted ) {
-			$is_short = \wp_validate_boolean(
-				\apply_filters(
-					'atmosphere_is_short_form_post',
-					$this->is_short_form( $this->object ),
-					$this->object
-				)
-			);
-		}
+		$is_short = $redacted ? true : $this->is_short_form_post();
```
kraftbj, line 328: **In-body images are dropped when an overflowing post is reclassified to long-form.** (Codex, new finding; verified.)

The short-form branch calls `build_images_embed()`, which parses up to 4 in-body `core/image` blocks and falls back to the featured image. The long-form branch reached from here calls `build_embed()`, whose only media is `get_post_thumbnail_id()` (the featured thumbnail). So a titleless / `aside` post with in-body images plus a body over 300 chars changes from a native multi-image Bluesky post into a single link card carrying at most the featured thumbnail (and, per the gate comment, often a blank card title).

Worth a conscious decision: is losing the in-body images acceptable for these posts, or should the long-form path preserve them?

includes/transformer/class-post.php
```diff
@@ -1316,22 +1299,31 @@ private function apply_post_embed_filter( ?array $embed, string $strategy ): ?ar
...
 	 */
 	private function is_short_form( \WP_Post $post ): bool {
-		if ( ! \post_type_supports( $post->post_type, 'title' ) || empty( $post->post_title ) ) {
-			return true;
+		if ( \post_type_supports( $post->post_type, 'title' ) && ! empty( $post->post_title ) && ! \get_post_format( $post ) ) {
+			return false;
 		}
 
-		return (bool) \get_post_format( $post );
+		return \mb_strlen( $this->render_post_content_plain( $post ) ) <= self::BLUESKY_MAX_GRAPHEMES;
```
kraftbj: **What the long-form fallback actually ships for a titleless post has two problems.** (Both reviewers.)

1. `build_embed()` sets `external.title` to `sanitize_text( get_the_title() )`, which is empty for a titleless post, so the link card has a **blank title**. `build_text()` then composes `excerpt + permalink` with the empty title filtered out. The PR's headline ("a summary with a link card") is generous: it is a blank-title card with an auto-excerpt.
2. This gate measures the **rendered** body (`render_post_content_plain()` runs `the_content`), but `get_excerpt()` derives the teaser from **raw `post_content`**. For a post that only overflows after block or shortcode rendering, the teaser can omit the very content that triggered the overflow.

Consider deriving a fallback card title and building the teaser from the same rendered string the gate measured, so the gate and the composition agree.

includes/transformer/class-base.php
```diff
@@ -22,6 +22,19 @@
...
+	 *
+	 * The `app.bsky.feed.post` lexicon caps `text` at 300 graphemes. The
+	 * transformers approximate that with `mb_strlen` (code points) — every
+	 * grapheme is at least one code point, so a code-point cap never exceeds
+	 * the grapheme limit. Shared by the post and comment transformers, which
+	 * both emit Bluesky records bounded by this cap.
+	 *
+	 * @var int
+	 */
+	public const BLUESKY_MAX_GRAPHEMES = 300;
```
kraftbj, line 36: **Constant name vs measurement mismatch.** (Both reviewers; they disagreed on severity.)

The constant is `BLUESKY_MAX_GRAPHEMES` but every gate measures `mb_strlen()` (code points). The docblock acknowledges this and argues it is conservative, and that holds: a code-point count never exceeds the grapheme count, so nothing ships over the lexicon cap. The one real effect: a roughly 300-grapheme body built from emoji / ZWJ / combining-mark sequences counts well above 300 code points and gets needlessly routed to a link card when it would have fit as a native post. The repo already has a grapheme-aware truncation helper in `functions.php`.

Codex rated this High; I would call it P2/P3 since the error never produces an invalid record. Renaming the constant (or using the grapheme helper) removes the trap either way.

includes/transformer/class-post.php
```diff
@@ -1316,22 +1299,31 @@ private function apply_post_embed_filter( ?array $embed, string $strategy ): ?ar
...
+	 * with `mb_strlen` to match `build_short_form_text()`'s own
+	 * `truncate_text()` cap, so the gate and the truncation it avoids agree.
+	 *
 	 * @param \WP_Post $post Post being transformed.
 	 * @return bool
 	 */
 	private function is_short_form( \WP_Post $post ): bool {
-		if ( ! \post_type_supports( $post->post_type, 'title' ) || empty( $post->post_title ) ) {
-			return true;
+		if ( \post_type_supports( $post->post_type, 'title' ) && ! empty( $post->post_title ) && ! \get_post_format( $post ) ) {
```
kraftbj, line 1353: **Republish hazard for already-published posts.** (Both reviewers, independently.)

This reclassification means a post previously published as short-form (one record) can now resolve to long-form. Under the `teaser-thread` long-form composition, an edit that crosses the 300 boundary turns a stored 1-record post into a 2-3 record thread. `Publisher::update_post()` picks the strategy from this discriminator; the stored-count vs new-count mismatch falls through to `rewrite_thread()`, which deletes the old record(s) before republishing and (per its own docblock) orphans existing likes, reposts, and replies.

Narrow (needs the teaser-thread composition plus an edit crossing the boundary) but destructive when hit, and there is no update-path test for it.

includes/transformer/class-post.php
```diff
@@ -1371,6 +1363,23 @@ public function is_short_form_post(): bool {
 			return true;
 		}
 
+		/**
+		 * Filters whether the post should be treated as short-form for Bluesky.
```
kraftbj, line 1408: **Filter contract change, plus a split-brain edge.** (Codex + internal api-contract reviewer.)

- **Contract change:** the `atmosphere_is_short_form_post` default argument now flips from `true` to `false` for a titleless / post-format post whose body exceeds 300 chars. Third-party callbacks that observe the default and pass it through will silently see a different value, and the emitted record shape changes. The in-source docblock is updated, but the changelog is not (see the changelog comment).
- **Split-brain (Codex, High):** `is_short_form_post()` is evaluated more than once per publish (Publisher precompute and routing, plus `transform()`), and the filter is re-applied each call with no memoization. A *stateful* filter returning `true` then `false` can skip the document strongRef precompute yet still publish a long-form external card, so the card ships without its document `associatedRef`. Requires a misbehaving stateful filter (uncommon), but memoizing the verdict on the instance closes it cheaply.

.github/changelog/fix-short-form-overflow
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Long posts without a title are now shared to Bluesky as a summary with a link back to the original, instead of being cut off mid-sentence with no way to reach the full post.
```
kraftbj, line 4: **The changelog under-discloses the behavior change.** (Internal api-contract reviewer.)

This entry is purely end-user framed, but the PR also (a) changes the `atmosphere_is_short_form_post` filter's default value for integrators and (b) changes the AT-record shape of existing titleless long posts (plain text becomes an external embed card) on their next republish. Consider adding a one-line note for integrators, and reconsider `Significance: patch` vs `minor` given the record-shape change on republish.

tests/phpunit/tests/transformer/class-test-post.php
```diff
@@ -168,11 +168,15 @@ public function test_titled_post_with_aside_format_is_short_form() {
...
+	 * A titleless post whose body overflows the 300-char native cap is not
+	 * really short-form: it falls back to the long-form link-card so the
+	 * reader gets a teaser plus a route back to the original, instead of a
+	 * sentence fragment with no link home.
 	 *
 	 * @covers ::transform
+	 * @covers ::is_short_form_post
 	 */
-	public function test_short_form_truncates_over_cap() {
+	public function test_long_titleless_post_falls_back_to_link_card() {
```
kraftbj, line 179: **Test coverage gaps for the new gate.** (Internal testing reviewer; Codex concurs.)

- No boundary test at exactly 300 (stays short) and 301 (becomes long-form). The existing tests use a body of roughly 1400 chars, far from the `<=` edge.
- Replacing `test_short_form_truncates_over_cap` dropped the only coverage of `build_short_form_text()`'s defensive truncation, which is still reachable when the filter forces short-form on an overflowing body. Add a filter-forces-short test asserting `mb_strlen( text ) <= 300`.
- No update-path / `teaser-thread` test for the 1-to-N record reclassification (see the `is_short_form()` comment).
- `test_long_format_post_falls_back_to_link_card` only asserts the embed `$type`; mirror the titleless test's permalink and URI assertions.

kraftbj (commented):
## Automated review summary (AI-generated)

Heads up: this is an AI-generated review, a Claude multi-persona `/ce-code-review` pass plus an independent Codex adversarial second opinion. Two models reviewed separately and converged. Treat it as input, not a verdict.

**Verdict: ready with fixes.** The core logic is correct. The inverted `is_short_form()` guard is the exact complement of the old categorical check, the `<= 300` gate measures the same memoized string that `build_short_form_text()` truncates (no off-by-one), and redacted handling is preserved. No crash or data-loss bugs. The findings are about how *complete* the fix is for its own target case, plus disclosure and test coverage.

Priority order:
1. **In-body images dropped on overflow** (Codex, new): an overflowing titleless / `aside` post with in-body images degrades to a single link card. Strongest concrete regression.
2. **Blank card title + raw-vs-rendered teaser**: the "summary with a link card" ships a blank-title card whose teaser can omit the rendered content.
3. **Changelog disclosure**: filter-default flip and record-shape-on-republish; reconsider patch vs minor.
4. **Test gaps**: 300/301 boundary, lost truncation coverage, update-path test.
5. Lower priority: grapheme-cap naming, memoize the verdict, the `280` literal.

**Cross-model note:** 4 of 5 Codex findings overlapped the internal review. The non-overlap (image-embed loss) was a real gap. The models disagreed on the grapheme finding's severity (Codex High, Claude P2/P3, since that error is always in the safe direction).

Inline comments below. No files were changed; this is review-only.

includes/transformer/class-post.php
```diff
@@ -186,30 +186,13 @@ public function set_document_strong_ref( array $ref ): void {
...
-					'atmosphere_is_short_form_post',
-					$this->is_short_form( $this->object ),
-					$this->object
-				)
-			);
-		}
+		$is_short = $redacted ? true : $this->is_short_form_post();
 
 		$text  = $redacted ? '' : ( $is_short ? $this->build_short_form_text() : '' );
 		$embed = null;
```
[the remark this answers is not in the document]
pfefferle: Done in 5585596 — `$is_short` is flipped to `false` in the empty-body link-card fallback, so the embed strategy label and the `atmosphere_transform_bsky_post` context report `link-card`, not `short-form`.

includes/transformer/class-post.php
pfefferle: Good catch — fixed by narrowing the conversion in 5585596: an overflowing categorically-short post now **stays short-form when it has in-body images**, so the native `app.bsky.embed.images` gallery is preserved. Only the text-only (link-blog) case converts to long-form. Added `test_overflowing_post_with_inbody_images_stays_short_form`.

includes/transformer/class-post.php
pfefferle: Acknowledged. The blank card title and the raw-vs-rendered teaser are pre-existing `build_embed()` / `build_text()` behaviors that also affect explicitly-titleless long-form posts, so I've left them out of this narrowed fix to keep it focused. Tracking as a follow-up: a fallback card title and a teaser built from the same rendered string the gate measures.

includes/transformer/class-base.php
pfefferle: Left as-is for this PR. As the docblock notes, `mb_strlen` only errs in the safe direction (a code-point count never exceeds the grapheme count, so nothing ships over the lexicon cap); the one effect is a dense emoji/ZWJ body near 300 graphemes routing to a link card a little early. Happy to switch the gate to a grapheme count in a follow-up if you'd prefer.

includes/transformer/class-post.php
pfefferle: Acknowledged — narrowed but not eliminated: a text-only titleless post edited past 300 can still reclassify short→long, and under the `teaser-thread` composition `update_post()` routes through `rewrite_thread()`, which orphans existing engagement. Not guarded in this PR; flagging a follow-up to detect the transition and avoid the destructive rewrite. The default `link-card` composition reuses the rkey and is unaffected.

includes/transformer/class-post.php
pfefferle: The filter-default flip is real — I'll document it for integrators in the PR description. The split-brain (a stateful filter returning `true` then `false` across the multiple `is_short_form_post()` calls in a publish) needs a misbehaving filter; memoizing the verdict on the instance is a cheap follow-up I can add here if you want it in this PR.

.github/changelog/fix-short-form-overflow
pfefferle: Kept the changelog end-user-only per AGENTS.md (no integrator jargon in the update-screen text); the filter-default flip and the record-shape-on-republish note go in the PR description instead. On significance: it's fundamentally a bug fix, and the narrowing shrank the behavior change to text-only posts, so I've kept it `patch` — happy to bump to `minor` if you'd rather flag the record-shape change explicitly.

tests/phpunit/tests/transformer/class-test-post.php
pfefferle: Added in 5585596: a 300/301 boundary test (`test_short_form_overflow_boundary`), restored the defensive-truncation coverage via a filter-forced-short test (`test_forced_short_form_over_limit_is_clamped`), and the in-body-image regression test. The update-path / `teaser-thread` reclassification test ties to the republish-hazard follow-up above.

includes/transformer/class-post.php
kraftbj: Resolved in 5585596: an overflowing post with in-body images now stays short-form, so its native `app.bsky.embed.images` gallery is preserved; only text-only overflow converts to long-form. Covered by `test_overflowing_post_with_inbody_images_stays_short_form`.

includes/transformer/class-base.php
kraftbj: Resolving as a documented, conscious tradeoff: `mb_strlen` only errs in the safe direction (a code-point count never exceeds the grapheme count, so nothing ships over the lexicon cap), and the docblock says so. A grapheme-aware gate is noted as an optional follow-up.

includes/transformer/class-post.php
kraftbj: Resolved: the filter-default flip is documented for integrators in the PR description, and the split-brain is closed in d774e8c by memoizing the verdict on the transformer instance so the multiple `is_short_form_post()` calls per publish can't disagree (`test_short_form_verdict_is_memoized`).

.github/changelog/fix-short-form-overflow
kraftbj: Resolved: the changelog stays end-user-only per AGENTS.md, with the integrator-facing filter-flip and record-shape-on-republish notes in the PR description instead. Significance left at patch (bug fix, narrowed to text-only overflow).

tests/phpunit/tests/transformer/class-test-post.php
kraftbj: Resolved in 5585596: added the 300/301 boundary test, restored the defensive-truncation coverage (`test_forced_short_form_over_limit_is_clamped`), and the in-body-image regression test. The update-path / teaser-thread test stays with the open republish-hazard thread.

kraftbj (approved):

PR #149, opened 2026-06-19
Harden pre-publish projection and client-metadata error handling

pfefferle:
## Proposed changes:

Follow-up to #148, addressing @kraftbj/@jeherve review notes that are about the **pre-publish panel / client-metadata** rather than the reactions block, kept out of that PR to avoid scope creep.

* **Pre-publish controller** (`get_preview`): `Post::project()` runs `the_content` over raw, unsaved editor markup, so a malformed block or a misbehaving content/shortcode filter could throw and fatal the keystroke-driven endpoint into an opaque 500. It's now wrapped in `try/catch(\Throwable)/finally` — on a throw it logs the post ID via `debug_log()` and returns a structured `WP_Error` (500); the `finally` still removes the HTTP-blocking filter on every path.
* **Pre-publish panel** (`plugin.js`): the fetch `.catch` previously collapsed every failure into one generic message and threw the error away. It now logs the error and shows a **distinct message for permission failures** (401/403 or `rest_forbidden`) vs transient ones.
* **Client-metadata controller**: an invalid `atmosphere_client_metadata` filter return was only surfaced via `_doing_it_wrong()`, which is silent in production. It now *also* routes through `debug_log()`, so operators can opt into the signal via the `atmosphere_debug_log` filter without enabling `WP_DEBUG` site-wide.

### Other information:

## Testing instructions:

* `composer lint && npm run lint:js && npm run build`
* `npm run env-test` — 720 tests pass, incl. the new `test_preview_returns_error_when_projection_throws` (a throwing `the_content` filter → `WP_Error` 500, not a fatal).

Manual:

1. In the editor, the Bluesky pre-publish panel still previews normally.
2. If the projection can't be generated, the panel shows a clear message (and a distinct one for permission errors) instead of a bare "couldn't load".

### Changelog entry

Entry included at `.github/changelog/fix-pre-publish-error-handling` (Patch / Fixed).


jeherve (commented):
Nice work on this.

One thing Claude and I would flag before merge: the `isAuth` logic isn't covered. This is the code behind the nonce-vs-permission fix, and it's the bit most likely to quietly regress someday. There's already a render precedent in `src/settings-connect/__tests__/`, so I'm wondering if we could pull the decision out into a small helper (like `strategyLabel`) and table-test the matrix. Would make the "expired nonce shouldn't read as denied" guarantee explicit.

That's not a blocker though.

pfefferle:
Done in 20fd87a. I pulled the decision into `isAuthError()` in `utils.js`, next to `strategyLabel`, and table-tested the matrix: `rest_forbidden` and a bare 401/403 read as denied, `rest_cookie_invalid_nonce` stays transient. No behavior change, it just makes the nonce case explicit so it can't regress quietly. Thanks @jeherve!

jeherve (approved):
This should be good now. 🚢

PR #152, opened 2026-06-22
Add: Custom Bluesky text for individual posts

### LINKED ISSUE #151: Feature Request: Customize the text sent to Bluesky, independently of the post content
Reported on the WordPress.org support forum: [Feature request: connect manually posted Bluesky posts to blog entry](https://wordpress.org/support/topic/feature-request-connect-manualy-posted-bluesky-posts-to-blog-entry/)

### What

Add a per-post field in the post settings — similar to the native **Excerpt** field — that lets the author write custom text to be sent to Bluesky, instead of always cross-posting the post content/title verbatim. When the field is empty, fall back to the current behavior.

### Why

People who publish across several social networks tailor their message per network: different hashtags, different @-mentions, a punchier hook, etc. Today ATmosphere cross-posts the post content automatically with no way to adjust what Bluesky receives, so the Bluesky post cannot be customized to fit the network's conventions.

### How

A textarea in the ATmosphere sidebar panel (mirroring the Excerpt UX) whose value is used as the short-form Bluesky post body when set. The original poster also mentioned wanting to compose the Bluesky post separately and pull its replies in via WordPress comments — the comment side is already covered by reaction/comment sync, so this issue focuses on the custom-text field.

pfefferle:
Fixes #151.

## Proposed changes:

Adds a per-post **Custom Bluesky text** field so authors can tailor what goes to Bluesky — different hashtags, mentions, or a punchier hook — without changing the blog post itself. This is the most-requested item from the linked WordPress.org thread.

**Why this design:** the site already has a *Long-form posts* composition setting (`link-card` / `truncate-link` / `teaser-thread`), so a free-text field risks confusing authors about how the two interact. The field is therefore a clean **override**: when set, that exact text is posted with a link card back to the post (so it still "connects to the blog entry"), and the composition style is skipped for that post. Empty = today's behavior, unchanged. The pre-publish panel labels it **"Custom text"** so the author can see the composition setting no longer applies.

What's included:
- New REST-exposed post meta `atmosphere_custom_text` (string, `edit_post`-gated, sanitized with `sanitize_textarea_field`, cleaned up on uninstall).
- Editor sidebar textarea under the existing share toggle; pre-publish preview projects the *unsaved* text and warns on the 300-grapheme limit.
- Transformer honors the override across publish (short- and long-form paths) and projection; preserves author line breaks; still extracts `@mention` / `#hashtag` / link facets.
- Editing the text on an already-shared post reconciles the Bluesky record.
- Redacted (private / password-protected) posts never emit the custom text.
- `atmosphere_transform_bsky_post` gains an additive `is_custom_text` context flag for hook authors (structural strategy stays `link-card`).

### Other information:

New PHPUnit coverage in `class-test-post-custom-text.php` (override across paths, truncation, line breaks, redaction, projection strategy/length, filter-context flag), a controller test for unsaved-text projection, an `on_share_meta_changed` reconcile test, and a JS strategy-label test.

## Testing instructions:

1. Connect the site to Bluesky and ensure auto-publishing is on.
2. Create a new post with a title and body. In the editor's **Bluesky** sidebar panel, type something into **Custom Bluesky text** (e.g. `My own take 👋 #blogging`).
3. Open the pre-publish panel (Publish flow) — it should report the strategy as **Custom text** and show your text's character count against 300.
4. Publish. On Bluesky, confirm the post shows your custom text with a link card back to the WordPress post (not the auto-composed title/excerpt).
5. Edit the custom text on the published post and update — the Bluesky post should update to the new text.
6. Clear the field and update — it should revert to the normal *Long-form posts* composition.
7. Edge checks: set a site-wide *Teaser thread* style, confirm a custom-text post still publishes as a single post; password-protect a post with custom text and confirm nothing leaks.


includes/transformer/class-post.php
```diff
@@ -352,7 +443,21 @@ public function transform(): array {
 		$link_facets = array();
 
 		if ( ! $redacted ) {
-			if ( $is_short ) {
+			if ( $this->has_custom_text() ) {
```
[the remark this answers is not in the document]
pfefferle: Good catch — fixed in d8acd71. A custom-text post is always a single link-card record, so `transform()` now resolves `$is_short` directly for the redacted and custom-text cases and only calls `is_short_form_post()` on the normal path, skipping the body-length work and the `atmosphere_is_short_form_post` filter for custom-text posts.

src/editor-plugin/plugin.js
```diff
@@ -56,7 +62,7 @@ const EditorPlugin = () => {
 				label={ __( 'Share this post to Bluesky', 'atmosphere' ) }
 				checked={ enabled }
 				onChange={ ( value ) =>
-					setMeta( { ...meta, [ META_KEY ]: ! value } )
+					setMeta( { ...meta, [ DISABLED_META_KEY ]: ! value } )
 				}
```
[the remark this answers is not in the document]
pfefferle: Declining this one: **object** spread of `null`/`undefined` is a no-op, not an error — `{ ...undefined, [KEY]: v }` evaluates to `{ [KEY]: v }`. (Only *array* spread throws on null/undefined.) This is also the pre-existing share-toggle pattern, so there's no new risk here.

src/editor-plugin/plugin.js
```diff
@@ -71,6 +77,21 @@ const EditorPlugin = () => {
...
 			/>
 
+			{ enabled && (
+				<TextareaControl
+					__nextHasNoMarginBottom
+					label={ __( 'Custom Bluesky text', 'atmosphere' ) }
+					value={ customText }
+					onChange={ ( value ) =>
+						setMeta( { ...meta, [ CUSTOM_TEXT_META_KEY ]: value } )
+					}
```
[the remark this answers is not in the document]
pfefferle: Same as above — `{ ...meta, [CUSTOM_TEXT_META_KEY]: value }` is safe when `meta` is `undefined`/`null` because object spread of nullish values is a no-op. No change needed.

src/editor-plugin/plugin.js
```diff
@@ -71,6 +77,21 @@ const EditorPlugin = () => {
...
+					help={ __(
+						'Leave empty to use your normal sharing style. When set, this exact text is posted to Bluesky with a link back to this post.',
```
jeherve: Maybe it should say


```suggestion
						'Customize how this post is shared on Bluesky. Leave empty to use the default sharing message, or enter a custom message that will be shared with a link back to this post. You can mention other Bluesky users in your message.',
```

that's a bit long though.

src/editor-plugin/plugin.js
```diff
@@ -71,6 +77,21 @@ const EditorPlugin = () => {
 				}
 			/>
 
+			{ enabled && (
+				<TextareaControl
```
jeherve, line 116: Maybe we would need to remove the no margin rule for the component above, so this one has room to breathe?

jeherve (commented):
This is looking good. I have some questions / suggestions below.

includes/rest/admin/class-pre-publish-controller.php
```diff
@@ -227,7 +244,11 @@ public function get_preview( WP_REST_Request $request ) {
 		\add_filter( 'pre_http_request', $block_http, 0 );
 
 		try {
-			$projection = ( new Post( $draft ) )->project();
+			$transformer = new Post( $draft );
+			// Project against the *unsaved* custom text, not the saved meta,
+			// so the preview tracks the textarea as the author types.
+			$transformer->set_custom_text_override( (string) $request['customText'] );
+			$projection = $transformer->project();
```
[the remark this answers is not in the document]
pfefferle: Good catch — fixed in 43c7401. The projector now only overrides when the request actually carries the param (`$request->has_param( 'customText' )`), so an older/cached editor that omits it falls back to the saved meta instead of being forced to the default composition by a cast-from-missing empty string. `has_param()` distinguishes absent (false) from sent-empty (true), so a deliberately-cleared textarea still forces the default. Covered by `test_preview_without_custom_text_param_reads_saved_meta`.

includes/rest/admin/class-pre-publish-controller.php
```diff
@@ -106,33 +106,50 @@ public function register_routes(): void {
...
+						 * line breaks while stripping tags, matching the meta's
+						 * registered sanitizer.
+						 */
+						'customText' => array(
+							'description'       => \__( 'The unsaved custom Bluesky text (empty to use the default composition).', 'atmosphere' ),
+							'type'              => 'string',
+							'default'           => '',
+							// Bounds the per-keystroke projection work; the text
+							// is clamped to 300 graphemes when published anyway.
+							'maxLength'         => 2000,
```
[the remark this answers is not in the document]
pfefferle: Fixed in 43c7401 — reworded to "Bluesky's 300-character limit" so the comment no longer implies a grapheme-based clamp.

includes/transformer/class-post.php
```diff
@@ -331,6 +358,82 @@ public function set_document_strong_ref( array $ref ): void {
...
+	 */
+	private function custom_text_body(): string {
+		return \trim( \html_entity_decode( $this->get_custom_text(), ENT_QUOTES, 'UTF-8' ) );
+	}
+
+	/**
+	 * The custom text shaped into a Bluesky post body.
+	 *
+	 * Hard-clamped to the 300-grapheme limit so an over-long custom text is
+	 * shortened rather than rejected, matching the short-form path.
```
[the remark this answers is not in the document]
pfefferle: Fixed in 43c7401 — the docblock now states the clamp is by code points (`mb_strlen` via `truncate_text()`) toward Bluesky's 300-grapheme limit, conservative and matching the short-form path.

src/editor-plugin/plugin.js
pfefferle: Thanks! Took your wording but trimmed it since (as you noted) the full version ran long — fixed in 5b5150d:

> Leave empty to use the default message, or write your own. It's shared with a link back to this post, and you can mention other Bluesky users.

Keeps the default-vs-custom choice, the link-back, and the @-mention note.

src/editor-plugin/plugin.js
pfefferle: Done in 5b5150d — dropped `__nextHasNoMarginBottom` from the share toggle so its default bottom margin gives the custom-text field room to breathe.

(FWIW I first tried wrapping the controls in a `VStack`, but the plugin's `@wordpress/no-unsafe-wp-apis` lint rule disallows `__experimentalVStack`, so the toggle margin is the simpler lint-clean fix.)

src/editor-plugin/plugin.js
```diff
@@ -52,11 +59,10 @@ const EditorPlugin = () => {
 			title={ __( 'Bluesky', 'atmosphere' ) }
```
jeherve: Should we add a butterfly icon, like we do for the Fediverse panel?

src/editor-plugin/plugin.js
pfefferle: Updated approach in 15f417e: rather than the toggle margin (which does nothing in WP 7.0 — the default `.components-base-control` bottom margin is gone), the panel now sets `className="block-editor-block-inspector"`, exactly like this plugin's ActivityPub counterpart. That pulls in core's `.block-editor-block-inspector .components-base-control:where(:not(:last-child)) { margin-bottom: 16px; }` rule, so the controls space themselves with no custom CSS. Thanks for the nudge — it sent me to the right reference.

jeherve (dismissed):
This tests well for me. I'd say it's good to merge. I only have a tiny suggestion.

src/editor-plugin/plugin.js
pfefferle: Done in e985c7f. There's no non-emoji butterfly glyph (the only Unicode butterfly is the 🦋 emoji), so rather than a character I added the Bluesky butterfly as an inline `<SVG>` (`currentColor`, `aria-hidden`), placed after the title like the Fediverse panel's `⁂`.

While there, I also renamed the panel from "Bluesky" to **ATmosphere** (matching the admin menu / settings page) and relabelled the toggle to "Share this post", since it actually gates all sharing — the Bluesky post *and* the standard.site document — not just Bluesky. The custom-text field stays Bluesky-specific.

jeherve (dismissed):

pfefferle:
Screenshot 2026-06-24 at 08 27 26

jeherve (approved):

PR #157, opened 2026-06-24
Surface quoted posts when importing Bluesky replies as comments

### LINKED ISSUE #133: Imported Bluesky comments drop quote-post embeds (app.bsky.embed.record)
## Problem

When a Bluesky reply quotes another post, the quote is lost on import. The reply's record carries an \`app.bsky.embed.record\` pointing at the quoted post's AT-URI, but the inbound reaction-sync path only reads \`text\` and ignores \`embed\` entirely.

## Example

Reply record \`app.bsky.feed.post/3mnzucbkwfs2h\` includes:

\`\`\`json
"embed": {
  "$type": "app.bsky.embed.record",
  "record": {
    "cid": "bafyreifz7yief4dwg7wgyotql3zkknx4nolwcl2ukio6jdld4ejn7wclfm",
    "uri": "at://did:plc:4i6hvdii3km3kbnj3losmwnt/app.bsky.feed.post/3mnzu7nvcss2e"
  }
}
\`\`\`

The imported WordPress comment shows none of this — the quoted post is dropped.

## Proposed direction

In the same inbound-enrichment pass that applies facets (see the companion facets issue), detect \`app.bsky.embed.record\` (and the \`#view\` / \`recordWithMedia\` variants) and surface the quoted post — e.g. append a blockquote with a link to \`https://bsky.app/profile/<handle>/post/<rkey>\`, optionally resolving the quoted record for its text.

## Priority

Follow-up to the facets fix; can land separately.

jeherve:
Fixes #133

## Proposed changes:

* A Bluesky reply that quotes another post carries the quoted post's AT-URI in the reply record's `embed` (`app.bsky.embed.record`, or `recordWithMedia` when the quote also carries media), not in `text` — so `Reaction_Sync` was silently dropping the quote on import.
* This detects the quote embed in the same inbound pass that resolves facets and appends a `<blockquote>` linking to the quoted post's `bsky.app` page, built from the AT-URI's DID so no handle lookup is needed; only `app.bsky.feed.post` quotes get a link, and every nested shape is `is_array()`-guarded against malformed PDS JSON.
* The empty-text gate now runs *after* the quote is resolved, so a quote-only reply (empty `text`, populated `embed`) imports instead of being dropped.

### Other information:

## Testing instructions:

* From Bluesky, reply to that post and use the "quote post" action to quote another Bluesky post (with or without text in the reply itself).
* Run the reaction sync (`wp atmosphere sync-reactions`, or wait for the scheduled run) and open the imported comment.
* **Before:** the quoted post is dropped entirely. **After:** the comment shows the reply text followed by a blockquote linking to the quoted post on Bluesky. A reply that is *only* a quote (no text) now imports too.

### Changelog entry

A changelog entry is already committed on this branch (`.github/changelog/fix-comment-quote-embed`), so no auto-entry is needed.

### DIFF
--- a/.github/changelog/fix-comment-quote-embed
+++ b/.github/changelog/fix-comment-quote-embed
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Bluesky replies that quote another post now keep a link to the quoted post when imported as comments, instead of dropping it.
--- a/includes/class-reaction-sync.php
+++ b/includes/class-reaction-sync.php
@@ -566,10 +566,6 @@ class Reaction_Sync {
 
 		$text = $record['text'] ?? '';
 
-		if ( '' === $text ) {
-			return false;
-		}
-
 		/*
 		 * Bluesky stores `text` as the display string — long URLs are
 		 * truncated to e.g. `bsky.app/profile/jere...` and the real
@@ -583,6 +579,28 @@ class Reaction_Sync {
 		 */
 		$facets = $record['facets'] ?? array();
 		$text   = Facet::apply( $text, \is_array( $facets ) ? $facets : array() );
+
+		/*
+		 * A reply that quotes another post carries the quoted post's
+		 * AT-URI in `embed` (app.bsky.embed.record), not in `text`.
+		 * Surface it as a linked blockquote so the quote isn't dropped on
+		 * import; append after the reply text, mirroring how Bluesky
+		 * renders the quote card below the reply.
+		 */
+		$quote = self::build_quote_html( $record );
+
+		if ( '' !== $quote ) {
+			$text = '' === $text ? $quote : $text . "\n\n" . $quote;
+		}
+
+		/*
+		 * Drop replies that carry neither text nor a resolvable quote —
+		 * the gate is here, after both are resolved, so a quote-only
+		 * reply (empty `text`, populated `embed`) still imports.
+		 */
+		if ( '' === $text ) {
+			return false;
+		}
 
 		/**
 		 * Filters whether a reply should be synced as a WordPress comment.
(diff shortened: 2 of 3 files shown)


pfefferle:
## Review

### Findings (all minor / non-blocking)

**Pattern consistency — `quoted_post_web_url()`**
- Reimplements AT-URI parsing (`explode('/', substr($uri, 5))` + count check) that already exists as `Atmosphere\parse_at_uri()` (`includes/functions.php`), which returns exactly `['did','collection','rkey']|false`. Reusing it would be cleaner and matches the established helper.

**Pattern consistency / cross-PR — hardcoded `https://bsky.app/...`**
- This adds another hardcoded appview host. PR #159 centralizes all five existing ones behind `appview_url()` + the `atmosphere_appview_host` filter, and **also edits `class-reaction-sync.php`**. Whichever lands second must reconcile — and the new quote link won't honor the appview filter unless routed through `appview_url()`. Worth coordinating so the quote link isn't the one site that ignores the filter.

**Minor consistency — `quoted_post_web_url()`**
- `$repo` (the DID) is concatenated into the URL without `rawurlencode()`, unlike the sibling `build_bsky_web_url()` which encodes its handle. Not a security hole — `build_quote_html()` runs the result through `esc_url()` and the comment is `wp_kses_post()`-gated — but inconsistent with the neighbor.

Everything else is clean: every nested embed level is `is_array()`/`is_string()`-guarded against malformed PDS JSON, the empty-text gate is correctly relocated after quote resolution (quote-only replies now import), output is escaped + kses-gated, the link text is i18n'd with the right text domain, and the perf-conscious choice to *not* fetch the quoted post's text is documented. The 9 new tests (incl. a malformed-embed data provider) all pass.

```
+------------------------------------------------------------------+--------+
| Check                                                            | Result |
+------------------------------------------------------------------+--------+
| Follows WordPress coding standards and best practices            |   ✅   |
| No security issues (escaping, sanitization, SQL, XSS, auth)      |   ✅   |
| No new warnings, fatals, errors, or type errors (PHP/TS)         |   ✅   |
| Function calls verified (exist, correct params, return handled)  |   ✅   |
| No potential mid-deploy issues                                   |   ✅   |
| Follows existing patterns in the repo                            |   ❌   |
| i18n handled (translations, text domain, plural forms)           |   ✅   |
| Accessibility (a11y) handled for UI changes                      |   ✅   |
| No performance or caching regressions                            |   ✅   |
| Backwards compatible (filters, hooks, REST, exports, schema)     |   ✅   |
+------------------------------------------------------------------+--------+
```

**Risk Assessment: MEDIUM** — modifies the live reaction-import path and changes comment content shown to readers, but scope is bounded and defensively guarded.

**Approval: 90%**

The ❌ is only the pattern items above — all minor. The one worth acting on is reusing `parse_at_uri()` and deciding how this interacts with #159's `appview_url()`, so the quote link doesn't become the lone host that ignores the filter.

jeherve:
Thanks for the review @pfefferle! Addressed in `9b8e1eb` (after merging trunk, which brought in #159's `appview_url()`, in `736c4fd`):

- **Reuse `parse_at_uri()`** — `quoted_post_web_url()` now calls `Atmosphere\parse_at_uri()` instead of reimplementing the `explode`/`count` parsing.
- **Route through `appview_url()`** — the quote link is now built via `appview_url()`, so it honors the `atmosphere_appview_host` filter like every other appview link instead of being the lone hardcoded `bsky.app`. Added a test (`test_process_reply_quote_honors_appview_host_filter`) that swaps the host to `deer.social` and asserts the quote link follows.
- **`rawurlencode()` on the DID** — I went the other way here for consistency: the two *other* `profile/<did>/post/<rkey>` builders (`Atmosphere::*` and `Facet`) pass the DID **raw**, and only *handles* get `rawurlencode()`d. A DID's colons are valid unencoded path chars the appview expects, and encoding them (`did%3Aplc%3A…`) would make this the one site that diverges from those siblings (and risk the link). Added a code comment noting why.

All 775 tests pass; `composer lint` is clean on the changed files.

pfefferle (approved):

PR #159, opened 2026-06-24
Make the Bluesky appview host filterable

### LINKED ISSUE #158: Make the Bluesky appview host filterable
We build Bluesky web links (`https://bsky.app/profile/...`, `/hashtag/...`) in a handful of places across the plugin, and the `bsky.app` domain is hardcoded in each of them.

That works fine for most people, but `bsky.app` isn't the only way to view `app.bsky.*` records. Some folks would rather send their readers to an alternative AT Protocol appview, mu.social, for example. Today there's no way to do that without editing the plugin.

I'm wondering if we could extract the URL building into a single helper and add a filter so the appview host can be swapped out. Host-only is enough for the realistic case: alternative appviews tend to mirror bsky.app's `/profile/...` and `/hashtag/...` routes, so we don't need to rewrite full paths.

Rough idea:

- One helper, `Atmosphere\appview_url( $path, $context )`, that owns the `https://` scheme and the host.
- A new `atmosphere_appview_host` filter, defaulting to `bsky.app`, that receives the path and whatever context the call site has on hand (type, did, handle, rkey, tag) so a callback can route conditionally if it wants to.
- All the existing call sites keep escaping at the point of use; the helper just builds the string.

These are all display-side links, so none of this touches the records we actually publish to the network. It only changes where we point people when they click through.

jeherve:
Fixes #158

## Proposed changes:

We build Bluesky web links (`https://bsky.app/profile/...`, `/hashtag/...`) in five places across the plugin, with the `bsky.app` host hardcoded in each. This pulls that construction into a single helper and adds filtering so people can point readers at an alternative AT Protocol appview (deer.social, a self-hosted one, etc.).

* New helper `Atmosphere\appview_url( string $path, array $context = array() )` owns the `https://` scheme and the host, and runs it through the filters below.
* **`atmosphere_appview_host`** (default `bsky.app`) swaps the host. It accepts a bare host, a subdomain, or a host with a path prefix (`something.social/atblue`), with or without a scheme or trailing slash — the value is normalized with `wp_parse_url` (host lower-cased, scheme clamped to http/https, no doubled/empty segments), so appviews on a subpath work cleanly. A malformed return falls back to `bsky.app` (and flags `_doing_it_wrong`).
* **`atmosphere_appview_url`** wraps the fully assembled URL, so a callback can rewrite the whole route — e.g. `/account/<did>` instead of `/profile/<did>`, or a custom hashtag route — using the parts in `$context` (`type`, `did`, `handle`, `rkey`, `tag`).
* All five call sites go through the helper and keep escaping at the point of use (`esc_url_raw` for stored/meta values, `esc_url` for HTML). The helper returns an unescaped string on purpose, so escaping stays as late as possible.

These are all display-side links, so nothing here touches the records we publish to the network. It only changes where we point people when they click through. :)

### Other information:

Unit tests for the helper (default host, filter override, subpath/subdomain bases, scheme preserve/clamp, port, trailing-slash and malformed-host handling, both filters' argument passing, and a full-URL route rewrite) plus an end-to-end test that runs the real Facet render path through the filter.

## Testing instructions:

* With no filter, confirm Bluesky links still point at `bsky.app` — open a synced post's "View on Bluesky" link, and check rendered reaction/comment links and facet mentions/hashtags. Nothing should change.
* Swap the host with a snippet (plugin or theme `functions.php`):
  ```php
  add_filter( 'atmosphere_appview_host', fn() => 'something.social/atblue' );
  ```
  Render a post with facet mentions/hashtags and confirm links become `https://something.social/atblue/...`.
* Rewrite a route with the URL filter:
  ```php
  add_filter( 'atmosphere_appview_url', function ( $url, $path, $context ) {
      return 'mention' === ( $context['type'] ?? '' )
          ? 'https://my.appview/account/' . $context['did']
          : $url;
  }, 10, 3 );
  ```
* Of note: links rendered on the fly pick up the filters immediately, while author/source links already stored on synced reaction comments keep the host they were synced with. Documented in `docs/developer-docs.md`.

### Changelog entry

Changelog entry already added manually on the branch (`.github/changelog/add-filterable-appview-host`), so no need to auto-generate one here.

### DIFF
--- a/.github/changelog/add-filterable-appview-host
+++ b/.github/changelog/add-filterable-appview-host
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Add a filter so links to Bluesky can point at an alternative AT Protocol appview.
--- a/.github/changelog/add-filterable-appview-host
+++ b/.github/changelog/add-filterable-appview-host
@@ -1,4 +1,4 @@
 Significance: minor
 Type: added
 
-Add a filter so links to Bluesky can point at an alternative AT Protocol appview.
+Add filters so links to Bluesky can point at an alternative AT Protocol appview, including ones hosted on a subdomain or subpath.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -72,6 +72,8 @@ Text domain: always `'atmosphere'`.
 
 **MUST** use `use` imports for cross-namespace references — no inline `\Namespace\Class`.
 
+**MUST** build any front-end-displayed Bluesky/appview web link (`profile`, `post`, `hashtag`, `mention`) through `Atmosphere\appview_url( $path, $context )` in `includes/functions.php` — never hardcode `https://bsky.app/...`. The helper centralizes the host and makes it filterable (`atmosphere_appview_host` / `atmosphere_appview_url`). Keep escaping at the call site (`\esc_url()` for HTML, `\esc_url_raw()` for storage); the helper returns an unescaped URL on purpose. This applies to display links only, not to AT Protocol records being published.
+
 ## Autoloading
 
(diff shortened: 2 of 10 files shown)


pfefferle (commented):
Shouldn't we make the full URL filterable?

pfefferle (dismissed):

pfefferle (approved):

PR #165, opened 2026-06-25
Add @handle.tld mention support: profile links and Bluesky notifications

### LINKED ISSUE #142: Auto-link @mentions in post content and trigger Bluesky notifications
## Summary

Support `@handle` mentions inside post content so that they (a) auto-link on the web / in the `site.standard.document` entry, and (b) trigger Bluesky notifications for the mentioned accounts — similar to how the ActivityPub plugin handles `@user@domain` mentions.

## Background

Most of the plumbing already exists:

* `includes/transformer/class-facet.php` is a full rich-text facet pipeline. `Facet::mentions()` finds `@handle.tld` patterns and `resolve_mention()` resolves each handle to a DID (via `_atproto.<handle>` DNS TXT lookup, falling back to `did:web:`), emitting proper `app.bsky.richtext.facet#mention` facets.
* `Facet::extract()` is already wired into the outbound Bluesky post record in `class-post.php` (link-card `transform()` at the main path and teaser-thread entries) and in `class-comment.php` (outbound replies).

A `app.bsky.richtext.facet#mention` facet in the published `app.bsky.feed.post` record is exactly what makes Bluesky's AppView notify the mentioned DID. So notifications already fire **for any** `@handle.tld` **that survives into the post** `text` **field**.

## The gap

For the default long-form post, `build_text()` (`class-post.php`) composes the companion Bluesky post as `title + excerpt + permalink`, where the excerpt is the manual excerpt or the first ~30 words of the body. **A** `@handle` **written deeper in the post body never reaches the companion Bluesky post text → no facet → no notification.** (Short-form formats and teaser-threads that already carry more body text work today, since `Facet::extract()` runs on whatever ships.)

There is also no display-side auto-linking for mentions in the rendered web content / `site.standard.document` entry — content parsers just render `the_content`.

## Proposed work

Two distinct pieces:

1. **Notifications (Bluesky side).** Collect mentions from the post content (not just title/excerpt) and ensure they are carried into the companion Bluesky post text so a mention facet attaches. Reuse the existing `class-facet.php` layer — do not rebuild it. Mind the 300-grapheme limit and multibyte byte-offset math when assembling the post text.
2. **Auto-linking (web / standard.site side, display only).** Linkify `@handle.tld` mentions in the rendered content to the mentioned account's Bluesky profile page.

## ActivityPub coexistence

The ActivityPub plugin hooks `the_content` at priority 99 and rewrites mentions to anchors. Any display-side linkifier here must:

* Target the Bluesky `@handle.tld` form (a bare domain-form handle), distinct from ActivityPub's webfinger-style `@user@domain.tld`, to avoid double-linking.
* Skip text already wrapped in an `<a>` by another filter (pick a priority and respect existing anchor wrapping either way).

The two notification systems are orthogonal: ActivityPub notifies via inbox delivery to the mentioned actor; Bluesky notification is purely a side effect of the mention facet existing in the published post. They don't collide.

## Security note

`resolve_mention()` (`class-facet.php`) already documents that resolving arbitrary handles fires DNS lookups against attacker-controlled domains. Broadening mention collection to the full post body widens that surface — consider the allowlist / skip-on-commenter notes already in that docblock.

## Related

* Automattic/wordpress-atmosphere#132 (closed) — inbound comment-import facet handling, the inverse direction.

jeherve:
Fixes https://github.com/Automattic/wordpress-atmosphere/issues/142

## Proposed changes:

Adds Bluesky `@handle.tld` mention support so authors can notify accounts and link to their profiles. This works in two places: a display-side linkifier (`includes/class-mention.php`) wraps bare handles in `the_content` as profile links routed through the `appview_url()` helper (so they honor the `atmosphere_appview_host` filter), and a publishing-side carry-over pulls resolvable body mentions into long-form and teaser-thread Bluesky post text so mentioned accounts are notified even when the mention is buried deep in the body. It also replaces the unreliable `did:web:<handle>` resolution fallback in `Facet::resolve_mention()` with the real handle-resolution chain (`Resolver::handle_to_did()`), memoized per request to bound lookups.

### Other information:

## Testing instructions:

* Publish a post whose body mentions a real Bluesky account using `@handle.tld` (e.g. `@bsky.app`).
* On the front end, confirm the handle renders as a link to the account's profile on your site's appview host.
* On Bluesky, confirm the published post (including long-form and teaser-thread variants) carries the mention so the account is notified, and that the mention resolves to the correct DID rather than a dead `did:web` profile.
* Run `composer lint` and `npm run env-test` — both pass (786 tests).

### Changelog entry

A changelog entry is already committed at `.github/changelog/add-content-mentions` (Significance: minor, Type: added).


includes/class-mention.php
```diff
@@ -0,0 +1,170 @@
...
+			if ( \preg_match( '#^<(/)?([a-z0-9]+)\b[^>]*>$#i', $chunk, $m ) ) {
+				$tag = \strtolower( $m[2] );
+				if ( '/' === $m[1] ) {
+					$i = \array_search( $tag, $tag_stack, true );
+					if ( false !== $i ) {
+						$tag_stack = \array_slice( $tag_stack, 0, $i );
+					}
+				} else {
+					$tag_stack[] = $tag;
+				}
```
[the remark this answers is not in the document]
jeherve: Good catch — fixed in 9ae65d7. The closing-tag handler now unwinds to the *most recently* opened tag of that name (`array_keys( $stack, $tag, true )` + `end()`) instead of the first match, so nested same-name protected tags like `<code><code>…</code>…</code>` keep the outer tag's text protected. Added a regression test (`test_skips_nested_same_name_protected_tags`).

includes/transformer/class-facet.php
```diff
@@ -20,6 +22,34 @@
...
+	/**
+	 * Regex matching an AT Protocol `@handle.tld` mention.
+	 *
+	 * Capture group 1 is the bare handle (no leading `@`). Requires at
+	 * least two dot-separated labels, mirroring DNS-name handle syntax.
+	 * Shared by {@see self::mentions()} and {@see self::resolve_handles()}.
+	 *
+	 * @var string
+	 */
+	private const MENTION_PATTERN = '/@([a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(\.[a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)+)/u';
```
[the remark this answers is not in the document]
jeherve: Fixed in 9ae65d7. Added the same `(?<![\w@.])` boundary guard that `Mention::linkify()` already uses to `Facet::MENTION_PATTERN`, so `bob@example.com` and `@user@domain.tld` no longer match — no resolution lookup and no stray `#mention` facet. New regression test `test_email_and_webfinger_domain_halves_are_not_mentions` covers both cases, with a `pre_http_request` tripwire asserting that no network lookup fires.

tests/phpunit/tests/transformer/class-test-post.php
```diff
@@ -31,9 +32,47 @@ public function tear_down() {
 		\remove_all_filters( 'atmosphere_transform_bsky_post' );
 		\remove_all_filters( 'atmosphere_post_embed' );
 		\remove_all_actions( 'atmosphere_long_form_strategy_downgraded' );
```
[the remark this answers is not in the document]
jeherve: Fixed in 9ae65d7 — `tear_down()` now calls `\remove_all_filters( 'pre_http_request' )`, matching the inline cleanup the other `pre_http_request` tests in this file already do. (The WP test harness also restores hooks from a per-class snapshot, so it did not leak across tests in practice, but the explicit cleanup is clearer and consistent with the surrounding code.)

tests/phpunit/tests/transformer/class-test-facet.php
```diff
@@ -15,6 +15,53 @@
...
+	 * `Facet::$resolution_cache` is static, so a DID resolved (or a miss
+	 * cached) in one test would otherwise leak into the next.
+	 */
+	public function tear_down() {
+		$cache = new \ReflectionProperty( Facet::class, 'resolution_cache' );
+		$cache->setAccessible( true );
+		$cache->setValue( null, array() );
+
+		parent::tear_down();
+	}
```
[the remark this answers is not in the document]
jeherve: Fixed in 9ae65d7 — `tear_down()` now calls `\remove_all_filters( 'pre_http_request' )` so the stub never outlives the test that set it.

tests/phpunit/tests/transformer/class-test-facet.php
```diff
@@ -15,6 +15,53 @@
...
+	 *
+	 * Handles resolve via `_atproto.<handle>` DNS first (empty for these
+	 * fixtures) and then `https://<handle>/.well-known/atproto-did`, which
+	 * is the request this short-circuits. A handle absent from the map (or
+	 * mapped to an empty string) is left unresolved.
+	 *
+	 * @param array<string,string> $map Handle => DID to return.
+	 */
+	private function mock_handle_resolution( array $map ) {
+		add_filter(
```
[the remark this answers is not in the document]
jeherve: Fixed in 9ae65d7 — prefixed to `\add_filter()` to match the namespaced-global convention used everywhere else in the suite.

pfefferle (commented):
Some things that I checked:

* The @-mention detection misinterprets domain like WebFinger handles as profile handles: `@notiz.blog@notiz.blog`
* There is no lookup if accounts really exist, so even an example mention like `@example.com` will be linked.

Not (yet) checked: Does the implementation also takes care of the recent implemented "Custom Bluesky text"!?

pfefferle (changes requested):

jeherve:
Thanks @pfefferle — addressed all three. Two new commits: 4aab0d2 (tests for the custom-text interaction) and a3a2dd1 (the WebFinger fix + shared pattern + filter).

### 1. `@notiz.blog@notiz.blog` misread as a profile handle — fixed

The detection only had a *leading* boundary, so it caught the domain half of `@user@domain.tld` but not a WebFinger handle whose **user** half is itself domain-shaped (`@notiz.blog@notiz.blog`) — the leading `@notiz.blog` was linked and faceted.

Added a trailing `(?![\w@])` boundary so a handle immediately followed by `@` (or another word char) is rejected. `.` is deliberately kept out of the trailing class so a handle ending a sentence (`@bsky.app.`) still matches. While here, I made `Facet::MENTION_PATTERN` the single source of truth and pointed the display linkifier at it too, so the publish path and the front-end can't drift apart again (the bug existed precisely because they were two separate regexes). Regression tests added on both sides.

### 2. No existence check — `@example.com` still gets linked

Worth splitting this into the two paths, because they already behave differently:

- **Publishing / notifications:** existence *is* checked. `Facet::resolve_mention()` runs the real handle-resolution chain and only mints a `#mention` facet when the handle resolves to a DID. So `@example.com` never produces a Bluesky mention or notification — it ships as plain text. (Pinned by `test_unresolvable_mention_produces_no_facet`.)
- **Display linkifier:** intentionally does **no** network lookup. A per-render resolution would turn every front-end page view into an outbound DNS/HTTP request to each mentioned domain — a real perf and availability/SSRF-amplification concern (a post mentioning `@attacker.example` would make every visitor hit that host). So by default we link by syntax and let the appview render an unknown handle gracefully, same as the ActivityPub plugin's display-side linkifier.

To still give a seam for stricter sites, I added an `atmosphere_link_mention` filter — return `false` for a handle to leave it as plain text (e.g. gate on a cached existence check or an allowlist) without imposing the lookup cost on everyone by default.

If you'd prefer the plugin itself to gate display links on existence, the non-network way to do it would be to persist the publish-time resolved handle→DID set to post meta and only link handles present there — happy to do that as a follow-up if you think it's worth the coupling.

### 3. Does it handle the Custom Bluesky text?

Yes — custom text takes precedence, and the interaction is now pinned by tests (commit 4aab0d2):

- When custom text is set, the record text is **exactly** what the author wrote. The body-mention carry-over is intentionally *not* applied — injecting a handle the author chose to omit would break the "post exactly what I typed" contract. A body-buried mention isn't even resolved on that path (`test_custom_text_does_not_carry_body_only_mention`, with an HTTP tripwire).
- A handle the author types **into** the custom text still works: facet extraction runs on the final record text regardless of which branch composed it, so it produces a `#mention` facet and notifies (`test_custom_text_mention_still_produces_facet`).

`composer lint` and the full PHPUnit suite pass.

pfefferle:
**Review — `@handle.tld` mention support**

Synthesized from three parallel lensed passes (security · correctness · WP-standards/i18n/BC) plus a manual read. Net: solid, well-defended feature; one blocking correctness bug and a consistency fix, then easy to merge.

### Correctness — blocking — `includes/transformer/class-post.php` (~L751)
The `} elseif ( '' !== $permalink && $text === $permalink )` branch sets `$suffix = $permalink` instead of `$sep . $permalink`. When a handle is carried in, the mention glues to the URL (`@handle.tldhttps://…`); `MENTION_PATTERN`'s trailing `(?![\w@])` then over-extends into a bogus handle, so **no `#mention` facet is emitted — a silent notification loss**. Reachable via an empty-title + empty-excerpt post (`array_filter` at L667 collapses `$parts` to `[$permalink]`). Fix: mirror the sibling branch (`$suffix = $sep . $permalink`) and add a test for that post shape.

### Consistency — should fix — `class-post.php` (~L851)
`carry_mentions_into_teaser()` uses `break` on the first overflowing handle (dropping all shorter ones after it); `carry_body_mentions()` correctly uses `continue`. The teaser path also lacks the `debug_log` drop-accounting the link-card path has. Make it `continue` + log to match.

### Robustness (not XSS) — `includes/class-mention.php` (L29)
`PROTECTED_TAGS` omits `script`, `noscript`, `svg`, `iframe`, `title`, so a `@handle` inside a `<script>`/`<svg>` body gets linkified into `<a>` and corrupts the element. Not an XSS (output is `esc_url`/`esc_html`'d; KSES blocks these tags from non-`unfiltered_html` users), but worth the one-line array extension.

### Minor / hygiene
- `mb_strlen` vs `grapheme_length` in both carry methods (`class-post.php` ~L769/775/799/803/846/851) — safe (never exceeds the 300 cap) but over-truncates prose with emoji and is inconsistent with the rest of the file.
- Missing `@since unreleased` on `Facet::MENTION_PATTERN`, `Facet::resolve_handles()`, and `class-mention.php`'s header/class + the `atmosphere_link_mention` filter (the release script needs it).
- `Facet::is_valid_handle()` duplicates `Resolver::is_valid_handle()` and has drifted (no reserved-TLD check); Resolver re-validates so SSRF stays closed, but prefer calling the canonical copy.
- `$resolution_cache` caches transient-failure misses, so one network blip can deflate `#mention` facets across a WP-CLI/cron bulk run — consider caching only successes.
- Tag regex skips hyphenated custom elements; `mock_handle_resolution()` duplicated across 3 test files; silent 1 MiB `the_content` guard; no test for the truncate-link carry path.

### Verified solid
Output escaping; SSRF gate (`wp_safe_remote_get` + reserved-TLD rejection + `is_valid_handle`); **no ReDoS** (bounded quantifiers + 1 MiB cap); single-source `MENTION_PATTERN` with correct email/WebFinger boundaries; `without_links()` try/finally guard; projection + redacted guards keep DNS off the display and editor-preview paths; correct `PREG_OFFSET_CAPTURE` byte offsets; hit+miss memoization; reflection-based cache reset in tests; the `did:web`→real-resolution regression test pinned to a real handle.

**Risk: MEDIUM** (display linkifier runs on `the_content` site-wide, but read-only and well-guarded). Blocking item is just the separator bug — fix that + the teaser `break`/`continue` and this is good to go.

jeherve:
Thanks for the thorough review. Addressed in 8382ebb:

**Blocking — separator bug:** The bare-permalink branch in `carry_body_mentions()` now keeps the `\n\n` separator (`$suffix = $sep . $permalink`). Added `test_bare_permalink_carries_body_mention_with_separator` and confirmed it reproduces the bug — reintroducing the old code fails with the exact `@alice.bsky.socialhttp://…` symptom, the fix makes it pass. (Hitting that branch needs an empty title *and* an excerpt that sanitises to empty, so the test uses an HTML-comment body plus a `the_content` filter that pushes the rendered body onto the link-card path.)

**Should fix — teaser consistency:** `carry_mentions_into_teaser()` now uses `continue` instead of `break` and emits the same `debug_log` drop-accounting as the body-mention path.

**Robustness:** `PROTECTED_TAGS` extended with `script`, `noscript`, `svg`, `iframe`, `title` (with a parametrized test), and the tag regex now matches hyphenated custom elements (`[a-z][a-z0-9-]*`) so they track correctly on the stack.

**Minor / hygiene:**
- Both carry methods now count graphemes instead of `mb_strlen` code points.
- `@since unreleased` added to `Facet::MENTION_PATTERN`, `Facet::resolve_handles()`, the `Mention` class, and the `atmosphere_link_mention` filter.
- `Facet` now calls the canonical `Resolver::is_valid_handle()`; the drifted local copy (missing the reserved-TLD check) is removed.
- The resolution cache stores successes only, so a transient lookup failure can't deflate `#mention` facets across a WP-CLI/cron bulk run.

**Deliberately left** (happy to follow up if you'd like):
- `mock_handle_resolution()` duplicated across 3 test files — pure test hygiene; left out to keep this diff focused on passing files.
- The silent 1 MiB `the_content` guard — it runs on every front-end render and mirrors ActivityPub's own silent guard, so logging there risks noise.
- A dedicated test for the short-form truncate-link carry path.

`composer lint` and the full PHPUnit suite (806 tests) are green.

pfefferle:
## Code review — typed multi-lens pass

Reviewed with three parallel lenses (correctness/fatals, security/escaping, perf-patterns-i18n-compat), each reading whole files. Blocking findings were confirmed against the code and `origin/trunk`. `phpcs` passes clean.

### 🔴 Blocking

**1. Uncapped full-body handle resolution — `includes/transformer/class-facet.php:395-419` + `includes/transformer/class-post.php:721-727`**
`body_mentions()` runs `Facet::resolve_handles()` over the **entire untruncated post body** (`render_post_content_plain()`), with **no cap** on distinct handles. Every unresolvable handle now costs a DNS TXT lookup *plus* a `wp_safe_remote_get( 'https://<handle>/.well-known/atproto-did', timeout 10 )` (`class-resolver.php:45-49`). An Author-role user can publish a body with thousands of `@fake-N.example.com` tokens → tens of thousands of seconds of blocking outbound I/O in one `atmosphere_publish_post` cron run, aimed at attacker-chosen public hosts. Backfill (500 posts/chunk) multiplies it. Every other `Facet::extract()` caller has an implicit ~300-grapheme bound; this one doesn't. **Suggest a hard cap on distinct handles resolved per call** (precedent: `META_ORPHAN_RECORDS` caps at 10).

**2. Egress widening onto the lower-trust comment path — `includes/transformer/class-facet.php:487-516`**
On trunk, `resolve_mention()` did DNS-only then fabricated `did:web:<handle>` (no HTTP). It now calls `Resolver::handle_to_did()`, which makes a real HTTPS request. `Comment::transform()` calls `Facet::extract()` on **commenter-supplied content** (`class-comment.php:100`), so an approved comment containing `@target.example.com` makes the server issue outbound HTTPS to an arbitrary public host. Bounded by the 300-grapheme comment cap and `wp_safe_remote_get` (blocks internal hosts) — so public-host egress amplification, not internal SSRF. The docblock at `class-facet.php:475-481` names "skip mention resolution on the commenter path" as the fix but doesn't implement it. **Gate the HTTP fallback off the comment path, or document as accepted risk.**

**3. Preview/publish grapheme-count divergence — `includes/transformer/class-post.php:721-727`, `225-247`**
`body_mentions()` returns `array()` in projection mode (good — no DNS in preview), but the carry-over **mutates the composed text** (inserts a whole `@handle` line), unlike facet extraction which only annotates. So `project()` measures text *without* the carried line and reports e.g. "120/300" to the pre-publish panel while the real record is longer — and near the cap the real prose can be silently truncated shorter than previewed. This violates the documented invariant ("*the grapheme count the preview reports is identical either way*", `class-post.php:225-229`).

**4. Self-closed protected tag suppresses linkification — `includes/class-mention.php:93-113`**
The tokenizer detects closing tags only by a leading `/` after `<`, not a trailing `/>`. A self-closed protected tag (`<iframe … />`, `<svg/>`) is pushed onto `$tag_stack` and never popped, so **every mention after it in the render is silently left as plain text** — no error, no log. Verified against real `wp_html_split()` output.

### 🟡 Non-blocking

- **No cross-request cache (`class-facet.php:67`).** `$resolution_cache` is per-request static only; publishing is a fresh process per post, so backfill/republish re-resolves the same handle every time. `docs/php-coding-standards.md` prescribes a transient layer here; `Reaction_Sync::resolve_author()` is the precedent (`HOUR_IN_SECONDS`).
- **Undocumented behavior change.** Dropping the `did:web` fallback means unresolvable handles now produce *no* `#mention` facet (was a dead-link facet). Legitimate fix, but the changelog only mentions the new feature — worth a line so "notifications stopped" reports trace back here.
- **`resolve_mention()` swallows `WP_Error` silently** (`class-facet.php:502-505`) — route through `debug_log()` to distinguish "doesn't exist" from "transiently unreachable".
- **Display links skip `Resolver::is_valid_handle()`** — the linkifier links reserved TLDs (`server.local`) the publish path would reject; cosmetic dead-link drift from the "single source of truth" claim.
- **Minor:** consecutive `//` comment paragraphs at `class-base.php:200-203` and `class-facet.php:507-510` (repo convention is `/* */` blocks); missing `@since` on new `Mention` members and `$resolution_cache`; `carry_body_mentions()`/`carry_mentions_into_teaser()` duplicate a greedy-fit loop; `build_text()` `$available <= 0` path bypasses carry-over without a drop log.

### ✅ Positives
Escaping is solid (`esc_url`/`esc_html` at every call site); the display linkifier does zero network I/O by design; projection correctly skips DNS; `MENTION_PATTERN` correctly rejects webfinger/email collisions; `without_links()` uses `try/finally` restoring prior state; the `did:web`→real-resolution swap is a real bug fix; success-only caching is well-reasoned. `phpcs` clean.

---

**Risk: HIGH** — changes the core publish path for every post (full-body scan + live handle resolution) and adds network egress to both the publish and comment-sync paths.

Findings #1–#4 are concrete; #1 and #2 (egress) are the most important. Happy to draft fixes.

includes/transformer/class-post.php
```diff
@@ -705,7 +706,194 @@ private function build_text(): string {
...
+			// straight onto the URL (`@handle.tldhttps://…`), which over-extends
+			// MENTION_PATTERN and drops the #mention facet entirely.
+			$suffix = $sep . $permalink;
+			$prose  = '';
+		}
+
+		// Handles not already visible in the prose, in order.
+		$missing = array();
+		foreach ( $handles as $handle => $did ) {
+			if ( false === \mb_stripos( $prose, '@' . $handle ) ) {
```
kraftbj: The 'already visible' check is an unanchored substring test, so a resolvable body handle that is a string-prefix of a handle already in the composed text is treated as present and dropped. Example: title `News about @alice.company`, body `shout-out @alice.com` — `resolve_handles()` returns `alice.com`, but `mb_stripos( $prose, '@alice.com' )` matches inside `@alice.company`, so `@alice.com` is never carried and that account is silently not notified — which is the feature's core promise. Same test in `carry_mentions_into_teaser()` against `$shipped` (line 853). Suggest requiring a token boundary after the match (next char not in `[\w.@-]`), or comparing handle sets extracted via `MENTION_PATTERN` rather than substring containment.

includes/transformer/class-post.php
```diff
@@ -705,7 +706,194 @@ private function build_text(): string {
...
+	 * mentions over DNS — see {@see self::$projecting}).
+	 *
+	 * @return array<string,string>
+	 */
+	private function body_mentions(): array {
+		if ( $this->is_redacted() || $this->projecting ) {
+			return array();
+		}
+
+		return Facet::resolve_handles( $this->render_post_content_plain( $this->object ) );
```
kraftbj: The body-mention scan runs on `render_post_content_plain()`, which strips all tags to plain text — losing the protected-tag contexts the display linkifier deliberately honors. A `@handle.tld` inside a `<code>` sample or an existing `<a href>` link in the post body gets carried into the Bluesky text and emits a `#mention` facet + notification, even though `Mention::linkify()` leaves that exact handle unlinked on the front end. Publish and display then disagree on what counts as a mention, and an account named only inside a `<code>` block gets pinged. Consider extracting carry-over candidates with the same protected-tag rules as the display side, or skipping mentions inside `code`/`pre`/links before resolving.

kraftbj (commented):
Two additional items from a review pass, both inline below. Neither is covered by the existing threads (verified against the prior passes).

includes/transformer/class-post.php
jeherve: Fixed in 71e4983. The "already visible" test no longer uses `mb_stripos()` — both carry paths now compare handle *sets* extracted via `MENTION_PATTERN` (new `Facet::handles_in()`), so `@alice.com` is carried even when `@alice.company` is present. Regression test `test_link_card_carries_prefix_collision_mention` pins the title `@alice.company` / body `@alice.com` case and asserts both distinct `#mention` facets survive.

includes/transformer/class-post.php
jeherve: Fixed in 71e4983. `body_mentions()` no longer scans fully tag-stripped text — it runs the rendered HTML through `Mention::strip_protected()` (which shares the exact protected-tag tokenizer with the front-end linkifier) before resolving. A `@handle` inside `<code>`/`<pre>` or an existing `<a href>` is dropped, so publish and display agree. Regression test `test_link_card_does_not_carry_mention_inside_protected_tag` asserts a `<code>`-wrapped handle is neither carried into the text nor faceted.

jeherve:
Thanks @pfefferle — all four blocking findings from the HIGH-risk pass are addressed across three commits, each with regression coverage. `composer lint` and the full PHPUnit suite (833 tests) pass.

1. Uncapped full-body handle resolution → a5530fd
`Facet::resolve_handles()` now stops after `MAX_RESOLVED_HANDLES` (20) distinct handles and logs once when it truncates, so an author can't fan a body of `@fake-N.example.com` tokens into tens of thousands of DNS/HTTPS lookups. 20 is generous headroom over any post that can fit in 300 graphemes while keeping egress bounded. Test: `test_resolve_handles_caps_distinct_lookups` (30 handles → 20 lookups).

2. Egress widening onto the comment path → a5530fd
`Facet::extract()` gained a `$with_mentions` flag; `Comment::transform()` passes `false`, so commenter-supplied `@handle` content no longer triggers any DNS/HTTPS resolution (link/hashtag facets still emit — they never touch the network). The `resolve_mention()` docblock that named this fix now points at the implemented gate. Tests: `test_comment_mention_is_not_resolved` (tripwire on `pre_http_request`) and `test_extract_without_mentions_skips_resolution`.

3. Preview/publish grapheme divergence → 71e4983
Projection now sizes the carry-over instead of skipping it: in projecting mode `body_mentions()` returns the *syntactic* body handles (via `Facet::handles_in()`, no DNS) as an upper bound, so the reported grapheme count is never shorter than the record the publisher writes. Test: `test_projection_accounts_for_carried_body_mention` asserts the projected count equals the published record length.

4. Self-closed protected tag suppresses linkification → 2106156
The tokenizer no longer pushes self-closed tags (`<svg/>`, `<iframe … />`) onto the tag stack, so they can't leave a phantom protected tag open for the rest of the render. Test: `test_self_closed_protected_tag_does_not_swallow_later_mentions`.

On the non-blocking items: the `resolve_mention()` docblock is updated to reflect the new comment-path gate and the resolve cap; success-only caching stays as-is (deliberate, and now documented alongside the cap). The cross-request transient cache and the "unresolvable handle now yields no facet" changelog note I'd prefer to keep as separate follow-ups rather than fold into these correctness commits — happy to open an issue for them if you'd like.

includes/transformer/class-post.php
```diff
@@ -789,7 +797,220 @@ private function build_text(): string {
...
+	 * {@see self::$projecting}), but the carry-over still lengthens the composed
+	 * record at publish. So the preview returns the *syntactic* candidate
+	 * handles (via {@see Facet::handles_in()}, no lookups) with empty DIDs — an
+	 * upper bound, since some may not resolve — so the carry-over sizing, and
+	 * therefore the reported grapheme count, never under-reports the record the
+	 * publisher will write.
+	 *
+	 * @return array<string,string>
+	 */
+	private function body_mentions(): array {
```
pfefferle, line 838: Protected-tag mentions are only stripped on this carry-over path — the short-form and truncate-link texts ship the raw body verbatim, and `Facet::extract()` still resolves and mints a `#mention` facet for a handle inside `<code>`, `<pre>`, or an existing `<a>`. A titleless short-form note whose body contains `<code>@alice.bsky.social</code>` (or a truncate-link post with the same) publishes a record with a real `#mention` facet — alice gets a Bluesky notification — while the front end deliberately leaves that handle unlinked: the exact publish/display disagreement the `strip_protected()` docblock says it prevents.

includes/class-mention.php
```diff
@@ -0,0 +1,268 @@
...
+	 * on real resolution ({@see Facet::resolve_mention()}), so a non-existent
+	 * `@example.com` never notifies anyone; the display link merely points at
+	 * the appview, which renders an unknown handle gracefully. A site that
+	 * wants stricter display links can veto a handle through the
+	 * `atmosphere_link_mention` filter.
+	 *
+	 * @param string $text Plain-text chunk (no tags).
+	 * @return string
+	 */
+	private static function linkify( string $text ): string {
```
pfefferle: `linkify()` runs `MENTION_PATTERN` per text chunk, so the `(?<![\w@.])` lookbehind can't see characters that sit in the previous chunk across a tag boundary, defeating the email/WebFinger guards. Content like `<b>bob</b>@example.com` (any inline markup ending right before the `@`) is the email address bob@example.com, but the chunk handed to `linkify()` starts at `@example.com`, so the lookbehind passes and the front end wraps the domain half of the email in a bogus `https://bsky.app/profile/example.com` mention link. The publish path (which matches against joined plain text) correctly skips it, so display and record disagree.

includes/transformer/class-post.php
```diff
@@ -789,7 +797,220 @@ private function build_text(): string {
...
+			// MENTION_PATTERN and drops the #mention facet entirely.
+			$suffix = $sep . $permalink;
+			$prose  = '';
+		}
+
+		// Handles not already visible in the prose, in order. Compare handle
+		// sets on token boundaries, not by substring: a plain `mb_stripos`
+		// would treat `@alice.com` as present inside `@alice.company` and
+		// silently skip carrying (and notifying) it.
+		$present = Facet::handles_in( $prose );
```
pfefferle: `carry_body_mentions()` computes the missing-handle set against the pre-truncation prose, then shrinks the prose to make room for the carried line — a handle that only appeared in the truncated-away tail is deleted from the final text without being carried or logged. Example: the body mentions @alice.bsky.social near the end of the excerpt (so it counts as "present" and is skipped from `$missing`) and also mentions @carol.example.com, absent from the excerpt; carrying @carol pushes the prose over budget, `truncate_text()` cuts the tail containing @alice.bsky.social, and the published record silently loses Alice's mention — no facet, no notification — even though the untruncated composed text (≤300 graphemes) contained it.

includes/transformer/class-post.php
```diff
@@ -789,7 +797,220 @@ private function build_text(): string {
...
+	 * publisher will write.
+	 *
+	 * @return array<string,string>
+	 */
+	private function body_mentions(): array {
+		if ( $this->is_redacted() ) {
+			return array();
+		}
+
+		$plain = sanitize_text( Mention::strip_protected( $this->render_post_content_html( $this->object ) ) );
```
pfefferle: `body_mentions()` pipes `strip_protected()` output through `sanitize_text()`, whose `wp_strip_all_tags()` removes tags without inserting whitespace — gluing the character before a protected element onto a mention that immediately follows it, so the `MENTION_PATTERN` lookbehind rejects it. `Great tool<code>cc</code>@alice.bsky.social` becomes `Great toolcc@alice.bsky.social`: the mention is neither carried nor resolved, while the front-end linkifier (operating on the `@alice.bsky.social` text chunk) links it — parity broken in the opposite direction from the protected-tag gap above.

includes/transformer/class-facet.php
```diff
@@ -292,7 +369,7 @@ private static function links( string $text ): array {
 	 */
 	private static function mentions( string $text ): array {
 		$facets  = array();
-		$pattern = '/@([a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(\.[a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)+)/u';
+		$pattern = self::MENTION_PATTERN;
```
pfefferle: `MAX_RESOLVED_HANDLES` caps only `resolve_handles()`; `mentions()` (called via `extract()` per record at publish) resolves every syntactic mention with no cap, and resolution misses are deliberately not cached, so each record — and each teaser-thread entry — re-runs the full DNS + HTTPS chain per failing handle. A post packed with ~35 distinct unresolvable `@fake-N.example.com` tokens per 300-grapheme chunk, published as a 5-entry teaser thread, costs a blocking DNS TXT lookup plus a `wp_safe_remote_get()` well-known fallback (~5s timeout each) per miss per entry — a single publish can spend minutes in sequential network calls and exceed `max_execution_time`, despite the 20-handle cap suggesting egress is bounded.

includes/class-mention.php
```diff
@@ -0,0 +1,268 @@
...
+					 * stack top, so only it is popped; with same-name nesting
+					 * (e.g. `<code><code>…</code>…</code>`) popping the first
+					 * match would drop the still-open outer tag and linkify
+					 * text that is in fact still protected.
+					 */
+					$keys = \array_keys( $tag_stack, $tag, true );
+					if ( ! empty( $keys ) ) {
+						$tag_stack = \array_slice( $tag_stack, 0, \end( $keys ) );
+					}
+				} elseif ( ! \preg_match( '#/\s*>$#', $chunk ) ) {
```
pfefferle: The self-closing check `#/\s*>$#` misclassifies an opening tag whose last unquoted attribute value ends in `/` (e.g. `<a href=https://example.com/>`) as self-closed, so the tag is never pushed onto the protected stack. Legacy/unquoted markup `<a href=https://example.com/>see @alice.bsky.social</a>` is treated as if the `<a>` closed immediately: the mention inside the existing link is linkified anyway, emitting an `<a>` nested inside an `<a>` — invalid HTML that browsers split into broken adjacent links, double-linking text the protected-tag rule was supposed to leave alone.

includes/transformer/class-facet.php
```diff
@@ -12,24 +12,101 @@
...
+	 * Bluesky handle. Both boundaries keep these false positives from driving
+	 * real DNS/HTTP resolution or minting a bogus `#mention` facet. A `.` is
+	 * deliberately left out of the trailing class so a handle ending a
+	 * sentence (`@bsky.app.`) still matches.
+	 *
+	 * @since unreleased
+	 *
+	 * @var string
+	 */
+	public const MENTION_PATTERN = '/(?<![\w@.])@([a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(\.[a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)+)(?![\w@])/u';
```
pfefferle: The new lookbehind `(?<![\w@.])` includes `.`, so a mention directly after a period — the Twitter-style dot-mention idiom, `.@alice.bsky.social check this out` — no longer matches, whereas trunk's boundary-free pattern matched it and minted a `#mention` facet. Neither the publish path nor the front-end linkifier matches it now: no facet, no notification, no link — a silent regression relative to trunk for that input.

includes/transformer/class-comment.php
```diff
@@ -97,7 +97,15 @@ public function transform(): array {
...
-		$facets = Facet::extract( $text );
+		/*
+		 * Do not resolve @mentions on the comment path. The comment body is
+		 * third-party (commenter-supplied) content, and mention resolution
+		 * issues live DNS + HTTPS lookups to the mentioned host; extracting
+		 * mentions here would let an approved comment steer the server's
+		 * outbound requests at an arbitrary public host. Link and hashtag
+		 * facets, which never touch the network, are still emitted.
+		 */
+		$facets = Facet::extract( $text, false );
```
pfefferle: Replacing `Facet::extract( $text )` with `Facet::extract( $text, false )` removes mention facets from synced comments entirely — trunk resolved commenter @mentions and emitted `#mention` facets; the branch never does, and no other code path re-establishes them. Understood this is intentional SSRF hardening, but it's a user-visible capability removal with no setting or filter to opt legitimate comment authors back in — worth an explicit decision (or a filter) rather than a silent change.

includes/transformer/class-facet.php
```diff
@@ -371,66 +533,63 @@ private static function hashtags( string $text ): array {
...
-	 * character outside `[A-Za-z0-9-]` — including percent-encoded
-	 * forms.
-	 *
-	 * @param string $host Handle to validate.
-	 * @return bool
-	 */
-	private static function is_valid_handle( string $host ): bool {
-		if ( '' === $host || \strlen( $host ) > 253 ) {
-			return false;
+		$did = Resolver::handle_to_did( $handle );
```
pfefferle, line 645: The deleted `did:web:` fallback meant any well-formed handle always produced a mention facet even when DNS lookup failed; the replacement returns `''` on any `Resolver::handle_to_did()` `WP_Error`, so a transient DNS/HTTPS failure at publish time now silently and permanently drops the mention. Example: @example.com whose DID genuinely is `did:web:example.com`, resolver outage at publish time → trunk fell back and emitted a correct facet; now the record is written without the mention, and since misses aren't cached but the record is already published, nothing ever repairs it short of re-publishing.

includes/class-mention.php
```diff
@@ -0,0 +1,268 @@
...
+	 * @param string $content Rendered HTML.
+	 * @return string
+	 */
+	public static function strip_protected( string $content ): string {
+		if ( '' === $content ) {
+			return $content;
+		}
+
+		// Same pathological-input guard as the linkifier.
+		if ( \strlen( $content ) > MB_IN_BYTES ) {
```
pfefferle: `strip_protected()` returns the content unstripped when the rendered HTML exceeds `MB_IN_BYTES`, so for very large posts `body_mentions()` collects and carries handles from inside `<code>`/`<pre>`/`<a>` regions while the front-end linkifier (which also bails at 1 MB) shows them as plain text. A >1 MB post whose only `@alice.bsky.social` sits inside a code sample gets Alice notified by a published `#mention` the site's own display never renders as a link — the same divergence class as the short-form gap, triggered from the size-guard side.

pfefferle (commented):
Ran a deep review pass over the mention handling; 10 inline notes below, roughly ordered by severity.

The common thread in several of them (the protected-tag gap, per-chunk linkification, tag-strip gluing, the unquoted-attribute self-close check, and the 1 MB guard) is publish/display parity: the publish path and the front-end linkifier tokenize the content differently, so they disagree about what is a mention. A structural fix — both sides consuming one shared tokenization — would close that class at once, rather than patching case by case.

Three others are deliberate-looking behavior changes vs trunk (dot-mentions no longer matching, comment mention facets removed, the dropped `did:web:` fallback) that probably each deserve an explicit decision rather than a silent change.

includes/transformer/class-post.php
jeherve: Fixed in 29654c0. Protected-tag stripping was only on the carry-over path — now every post path (short-form, truncate-link, teaser, link-card) goes through the same `Mention` tokenizer. `Mention::classify_handles()` surfaces the protected-only handles, and `Facet::extract()` takes a deny-set (`protected − linkable`) that skips minting a `#mention` for a handle living only inside `<code>`/`<pre>`/`<a>`. So a titleless short-form note (or truncate-link post) with `<code>@alice.bsky.social</code>` ships the text but no facet and no notification, matching display. Regression: `test_short_form_does_not_mint_mention_inside_protected_tag`.

includes/class-mention.php
jeherve: Fixed in 29654c0. `walk()` now threads the last plain-text character across inline tags, and `linkify()` prepends it as a synthetic boundary before matching, so the `(?<![\w@])` lookbehind sees across the tag split. `<b>bob</b>@example.com` is now read as the email it renders as (no link) on the display side, matching the publish path's joined-text view. Regression: `test_skips_handle_glued_across_inline_tag`.

includes/transformer/class-post.php
jeherve: Fixed in 29654c0. `carry_body_mentions()` now fits the carried line to a fixpoint: after shrinking the prose it re-checks which handles actually survived and pulls any present-but-truncated-away handle into the carried line rather than dropping it silently (still logged as dropped only if it genuinely can't fit). Your exact scenario is the regression `test_carry_rescues_mention_truncated_by_shrink` — I verified it fails against the old single-pass logic and passes with the fixpoint.

includes/transformer/class-post.php
jeherve: Fixed in 29654c0. The publish side no longer flattens with `sanitize_text()` before detection — it collects handles from the same `Mention::classify_handles()` walk the linkifier uses, so `Great tool<code>cc</code>@alice.bsky.social` is classified identically on both sides. With the shared cross-tag boundary the handle reads as glued to the preceding word on *both* sides (so neither links it), closing the divergence. The parity is now structural: one tokenizer feeds both paths.

includes/transformer/class-facet.php
jeherve: Fixed in 29654c0. `Facet::mentions()` (the per-record `extract()` path) now caps distinct handle resolution at `MAX_RESOLVED_HANDLES` and memoizes within-call misses, so a record — or teaser-thread entry — packed with `@fake-N.example.com` tokens can no longer fan out into an unbounded run of blocking DNS + HTTPS lookups. Regression: `test_extract_caps_mention_resolution` (30 distinct unresolvable handles → 20 lookups).

includes/class-mention.php
jeherve: Fixed in 29654c0. `is_self_closing()` strips quoted attribute values first, then only treats a trailing `/>` as self-closing when the `/` sits right after the tag name (`<svg/>`) or is separated from the last attribute by whitespace, a quote, or `=`. `<a href=https://example.com/>` (unquoted value ending in `/`) now stays open, so a mention inside the still-open anchor remains protected and isn't double-linked. Regression: `test_unquoted_slash_attribute_does_not_self_close`.

includes/transformer/class-facet.php
jeherve: Fixed in 29654c0. Dropped the `.` from the leading boundary — `(?<![\w@.])` → `(?<![\w@])` — so the dot-mention idiom `.@alice.bsky.social` resolves and links again on both paths, as it did on trunk. Email / WebFinger false positives are still excluded by the `\w` and `@` boundaries (those always have a word char or `@` before the domain half). Regressions: `test_extract_dot_mention`, `test_links_dot_mention`.

includes/transformer/class-comment.php
jeherve: Addressed in 2b41fb8. Kept the SSRF-safe default (off), but the removal is no longer silent: an `atmosphere_resolve_comment_mentions` filter (passed the `WP_Comment`) lets a site that trusts its moderated commenters opt their @mentions back in. Explicit, opt-in decision rather than a hard `false`.

includes/transformer/class-facet.php
jeherve: Addressed in 29654c0. Kept the no-`did:web`-fabrication default on purpose — for the dominant `*.bsky.social` case the real DID is a `did:plc:…` over the well-known endpoint, so a `did:web:<handle>` guess mints a facet pointing at a dead profile — but the "transient failure permanently drops the mention" concern is now a deliberate, filterable decision. `atmosphere_mention_didweb_fallback` (passed the handle and the resolver `WP_Error`, whose code distinguishes a definitive miss from a transient network failure) restores the pre-1.0 fallback for a did:web-hosting site. Regression: `test_didweb_fallback_is_filterable`.

includes/class-mention.php
jeherve: Fixed in 29654c0. `Mention::classify_handles()` bails on >1 MB content and returns empty sets, and the post path additionally gates mention extraction off entirely (`mentions_enabled()`) when the rendered body exceeds 1 MB — so a >1 MB post links nothing on the front end *and* mints nothing on publish. Same size guard drives both sides. Regression: `test_classify_handles_bails_on_huge_content`.

jeherve:
Addressed all 10 notes across 29654c0 (parity + hardening) and 2b41fb8 (comment-mention filter); replies inline under each.

Took the structural route you suggested: the protected-tag gap, per-chunk linkification, tag-strip gluing, the unquoted-attribute self-close, and the 1 MB guard are all now driven by a single `Mention` tokenizer — `Mention::classify_handles()` produces the linkable/protected split that both the display linkifier and `Facet::extract()` consume, so publish and display can no longer disagree about what is a mention.

The three behaviour-vs-trunk notes each got an explicit decision: dot-mentions restored (regex boundary), comment mentions kept off but now opt-in via `atmosphere_resolve_comment_mentions`, and the `did:web` fallback kept off (a guess links to a dead profile for the `*.bsky.social` majority) but restorable via `atmosphere_mention_didweb_fallback`.

`composer lint` clean; full suite green (888 tests).

pfefferle (approved):

Issue #166, opened 2026-06-25
Support Standard.site social lexicons


pfefferle:
## Context

Issue #163 updated the core Standard.site publishing compatibility surface for publication and document records, but intentionally left the social lexicons out of scope.

The current Standard.site docs list:

- `site.standard.graph.subscription` for user-publication relationships
- `site.standard.graph.recommend` for document recommends

## Scope

Decide whether ATmosphere should publish or manage either social record type directly, or whether they belong in a separate app/service integration.

## Acceptance criteria

- Subscription and recommend use cases are mapped to WordPress/ATmosphere product behavior.
- OAuth scope needs are checked against Standard.site docs.
- If supported, implementation has focused tests and user-facing settings or filters as appropriate.
- If deferred, documentation explains why the plugin only publishes publication/document records.

PR #167, opened 2026-06-25
Close Standard.site compatibility gaps

### LINKED ISSUE #163: Close Standard.site compatibility gaps
## Context

Standard.site's public docs list several current fields and verification behaviours that ATmosphere should either support directly or intentionally document as out of scope.

Public references:

- Document lexicon: https://standard.site/api/docs/markdown/lexicons/document
- Publication lexicon: https://standard.site/api/docs/markdown/lexicons/publication
- Verification: https://standard.site/api/docs/markdown/verification
- Quick Start: https://standard.site/api/docs/markdown/quick-start
- Docs index: https://standard.site/llms.txt

## Gaps to close or explicitly defer

- Publication URL normalization: the publication lexicon says `url` should avoid trailing slashes. `Publication::transform()` currently uses `home_url( '/' )`, which produces a trailing slash.
- Document metadata fields: the document lexicon now lists optional `links`, `labels`, and `contributors`. ATmosphere currently emits the main document fields, but these are not first-class fields.
- Publication metadata fields: the publication lexicon lists optional `labels`, `preferences`, and `preferences.showInDiscover`. ATmosphere currently emits `url`, `name`, `description`, `icon`, and `basicTheme`, but not these fields.
- OAuth scopes: the public quick start documents `site.standard.authFull` and `site.standard.authSocial`. ATmosphere currently requests and advertises `atproto transition:generic identity:handle`; we should verify whether the Standard.site scopes should be added, replace the generic scope, or be documented as not required for the current supported PDS flow.
- Non-root publications: the verification docs describe `/.well-known/site.standard.publication/path/to/publication`. ATmosphere currently models a single root publication. We should either support non-root publication verification or document root-only publication support.
- Social lexicons: the public docs list `site.standard.graph.subscription` and `site.standard.graph.recommend` as core lexicons. ATmosphere does not currently publish or manage these records. These are probably follow-up feature issues, but should be triaged.

## Acceptance criteria

- Publication records avoid trailing slashes in `url`, with test coverage.
- Document support for `links`, `labels`, and `contributors` is either implemented with focused APIs/filters and tests, or explicitly documented as extension-only.
- Publication support for `labels` and `preferences.showInDiscover` is either implemented with focused APIs/settings/filters and tests, or explicitly documented as extension-only.
- OAuth scope handling is checked against the current public Standard.site docs and reflected in both the authorization request and client metadata when needed.
- Non-root publication support is either implemented or documented as out of scope.
- Social lexicons are triaged into separate follow-up issues if they are not included here.

## Non-goals

- Do not add fields that are not present in the current public Standard.site docs.
- Do not change Bluesky post embedding behaviour as part of this issue.

### LINKED ISSUE #169: Tighten OAuth scopes from transition:generic to granular repo: scopes
## Where we are today

Our OAuth client requests this scope string (in both `Client::SCOPES` and the client-metadata REST endpoint, which have to stay in lockstep):

```
atproto transition:generic identity:handle
```

`transition:generic` is the App Password-equivalent bucket. It does the job, but it's broad: it grants write access to *every* collection in the connected repo, not just the ones we actually touch. When you think about least-privilege, that's more than we need.

In practice we only ever write to three top-level collections:

- `app.bsky.feed.post` — posts, threads, and comment replies
- `site.standard.document` — one document per post
- `site.standard.publication` — the site's publication record

Of note: picking a different content format in the settings (HTML / Leaflet / Pckt / Markpub / a custom parser) doesn't change this. The parser output is nested inside the `content` field of the `site.standard.document` record, so it's always the same three collections regardless of the chosen format.

## What I'm wondering

I'm wondering if we could swap `transition:generic` for the granular `repo:` scopes that match what we actually write:

```
atproto
repo:app.bsky.feed.post
repo:site.standard.document
repo:site.standard.publication
identity:handle
```

Each `repo:<collection>` scope covers create/update/delete for that collection *plus the blobs those records reference*, so `uploadBlob` comes along for free. That cleanly covers our entire write path while dropping the blanket "can write anything" access we have today.

## The caveat

It's not quite that clean on the read side. The reaction-sync feature also calls a couple of app-view methods:

- `app.bsky.actor.getProfile`
- `app.bsky.notification.listNotifications`

Under the fine-grained model those need `rpc:` scopes with a pinned audience, which is fiddlier and more brittle than the read side of `transition:generic`. So there's a middle-ground option worth considering: tighten the *writes* to the three `repo:` scopes (the part that actually matters for least-privilege) and keep a small read bucket for those calls, rather than going fully granular everywhere.

I'd lean toward that middle ground, but I'm open to either.

## Things to keep in mind

- **Lockstep requirement.** Any scope change has to land in *both* `Client::SCOPES` (`includes/oauth/class-client.php`) and the client-metadata endpoint (`includes/rest/class-client-metadata-controller.php`). If they drift, the auth server silently downgrades every connection to the smaller value.
- **Migration impact.** OAuth tokens are scoped at issue time, so already-connected accounts won't pick up the new scopes until they reconnect. Same situation we ran into with #53. We'd want to make sure publishing keeps working for existing connections during the transition.

This isn't urgent — what we ship today works fine. It's more about cleaning up the permissions we ask for so we're only requesting what we genuinely use. What do you think? :)

### References

- AT Protocol permission spec: https://atproto.com/specs/permission
- Prior scope work: #53 (added `identity:handle`)

pfefferle:
Fixes #163
Fixes #169

## Proposed changes:
* Normalize Standard.site publication URLs without trailing slashes.
* Add focused extension filters for optional Standard.site document and publication metadata.
* Keep the documented `include:site.standard.authFull` OAuth scope in both authorization and client metadata for Standard.site compatibility.
* Replace broad `transition:generic` access with granular `repo:` scopes for `app.bsky.feed.post`, `site.standard.document`, and `site.standard.publication`.
* Add the image blob scope and exact Bluesky AppView RPC scopes needed by media upload and reaction sync.
* Document root-only publication verification and social lexicons as out of scope, with #166 tracking the social follow-up.

### Other information:

- No screenshots; this is publishing/OAuth metadata behavior with developer documentation.

## Testing instructions:
* Run `composer lint`.
* Run `npm run env-test -- --filter="Client_Authorize|Client_Metadata_Filter"`.
* Run `npm run env-test`.

### Changelog entry

Manual changelog entry included in `.github/changelog/update-standard-site-compatibility`.

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

-   [x] Changed - for changes in existing functionality

#### Message

Standard.site records and OAuth permissions are now more compatible with current long-form publishing tools and discovery.

</details>


includes/transformer/class-publication.php
```diff
@@ -90,6 +90,48 @@ public function transform(): array {
...
+		 *
+		 * Return true or false to emit `preferences.showInDiscover`.
+		 * Return null to omit the preference and let downstream appviews
+		 * apply their own default.
+		 *
+		 * @since unreleased
+		 *
+		 * @param bool|null $show_in_discover Whether to show in discovery, or null to omit.
+		 */
+		$show_in_discover = \apply_filters( 'atmosphere_publication_show_in_discover', null );
```
jeherve: What if instead of defaulting to `null`, we looked at the site's `blog_public` option, and defaulted to true when your site is public, while still having a filter so you can toggle that off if you want to.

jeherve (commented):
this tests well for me after another try. I think it didn't work for me yesterday because I had to re-authenticate with the new scopes for the cards to work.

I left a small question below.

includes/transformer/class-publication.php
pfefferle: Good idea — implemented in `975cdf7`. `preferences.showInDiscover` now defaults to your `blog_public` setting (public site → `true`, "discourage search engines" → `false`), so it follows the visibility preference already set under Settings → Reading. The `atmosphere_publication_show_in_discover` filter still overrides it, or returns `null` to omit the field entirely. Added tests for the public / private / override / omit cases.

jeherve (approved):

PR #170, opened 2026-06-25
Add typed AT Protocol previews

### LINKED ISSUE #168: Support typed AT Protocol record previews
## Context

The `?atproto` preview endpoint currently renders only the `site.standard.document` record for a singular WordPress post. That makes it hard to inspect the other records ATmosphere can publish for the same content, especially the Bluesky `app.bsky.feed.post` output and any long-form teaser/thread variants.

A typed query parameter would make the endpoint more useful for debugging and review, for example:

- `?atproto` or `?atproto=document` for the existing `site.standard.document` preview.
- `?atproto=bsky` or `?atproto=post` for the Bluesky post record preview.
- Potentially `?atproto=publication` for the site-level `site.standard.publication` record.
- Potentially `?atproto=all` to inspect every record shape the plugin can produce for the post.

## Proposal

Extend the preview endpoint to accept a type selector while preserving the current default behavior.

## Acceptance criteria

- Existing `?atproto` URLs keep returning the `site.standard.document` preview.
- A typed selector can preview at least the document record and the Bluesky post record.
- Unsupported selector values return an appropriate error status instead of silently falling back.
- The preview remains permission-gated to users who can edit posts.
- Test coverage verifies the default behavior, typed behavior, and invalid type behavior.

## Non-goals

- Do not change publishing behavior.
- Do not expose previews to anonymous visitors.
- Do not perform network writes or uploads from the preview endpoint.

### LINKED ISSUE #175: the_content chrome (Jetpack sharing/likes/related, etc.) leaks into AT Protocol records
### Summary
Third-party plugins that hook `the_content` to append display chrome — Jetpack's **Sharing** ("Share this:"), **Likes** ("Like this:"), **Related Posts**, subscribe forms, ad blocks, etc. — have their output pulled into the records ATmosphere publishes. The chrome ends up in both the `app.bsky.feed.post` text and the `site.standard.document` content, where it's noise (and sometimes broken markup).

### Steps to reproduce
1. On a site running Jetpack with Sharing/Likes/Related Posts enabled, publish a post.
2. View the record preview, e.g. `https://example.com/<post>/?atproto=all` (logged in with `edit_posts`).
3. Observe the share/like/related markup inside the rendered content.

Live example: `https://matthiaspfefferle.blog/2026/06/30/hallo/?atproto=all`

### Expected
Records contain only the author's article content — no sharing/likes/related/subscribe chrome.

### Actual
Display-only additions injected by other plugins' `the_content` callbacks are serialized into the published Bluesky post and standard.site document.

### Root cause
ATmosphere renders post content with `apply_filters( 'the_content', … )`, which fires **every** registered `the_content` callback — including display chrome from other plugins. Three call sites:
- `includes/transformer/class-base.php:199` — `render_post_content_plain()` → Bluesky post text
- `includes/transformer/class-post.php:1575` — rich render → inline-link facets
- `includes/content-parser/class-parser-base.php:143` — `site.standard.document` HTML content

### Proposed fix
Centralize content rendering behind a single guard (mirroring the existing `Mention::without_links()` wrapper) that suppresses known display chrome for the duration of the render, using each plugin's own opt-out hooks where available (`sharing_show`, `jetpack_relatedposts_filter_enabled_for_request`, Likes visibility) rather than fragile `remove_filter()` calls, plus an `atmosphere_`-prefixed filter so integrations can extend the suppression list. All three call sites route through it.

### Notes
- Affects published records, not just the preview — the preview just makes it visible.
- Should land with regression tests that register a fake `the_content` chrome filter and assert it's excluded from the post text and the document content.

pfefferle:
Fixes #168
Fixes #175

## Proposed changes:
* Add typed `?atproto` preview selectors: bare `?atproto` / `?atproto=site.standard.document`, `?atproto=app.bsky.feed.post`, front-page `?atproto=site.standard.publication`, `?atproto=all` (every family keyed by `$type`), and a clear JSON error for unsupported selectors.
* Resolve the whole preview endpoint in a dedicated `Atmosphere\Transformer\Preview` class instead of the `Atmosphere` class. Dispatch keys off each transformer's own collection NSID (`Base::get_collection()`), so a selector maps straight to a transformer with no per-type `switch` — this removes ~380 lines from the `Atmosphere` class.
* Add `Base::get_preview_records()` (default: the single `transform()` record; overridden by `Post` to project a long-form post into its thread) so every transformer is previewable through one method, mirroring the publish path without writing blobs or post meta.
* Add the `atmosphere_atproto_preview_transformers` filter so third-party plugins extend the preview with their own lexicon by appending a `Base` subclass — no new interface to implement. It supersedes the previous records-keyed filter (both unreleased).
* Fail safe: a non-array filter return falls back to the built-in transformers (`_doing_it_wrong`), non-`Base` entries are ignored, and a transformer sharing a built-in NSID supersedes it (matching `Content_Parser\Registry`).
* Document the endpoint and its extension filter in `docs/developer-docs.md`.
* Keep front-end display extras (sharing buttons, Related Posts, ad units) out of the previewed record by marking the preview as a REST request — the same conditions under which those plugins bail — so it renders the way the post is actually published, instead of leaking chrome the publish path never includes (#175).

### Other information:

- No visual UI changes.

#### Known follow-up (pre-existing, not introduced here)
The default `get_preview_records()` path calls `Document`/`Publication::transform()`, which can attempt a live thumbnail blob upload when the attachment blob cache is cold (the `Post` path already suppresses this via projection mode). This behaviour predates this PR; making the static `Post::upload_thumbnail()` projection-aware is tracked as a separate change so it can be reviewed against the publish path.

## Testing instructions:
* Open a published post URL while logged in as a user with `edit_posts`.
* Confirm `?atproto` still returns a `site.standard.document` record.
* Confirm `?atproto=app.bsky.feed.post` returns the Bluesky post record (or a thread for a long-form post).
* Confirm `?atproto=all` on a post returns an object keyed by `site.standard.document` and `app.bsky.feed.post`.
* Confirm `?atproto` / `?atproto=site.standard.publication` on the site front page returns the publication record.
* Confirm an unsupported selector returns a JSON error.
* Optionally register a `Base` subclass via `atmosphere_atproto_preview_transformers` and confirm it appears under its NSID and in `?atproto=all` (see `docs/developer-docs.md → Previewing AT Protocol Records`).
* Run `composer lint`.
* Run `npm run env-test`.

### Changelog entry

Manual changelog entry included in `.github/changelog/add-typed-atproto-preview`.

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Preview AT Protocol output by record type, including an all-record view for comparing WordPress, standard.site, and Bluesky records.

</details>


pfefferle:
Maybe we should require an active session before allowing access?

jeherve (commented):
This looks good to me. 

> Maybe we should require an active session before allowing access?

By active session, do you mean that your site should be connected to an AT account? In this case, should we instead display a message saying that there is no active connection when you try to preview? That may help us when debugging.

pfefferle:
@jeherve with active session, I meant that only logged-in users could query them!?

jeherve:
> @jeherve with active session, I meant that only logged-in users could query them!?

Isn't that already the case with the `if ( ! \current_user_can( 'edit_posts' ) ) {` check?

pfefferle:
The question was more, do we want to gate it or not? Should it be public like `?activitypub`

jeherve:
Ah, understood. It doesn't expose any private information, and the information is available publicly and searchable with tools like pds.ls, so I think we could open it up.

pfefferle:
Let's keep them behind the session check for now and check that in a different PR.

includes/transformer/class-preview.php
```diff
@@ -0,0 +1,254 @@
...
+		$type = \sanitize_text_field( (string) $type );
+
+		/*
+		 * This runs on a singular / front-page front-end request, so display
+		 * plugins (sharing buttons, Related Posts, ad units, …) would hook
+		 * `the_content` and bleed their chrome into the previewed record —
+		 * chrome the real publish path never picks up, because it renders
+		 * outside the loop / main query. Those plugins already bail on REST
+		 * requests, so mark this one as REST to render the record the way it
+		 * is actually published. The handler exits below, so the constant
```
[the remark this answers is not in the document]
pfefferle: The leakage described in #175 only reproduces through the preview endpoint (`?atproto=…`), which runs on a real singular front-end request inside the loop — that's exactly why other plugins' `the_content` chrome shows up there. The publish path renders `the_content` *outside* the loop/main query (not `is_singular()`, not `in_the_loop()`, typically under REST or cron), so Sharing, Likes, Related Posts, etc. already bail there on their own guards — the published records were never leaking. Marking the preview as a REST request simply puts it under the same conditions the publish path already meets, so it renders the way the post is actually published. That fully closes the reproducible bug. A global suppression guard across every render site isn't needed for the publish path and would risk stripping content the publish path legitimately includes, so I've kept the fix scoped to the preview.

includes/transformer/class-preview.php
```diff
@@ -0,0 +1,254 @@
...
+	 * the post's record families; on the site front page they expose the
+	 * `site.standard.publication` record. Requires the edit_posts
+	 * capability — other requests fall through untouched.
+	 */
+	public static function render(): void {
+		$type = \get_query_var( 'atproto', null );
+
+		if ( null === $type || \is_array( $type ) ) {
+			return;
+		}
```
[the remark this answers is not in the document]
pfefferle: `render()` is a thin `template_redirect` wrapper: it gates on `edit_posts`, marks the request REST, routes front-page vs. singular, and delegates the actual record resolution to `for_post()` / `for_site()` — both thoroughly covered here (selectors, `all`, unknown types, filter extension, fallbacks). It ends in `exit`, so it can't be exercised in-process. This matches how the plugin's other `template_redirect` exit handlers (`serve_wellknown_atproto_did` / `serve_wellknown_publication`) are covered — the resolvable logic is unit-tested, the exit wrapper isn't. Extracting an exit-free core purely to test the wrapper adds shipping-code churn for near-zero gain over the existing resolver tests, so I'm leaving it as-is.

includes/transformer/class-preview.php
```diff
@@ -0,0 +1,254 @@
...
+
+		/*
+		 * This runs on a singular / front-page front-end request, so display
+		 * plugins (sharing buttons, Related Posts, ad units, …) would hook
+		 * `the_content` and bleed their chrome into the previewed record —
+		 * chrome the real publish path never picks up, because it renders
+		 * outside the loop / main query. Those plugins already bail on REST
+		 * requests, so mark this one as REST to render the record the way it
+		 * is actually published. The handler exits below, so the constant
+		 * never outlives this preview request.
```
jeherve: > so the constant never outlives this preview request.

I think that in some scenarios (when we `return` without `exit`ing), it would outlive the request. Maybe we should only set `REST_REQUEST` when we `exit`?

pfefferle: Good catch — you're right. The `define()` ran before the routing, so the `else` branch (`?atproto` on a non-singular archive/search/404) and the non-`WP_Post` guard both `return`ed with `REST_REQUEST` left defined for the rest of the page render. Fixed in f4c3af8 by moving the marking into a `mark_rest_request()` helper that's only called on the two branches that immediately `exit`, so it can never outlive the preview request. Thanks!

includes/transformer/class-preview.php
```diff
@@ -0,0 +1,264 @@
...
+			self::mark_rest_request();
+			self::send( self::for_site( $type ) );
+			exit;
+		}
+
+		if ( ! \is_singular() ) {
+			return;
+		}
+
+		$post = \get_queried_object();
```
kraftbj, line 69: This authorizes with the generic `edit_posts` capability, then serves *this specific post's* records — including its saved custom Bluesky text and any third-party preview-transformer output — with no object-level `edit_post` check. The REST pre-publish controller gates on `edit_post( $post->ID )`; this front-end path doesn't. On a multi-author site, anyone who can edit their own posts can read the byte-exact record (and custom text) that would publish for another author's post. This is separate from the earlier public-vs-gated discussion: the post body is already public, but the custom text and exact record aren't otherwise exposed to that role. Suggest `current_user_can( 'edit_post', $post->ID )` for post-scoped previews (and a site-level cap like `manage_options` for the front-page publication / filter previews), or a comment noting `edit_posts` is the intended floor.

includes/transformer/class-post.php
```diff
@@ -320,6 +320,29 @@ public function project(): array {
...
+	 */
+	public function get_preview_records(): array {
+		$this->projecting = true;
+
+		try {
+			if ( $this->is_short_form_post() ) {
+				return array( $this->transform() );
+			}
+
+			return $this->build_long_form_records();
```
kraftbj, line 345: The long-form preview builds the bsky records directly, but the publish path first transforms the document, computes its CID, and calls `set_document_strong_ref()` before composing the post — so the published record's `associatedRefs` carries the document strongRef. Because the preview skips that orchestration, this emits `associatedRefs` without the document ref for a never-published post, and with the stored (potentially stale) meta CID for a post whose document changed since last publish. So the typed `app.bsky.feed.post` / `all` preview won't match what Publisher actually writes for long-form posts — the case the projection is most useful for. Worth deriving the document CID from the document preview record and injecting it (at least for `=all`), or documenting the limitation.

kraftbj (commented):
Two additional items from a review pass, both inline below. Neither has been raised in the existing threads.

includes/transformer/class-preview.php
pfefferle: Good point — fixed in 1d0772e. `render()` now authorizes per context via `Preview::current_user_can_preview()`: post previews require object-level `edit_post` on the queried post, the same gate the pre-publish REST controller uses, which closes the cross-author read of custom Bluesky text and third-party preview output. The front page keeps the `edit_posts` floor — there's no post to scope to, and the publication record is built from public site data — and the docblock now states that floor is intentional. Covered by cross-author / own-post / editor / logged-out tests.

includes/transformer/class-post.php
pfefferle: You're right on both counts — addressed in e57c063 and 9c3eaab.

**Published posts:** the preview no longer surfaces the stale `Document::META_CID`. `get_preview_records()` now mirrors the Publisher's precompute — the CID is derived from the projected document record at the reserved rkey — so the typed `app.bsky.feed.post` / `all` preview matches what `Publisher` writes on the next update.

**Never-published posts:** the ref is omitted (and the limitation documented) rather than fabricated. The document rkey only exists once the first publish reserves it, and the preview must stay read-only — `META_TID` is the publish-state marker `Publisher::update_post()` keys off. I considered minting a placeholder TID, but a URI that can never match the eventual record seemed worse than an honest omission.

While making the injection strictly read-only I also found that `Document` / `Publication` previews could upload an uncached cover image / site icon blob to the PDS as a side effect of the preview GET — 9c3eaab fixes both to project from the cached blob ref (or omit the image), so the previewed record and its CID still match a real publish whenever the blob is already on the PDS.

jeherve (approved):

PR #171, opened 2026-06-26
Harden AT Protocol HTTP redirects

pfefferle:
Fixes #

## Proposed changes:
* Disable redirects for AT Protocol OAuth PAR, token exchange, refresh, and revocation requests.
* Disable redirects for authenticated PDS API requests, including caller-provided request arguments.
* Treat non-2xx PDS and OAuth responses as failures now that redirects are not followed.
* Add regression tests for redirect handling across OAuth and PDS request paths.

### Other information:

## Testing instructions:
* Run `composer lint`.
* Run `npm run env-test`.

### Changelog entry

A changelog entry is included in `.github/changelog/harden-atproto-http-redirects`.

### DIFF
--- a/includes/oauth/class-client.php
+++ b/includes/oauth/class-client.php
@@ -16,6 +16,7 @@ use Atmosphere\Atmosphere;
 use function Atmosphere\clear_scheduled_hooks;
 use function Atmosphere\debug_log;
 use function Atmosphere\get_connection;
+use function Atmosphere\is_success_status;
 
 /**
  * OAuth client that manages the authorization lifecycle.
@@ -341,9 +342,9 @@ class Client {
 			}
 		}
 
-		if ( 2 !== \intdiv( (int) $status, 100 ) || empty( $data['request_uri'] ) ) {
+		if ( ! is_success_status( $status ) || empty( $data['request_uri'] ) ) {
 			$msg = $data['error_description'] ?? ( $data['error'] ?? \__( 'PAR request failed.', 'atmosphere' ) );
-			return new \WP_Error( 'atmosphere_par', $msg );
+			return new \WP_Error( 'atmosphere_par', $msg, array( 'status' => $status ) );
 		}
 
 		$params = array(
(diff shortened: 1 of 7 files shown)


includes/oauth/class-client.php
```diff
@@ -339,7 +341,7 @@ private static function authorize_via_par(
 			}
 		}
 
-		if ( $status >= 400 || empty( $data['request_uri'] ) ) {
+		if ( 2 !== \intdiv( (int) $status, 100 ) || empty( $data['request_uri'] ) ) {
```
jeherve: TIL about `intdiv`!

I think I've typically done `$status < 200 || $status >= 300` for things like that.

jeherve (commented):
This looks good to me.

I do have a question about error handling. When we stop following redirects, a 3xx now will now give us an empty body, and the message will be the "Token exchange failed" or "PAR request failed" fallback. 

In `class-api.php` we keep the status in the `WP_Error` data, but in `class-client.php` (`atmosphere_par`, `atmosphere_token`, `atmosphere_refresh`) we drop it.

Maybe we could pass `array( 'status' => $status )` along in those `WP_Error`s in there too? That way, if an auth server ever does redirect one of these endpoints, whoever's debugging can actually tell it apart from a 500 or a timeout.

What do you think?

includes/oauth/class-client.php
pfefferle: I hated `$status < 200 || $status >= 300`, so I searched for something shorter, but I am fine with reverting it :)

pfefferle:
Good call — done in `474b5df`. The `atmosphere_par`, `atmosphere_token`, and `atmosphere_refresh` errors now carry `array( 'status' => $status )` too, matching `atmosphere_pds`, so a 3xx (now treated as a failure since we no longer follow redirects) is distinguishable from a 500 or a timeout when debugging. Revocation already logs the status, so I left it as-is.

Also took your other suggestion and swapped the `intdiv` trick for a shared `is_success_status()` helper.

includes/oauth/class-client.php
jeherve: I'm okay with `intdiv`; I just didn't know that function existed until today :)

`$status < 200 || $status >= 300` is more readable to me, but I think that's just because I'm not used to seeing `intdiv`. 

Feel free to have your helper use `intdiv` if you prefer, I don't have any strong feelings there.

jeherve (approved):

PR #182, opened 2026-07-04
Add retry with backoff for failed publish and update cron workers

### LINKED ISSUE #179: Retry/backoff for failed post-publish cron events
## Summary

Publishing a post to Bluesky is deferred to a single one-shot WP-Cron event with **no retry**. A transient failure (PDS 429/5xx, network timeout, a token-refresh blip) fails that one attempt, and the post silently never reaches Bluesky until the post happens to be saved again.

## Details

`Atmosphere::on_status_change()` schedules the publish:

`includes/class-atmosphere.php:694`

```php
\wp_schedule_single_event( \time(), 'atmosphere_publish_post', array( $post->ID ) );
```

The cron worker runs the publish once and, on error, only logs it:

`includes/class-atmosphere.php:1455-1477` → `Publisher::publish_post()` → `self::log_cron_error(...)`

`API::request()` retries once on a DPoP nonce error and once on token refresh, but a `429`, `5xx`, or timeout (30s request / 60s blob upload) becomes a `WP_Error` with no backoff and no re-queue (`includes/class-api.php`). `wp_schedule_single_event` does not retry — acknowledged in-code at `class-atmosphere.php:1816`.

Because image posts perform an extra blocking `uploadBlob` call per image before the atomic `applyWrites`, they run longer and are statistically more likely to hit a transient failure — matching real reports where the failing post was an image post.

## Proposal

Add retry/backoff (or a durable queue) for failed `atmosphere_publish_post` / `atmosphere_update_post` events, mirroring what's proposed for the delete and comment paths.

## Related

* [#46](<https://github.com/Automattic/wordpress-atmosphere/issues/46>) (retry/backoff or durable queue for outbound-comment cron failures)
* Automattic/wordpress-atmosphere#141 (auto-retry a failed un-share)

### LINKED ISSUE #181: Disconnect/uninstall cron cleanup misses events scheduled with args
## Summary

`clear_scheduled_hooks()` / `clear_scheduled_hooks_all()` (`includes/functions.php`) call `wp_clear_scheduled_hook( $hook )` with no args. WordPress core's `wp_clear_scheduled_hook()` only unschedules events whose args match the given array — with the default empty array, **only argless events are removed** (core keys the lookup on `md5( serialize( $args ) )`, see `wp-includes/cron.php`).

Most of the plugin's queued events carry args (`atmosphere_publish_post [ $post_id ]`, `atmosphere_delete_records [ $tids, ... ]`, `atmosphere_revoke_refresh_token [ $ciphertext, $endpoint ]`), so disconnect/deactivate/uninstall left them queued — despite `get_cron_hooks()`'s own docblock stating the cleanup exists to prevent "queued events from a previous install/connection from firing against the current one and (worst case) issuing applyWrites against a different repo."

## Impact

* A per-post publish/update/delete event queued just before disconnect fires after a reconnect — potentially against a different DID/repo.
* The queued revoke event (with encrypted token ciphertext in its args) survives deactivation/uninstall in `wp_options['cron']`, which is exactly the orphaned-ciphertext scenario the code comments try to avoid.
* Retry/backoff (Automattic/wordpress-atmosphere#179) widens the stale-event window to ~21 minutes.

## Fix

Use `wp_unschedule_hook( $hook )` (clears all events for a hook regardless of args) in both helpers. Regression tests cover an arg-carrying publish/update event and the arg-carrying revoke event.

pfefferle:
Fixes #179
Fixes #181

## Proposed changes:

* **Retry with backoff for failed publish/update cron workers** (#179). Publishing is deferred to one-shot `wp_schedule_single_event` cron events; a single transient PDS failure (429/5xx/timeout) or missed cron tick previously dropped the post silently — the post showed "Published" in WordPress but never reached Bluesky (reported on wp.org as "inconsistent" publishing). Failed `atmosphere_publish_post` / `atmosphere_update_post` workers now re-queue themselves with an escalating 60s/5m/15m ladder, tracked in `_atmosphere_publish_retries` post meta and cleared on success, permanent rejection, exhaustion, or a fresh save. Mirrors the existing comment parent-defer pattern and the ActivityPub dispatcher's filterable-retry convention.
* **Transient-vs-permanent classification.** Retry-by-default with a permanent blocklist: local preconditions (`not_connected`, `needs_reauth`, `decrypt`, `did_mismatch`, …) and deterministic PDS 4xx are never retried; no-status network failures, 408, 429, and 5xx are. `atmosphere_thread_rollback_failed` is explicitly permanent — retrying after a failed rollback would mint fresh TIDs next to live orphan records and publish a duplicate copy.
* **New `atmosphere_publish_retry_delays` filter.** One entry per retry, so the ladder's length is also the attempt budget — an empty array disables retries, a longer array raises the cap. Documented in the hook reference.
* **Fix: disconnect/uninstall cron cleanup missed arg-carrying events** (#181). `wp_clear_scheduled_hook( $hook )` with no args only removes events scheduled with an empty args array, so every queued per-post event and the revoke event (encrypted token payload in its args) survived `clear_scheduled_hooks()` — the exact events the cleanup exists to stop from firing against a different connection. Both helpers now use `wp_unschedule_hook()`.

### Other information:

Reviewed pre-PR by three parallel review passes (correctness/WP-Cron races, conventions/test hygiene, edge cases/abuse); all blocking findings addressed in the last commit (duplicate-post guard, stranded-counter reset, uninstall sweep).

## Testing instructions:

* Start the environment: `npm run env-start`, connect a Bluesky account, and enable auto-publish.
* Simulate a transient PDS failure: `add_filter( 'atmosphere_pre_apply_writes', fn() => new \WP_Error( 'atmosphere_pds', 'boom', array( 'status' => 500 ) ) );` (e.g. via a mu-plugin), then publish a post.
* Check `wp cron event list` (or WP Crontrol): an `atmosphere_publish_post` event for the post is queued ~60s out; after it fails again, ~5m; then ~15m; then it stops. The post's `_atmosphere_publish_retries` meta tracks the attempt count and is removed at the end.
* Remove the failure filter before the next retry fires: the post publishes and the counter is cleared.
* Repeat with `array( 'status' => 400 )`: no retry is scheduled.
* For #181: publish a post so a per-post event is queued (or schedule one manually), then disconnect the account on the settings page and confirm `wp cron event list` shows no remaining `atmosphere_*` events.
* Run `npm run env-test` — 848 tests pass, including 10 new ones covering the ladder, the filter, the rollback guard, the counter reset, and the arg-carrying cleanup.

### DIFF
--- a/.agents/skills/code-style/SKILL.md
+++ b/.agents/skills/code-style/SKILL.md
@@ -60,7 +60,7 @@ Always reserve the rkey via meta in `get_rkey()` — that meta key is the marker
 
 **Content / composition:** `atmosphere_content_parser` (deprecated; use `Content_Parser\Registry::register()`), `atmosphere_document_content`, `atmosphere_long_form_composition`, `atmosphere_teaser_thread_posts`.
 
-**Gating:** `atmosphere_syncable_post_types`, `atmosphere_should_publish_comment`, `atmosphere_should_sync_reply`, `atmosphere_backfill_query_chunk_size`, `atmosphere_oauth_redirect_uri`, `atmosphere_client_metadata`.
+**Gating:** `atmosphere_syncable_post_types`, `atmosphere_should_publish_comment`, `atmosphere_should_sync_reply`, `atmosphere_backfill_query_chunk_size`, `atmosphere_oauth_redirect_uri`, `atmosphere_client_metadata`, `atmosphere_publish_retry_delays` (backoff ladder for failed publish/update cron workers; length = retry budget; empty array disables retries).
 
 **Actions:** `atmosphere_publishing`, `atmosphere_publish_post_result`, `atmosphere_publish_comment_result`, `atmosphere_update_skipped_unsynced_post`, `atmosphere_long_form_strategy_downgraded`, `atmosphere_reaction_synced`. `atmosphere_publishing` receives the current `WP_Post` and is not a request-wide guard.
 
--- a/.github/changelog/add-publish-retry-backoff
+++ b/.github/changelog/add-publish-retry-backoff
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Failed attempts to share a post to Bluesky are now retried automatically for about twenty minutes, so a brief network or server hiccup no longer means the post silently never appears.
--- a/.github/changelog/fix-unschedule-all-cron-events
+++ b/.github/changelog/fix-unschedule-all-cron-events
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Disconnecting or deactivating now reliably removes all pending background tasks, so a task queued under a previous connection can no longer run against a newly connected account.
--- a/docs/php-coding-standards.md
+++ b/docs/php-coding-standards.md
@@ -227,6 +227,7 @@ use function Atmosphere\is_connected;
 \apply_filters( 'atmosphere_should_publish_comment',      $bool, $comment );
 \apply_filters( 'atmosphere_should_sync_reply',           $bool, $notification, $post_id );
 \apply_filters( 'atmosphere_backfill_query_chunk_size',   500 );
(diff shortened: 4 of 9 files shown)


jeherve (approved):
This looks good to me! 🚢

PR #183, opened 2026-07-04
Surface per-post publish failures in the block editor

### LINKED ISSUE #180: Surface per-post publish failures (currently silent, log-only)
## Summary

When a post fails to publish to Bluesky, there is no user-visible signal. The failure is written only to `debug_log()`, which is gated behind `WP_DEBUG` and off by default in production. The post shows as "Published" in WordPress while nothing reached Bluesky, and the user has no way to tell which posts failed or why.

## Details

- Cron publish failures route through `Atmosphere::log_cron_error()` → `debug_log()`:
  - `includes/class-atmosphere.php:1824-1846`
  - `includes/functions.php:622-652` (gated on `WP_DEBUG` via the `atmosphere_debug_log` filter)
- No per-post error meta is written for the common single-record failure (only the narrow thread-rollback case writes `Post::META_ORPHAN_RECORDS`).
- The only user-facing failure surface today is the global `needs_reauth` admin banner, which does not fire for per-post publish failures.

## Proposal

Give operators a way to see that a specific post failed to publish and why, e.g.:

- Persist a "last publish error" per post (code + message + timestamp).
- Surface it in the editor sidebar panel and/or Site Health, alongside the orphan-records surfacing in #44.

## Related

- #44 (surface `META_ORPHAN_RECORDS` in admin / Site Health)

pfefferle:
Fixes #180

**Stacked on #182** — this branch builds on `add/publish-retry-backoff` (the diff here shows only the error-surfacing increment; rebase/retarget to `trunk` after #182 merges).

## Proposed changes:

* **Persist the most recent share failure per post.** The publish/update cron workers (and the delete worker's reconcile-to-publishable branch) now write `_atmosphere_last_publish_error` post meta on failure — code, sanitized + truncated message (PDS strings are attacker-influenced and end up rendered in the editor), a `retrying` flag reflecting whether the backoff ladder has another attempt queued, and a timestamp. Cleared on the next success and on fresh status transitions, so stale notices can't outlive their failure.
* **Expose it read-only via REST.** A new `atmosphere_publish_error` field (edit context only — WP core enforces this at the schema-context and `edit_post`-cap layers) mirrors the existing `atmosphere_url` field, with a typed object schema.
* **Show it in the editor.** The ATmosphere document panel renders an error notice when the last attempt failed: "Sharing to Bluesky failed. Your site will retry automatically." while the ladder is running, or "…Update the post to try again." once it gave up, plus the server-sanitized detail message.

Previously a failed share vanished into a `WP_DEBUG`-gated log line — the post showed "Published" in WordPress with no signal that nothing reached Bluesky.

### Other information:

Reviewed pre-PR by a security-audit pass (XSS via PDS-controlled message, REST context enforcement verified against WP core, payload sensitivity — all clean) and a conventions pass; all findings addressed.

## Testing instructions:

* Start the environment (`npm run env-start`), connect a Bluesky account, run `npm run build`.
* Force a failure: `add_filter( 'atmosphere_pre_apply_writes', fn() => new \WP_Error( 'atmosphere_pds', 'boom', array( 'status' => 500 ) ) );` in a mu-plugin, publish a post, run the queued cron event (`wp cron event run atmosphere_publish_post`).
* Reload the editor: the ATmosphere sidebar panel shows the red notice with "will retry automatically".
* Change the filter to `array( 'status' => 400 )` and repeat with a new post: the notice says "Update the post to try again."
* Remove the filter and update the post; after the cron event runs, the notice is gone.
* `GET /wp/v2/posts/<id>?context=edit` as an editor shows `atmosphere_publish_error`; `context=view` (or as a subscriber) never includes it.
* `npm run env-test` — 854 tests pass (10 new); `npm run test:unit:js` and `npm run lint:js` pass.


jeherve (dismissed):
This looks good to me, it should be good to merge after a rebase. 🚢

PR #186, opened 2026-07-07
Process reactions oldest-first so nested replies thread in one run

kraftbj:
## Proposed changes:

Bluesky's `listNotifications` streams **newest-first**, and `Reaction_Sync::paginate()` processed items in that order. But a reply's parent — a post record, or an earlier reply — is always *older* than the reply. So a nested reply was reached **before** its parent existed as a local comment, and `process_reply()` drops any reply whose parent it can't resolve (neither a local post nor an already-synced comment).

The only recovery was the next run's `WATERMARK_GRACE` re-walk (10 items). That loses deeper or bursty threads: a live-tweeted thread that lands entirely between two hourly syncs has its replies dropped, and the child several items back never re-enters the grace window.

This collects each stream and then processes it **oldest-first** (`array_reverse`), so a parent is synced before any reply that targets it — a whole thread arriving in one run now resolves in that run. Dedup via `source_id` keeps reprocessing idempotent, and the watermark is still the newest URI regardless of processing order.

Note: a child of a **spam/trash** parent still won't thread, since `find_comment_by_source_id()`'s `get_comments()` defaults to `status = 'all'` (approved + hold only). A **pending** (held) parent *does* resolve. That's intended — don't nest under spam.

### Other information:

- Updated the four `paginate()` tests to expect oldest-first processing (the collected set and watermark are unchanged; only order flips).
- Added `test_nested_reply_arriving_before_parent_resolves_in_one_run`: a child reply ahead of its parent in the stream. It **fails before this change** (child dropped) and passes after (child threads under the parent in a single run).

## Testing instructions:

* From a Bluesky account, reply to one of your published posts, then reply again to your own first reply — so a two-level thread exists.
* Ensure both arrive between sync runs (delete `atmosphere_last_seen_notification` to force a fresh walk).
* Run `wp cron event run atmosphere_sync_reactions`.
* Before this patch, the nested reply is missing (only the first-level reply syncs); after, both sync and the nested one is threaded under its parent in the same run.
* Automated: `vendor/bin/phpunit --filter 'paginate|nested_reply'`.

### Changelog entry

A changelog entry is already committed in this branch (`.github/changelog/fix-reaction-sync-oldest-first`, Patch / Fixed), so the automatic-entry box is intentionally left unchecked.

### DIFF
--- a/.github/changelog/fix-reaction-sync-oldest-first
+++ b/.github/changelog/fix-reaction-sync-oldest-first
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Reliably import nested Bluesky replies. Reactions are now processed oldest-first within each sync, so a reply threads under its parent comment in the same run instead of being dropped when a whole thread arrives between syncs.
--- a/includes/class-reaction-sync.php
+++ b/includes/class-reaction-sync.php
@@ -349,6 +349,7 @@ class Reaction_Sync {
 		$cursor    = null;
 		$pages     = 0;
 		$rewalk    = null;
+		$collected = array();
 
 		while ( $pages < self::MAX_PAGES ) {
 			$response = $fetch( $cursor );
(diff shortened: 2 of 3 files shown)


pfefferle (approved):
👍

jeherve (approved):

PR #189, opened 2026-07-10
Docs: document the programmatic publishing surface for integrating plugins

jeffpaul:
## What

Adds a **Publishing Programmatically** section to `docs/developer-docs.md` (plus ToC entry and a changelog file), documenting the integration surface a companion plugin needs to drive cross-posting per post:

- The registered per-post metas — `atmosphere_disabled` (opt-out) and `atmosphere_custom_text` (per-post Bluesky text)
- Taking over the save-flow: `atmosphere_auto_publish` option + `\Atmosphere\Publisher::publish_post()` (with its `is_post_publishable()` gate and `update_post()`/`delete_post()` lifecycle)
- Reading back the created record via `_atmosphere_bsky_uri` / `_atmosphere_bsky_tid`
- Pointers to the existing `atmosphere_publish_post_result` / `atmosphere_reaction_synced` hooks and the comment `protocol`/`source_url` markers for synced replies

## Why

I'm building a WordPress publishing plugin ([Moment](https://github.com/jeffpaul/moment) — a phone-first publishing prototype) that wants to delegate its Bluesky syndication to ATmosphere rather than maintain its own AT Protocol client. Everything needed for that already exists in ATmosphere — this PR just documents it as the supported integration surface, so integrators can rely on it (and so gaps get spotted if any of it isn't meant to be public API).

## Testing

Docs + changelog only, no code changes. `composer lint` passes; `npm run env-test` not run since no PHP changed.

🤖 Drafted with [Claude Code](https://claude.com/claude-code), human-reviewed.

### DIFF
--- a/.github/changelog/add-programmatic-publishing-docs
+++ b/.github/changelog/add-programmatic-publishing-docs
@@ -1,0 +1,4 @@
+Significance: patch
+Type: added
+
+Developer docs: document the programmatic publishing surface (per-post metas, Publisher::publish_post(), record read-back) for integrating plugins.
--- a/.github/changelog/add-programmatic-publishing-docs
+++ b/.github/changelog/add-programmatic-publishing-docs
@@ -1,4 +1,0 @@
-Significance: patch
-Type: added
-
-Developer docs: document the programmatic publishing surface (per-post metas, Publisher::publish_post(), record read-back) for integrating plugins.
--- a/docs/developer-docs.md
+++ b/docs/developer-docs.md
@@ -7,6 +7,7 @@
 - [Previewing AT Protocol Records](#previewing-at-protocol-records)
 - [Extending Content Formats](#extending-content-formats)
 - [Custom Post Type Support](#custom-post-type-support)
+- [Publishing Programmatically](#publishing-programmatically)
 - [Templates and Admin UI](#templates-and-admin-ui)
 
 ## Introduction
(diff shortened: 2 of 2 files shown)


pfefferle:
Thanks for this, @jeffpaul! I verified every documented symbol against the code (constants, `is_post_publishable()` checks, `publish_post()` signature/return/hook behavior, meta keys, anchors) and it all checks out. A few points, including the Copilot comments, which I checked against the code:

- **Changelog**: docs-only changes don't need a changelog entry — please drop `.github/changelog/add-programmatic-publishing-docs`; I'll add the "Skip Changelog" label.
- **`atmosphere_disabled` removal** (Copilot is right): the automatic record removal runs through the meta-change reconcile, which is skipped when `atmosphere_auto_publish` is off. A one-liner in the takeover section would close it, e.g. *"With auto-publish disabled, call `\Atmosphere\Publisher::update_post()` after changing per-post metas to reconcile the remote records"* (it routes ineligible posts to `delete_post()`).
- **Reactions paragraph** (Copilot is right here too): likes arrive as comments as well (`comment_type` `like`), and `source_url` is intentionally empty for likes *and* reposts (no bsky.app landing page; the author profile link is in `comment_author_url`). Suggested wording: *"Replies, likes, and reposts synced back from Bluesky arrive as native WordPress comments (`protocol` comment meta `atproto`; for replies, `source_url` links to the reply on Bluesky)."*
- **Namespace consistency**: fine either way, but fully-qualified names are safer for copy-pasting.

On your "gaps get spotted" point: merging this makes `Publisher::publish_post()`, the `_atmosphere_bsky_uri`/`_atmosphere_bsky_tid` metas, and the comment markers a committed integration surface — I'm comfortable with that; documenting it is exactly the right move.

pfefferle (dismissed):

pfefferle (approved):

PR #198, opened 2026-07-14
Fix: Require publishing capability for Bluesky comment replies

### LINKED ISSUE #196: Do not cross-post comments from users without publishing capabilities
## Problem

ATmosphere currently treats any approved local comment with a positive WordPress `user_id` as eligible for automatic publication to Bluesky.

Many WordPress sites allow visitors to register solely so they can comment. Those users normally have the Subscriber role and are not authorized to publish site content, but their comments are currently republished as `app.bsky.feed.post` records through the site-connected Bluesky account.

The outgoing record contains the comment text without WordPress author attribution, which can make a Subscriber comment appear to be speech from the site account.

## Steps to reproduce

1. Connect ATmosphere to a Bluesky account.
2. Cross-post a public WordPress post.
3. Enable user registration.
4. Register a Subscriber and submit an approved comment on the cross-posted post.
5. Run WP-Cron.

The Subscriber comment is published as a Bluesky reply from the site-connected account.

## Current behavior

The eligibility gate only checks whether the comment has a registered user ID:

```php
if ( (int) $comment->user_id <= 0 ) {
    return false;
}
```

A positive user ID distinguishes registered users from anonymous commenters, but it does not establish authorization to publish content.

## Expected behavior

Use WordPress capabilities for the outbound eligibility decision.

A local comment should only be eligible for Bluesky publication when its author has the `publish_posts` capability on the site:

```php
\user_can( (int) $comment->user_id, "publish_posts" )
```

This allows Authors, Editors, Administrators, and custom roles explicitly granted that capability while excluding Subscribers and other comment-only accounts.

The check must use the comment author ID rather than `current_user_can()`, because the actual publication runs asynchronously in WP-Cron without the original commenter logged in.

## Acceptance criteria

* Subscriber comments are not scheduled or published to Bluesky.
* Comments from users with `publish_posts` remain eligible when all existing checks pass.
* The cron handler rechecks the capability before writing, so removing the capability after scheduling prevents publication.
* Anonymous comments, inbound Bluesky comments, trackbacks, pingbacks, unapproved comments, and comments on unpublished/non-public posts remain excluded.
* The existing `atmosphere_should_publish_comment` filter remains available for integrations that deliberately override the default.
* Tests cover an Author, a Subscriber, and a capability removal between scheduling and cron execution.
* User-facing documentation says comments from users allowed to publish posts are cross-posted; it must no longer say every logged-in reader comment is cross-posted.

The separate global constant/option kill switch is tracked in Automattic/wordpress-atmosphere#197. Inbound reaction synchronization remains independent.

pfefferle:
Fixes #196

## Proposed changes:
* WordPress account registration currently lets Subscriber comments be republished through the site-connected Bluesky account, even though Subscribers cannot publish site content.
* Require the comment author to have the current `publish_posts` capability before scheduling or executing an outbound reply. This uses the stored author ID so the check remains valid in WP-Cron.
* Preserve the existing filter override and all other comment eligibility checks.
* Update public and developer documentation so it no longer says every logged-in reader comment is cross-posted.

### Other information:

- No visual changes; screenshots are not applicable.

## Testing instructions:

* Run `composer lint`.
* Run `npm run env-test`.
* For the focused regression coverage, run `npm run env-test -- --filter="test_(eligible_author|subscriber_comment|publish_comment_cron_rechecks_author_capability)"`.
* Confirm the focused tests show that Author comments remain eligible, Subscriber comments do not schedule an outbound publish, and a queued comment is not written after its author is downgraded to Subscriber.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

A manual changelog entry is included in `.github/changelog/fix-comment-publishing-capability`.

</details>

### DIFF
--- a/.github/changelog/fix-comment-publishing-capability
+++ b/.github/changelog/fix-comment-publishing-capability
@@ -1,0 +1,4 @@
+Significance: minor
+Type: fixed
+
+Comments from Subscribers and other comment-only accounts now stay on WordPress instead of being published through the site's Bluesky account.
--- a/docs/developer-docs.md
+++ b/docs/developer-docs.md
@@ -44,7 +44,7 @@ ATmosphere exposes a small set of filters and actions for plugins to extend beha
 | `atmosphere_publication_labels` | filter | Add standard self-labels to `site.standard.publication` records. |
 | `atmosphere_publication_show_in_discover` | filter | Override `preferences.showInDiscover` (defaults to the site's `blog_public` option) for `site.standard.publication` records. |
 | `atmosphere_syncable_post_types` | filter | Add or remove post types eligible for cross-posting. |
-| `atmosphere_should_publish_comment` | filter | Customise which approved comments are mirrored as Bluesky replies. |
+| `atmosphere_should_publish_comment` | filter | Customise which approved comments from users allowed to publish posts are mirrored as Bluesky replies. |
 | `atmosphere_should_sync_reply` | filter | Customise which inbound Bluesky replies become WordPress comments. |
 | `atmosphere_transform_bsky_post` | filter | Mutate the Bluesky post record before write. |
 | `atmosphere_transform_document` | filter | Mutate the document record before write. |
--- a/includes/class-atmosphere.php
+++ b/includes/class-atmosphere.php
@@ -1194,7 +1194,16 @@ class Atmosphere {
 			return false;
 		}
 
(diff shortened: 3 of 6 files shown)


jeherve (approved):
This looks good to me, let's merge!

PR #199, opened 2026-07-14
Add: Controls for publishing WordPress comments to Bluesky

### LINKED ISSUE #197: Add constant and setting to disable all outgoing reactions
## Summary

Add a global kill switch for outbound reaction publishing. Site owners must be able to disable it through both a deployment-level PHP constant and a WordPress option.

This is separate from the consent/default-policy discussion in Automattic/wordpress-atmosphere#196: even if outbound reactions remain available, administrators and hosting platforms need a reliable way to turn them off completely.

## Proposed controls

* A boolean `ATMOSPHERE_DISABLE_OUTGOING_REACTIONS` constant for `wp-config.php` and managed hosting environments.
* A registered boolean WordPress option, exposed on the ATmosphere settings page, that lets administrators disable outgoing reactions without editing PHP.
* The constant takes precedence. When it disables the feature, the settings control should be disabled and explain that configuration is enforced by the constant.

The exact option name can follow the existing settings conventions; for example, `atmosphere_publish_reactions` with an “Enable outgoing reactions” checkbox or `atmosphere_disable_outgoing_reactions` with disable semantics.

## Scope

When disabled, ATmosphere must not create, update, or delete Bluesky records in response to the local WordPress comment lifecycle:

* New or newly approved local comments.
* Edited local comments.
* Unapproved, trashed, or deleted local comments.
* Cron events that were queued before the switch was enabled.

Both the scheduling paths and the cron execution paths must check the centralized setting so an already-queued event cannot bypass it. Existing remote reaction records should remain unchanged while the feature is disabled.

The switch must not affect:

* Normal outbound post and [standard.site](<http://standard.site>) document publishing.
* Inbound Bluesky reply synchronization.
* Inbound like and repost synchronization.

## Acceptance criteria

* A single helper determines whether outgoing reactions are enabled.
* Defining `ATMOSPHERE_DISABLE_OUTGOING_REACTIONS` as true disables outgoing reactions regardless of the saved option.
* The saved option disables outgoing reactions when the constant does not force a value.
* The setting is registered through the Settings API and is available on the ATmosphere settings page.
* No outbound comment-reaction API write occurs while disabled, including from previously queued cron jobs.
* Tests cover the option, constant precedence, lifecycle scheduling, and queued cron execution.
* User-facing documentation explains both controls and makes clear that inbound synchronization remains independent.

pfefferle:
Fixes #197

## Proposed changes:

* Add an **Outgoing replies** setting (Settings → ATmosphere → Reactions) that keeps eligible WordPress comments, edits, and deletions from being published to Bluesky — without disabling incoming reaction sync. The inbound toggles are relabeled **Incoming reactions** / **Incoming replies** so the section reads as a clear directional split.
* Add the `atmosphere_should_publish_comments` filter as the programmatic override: it resolves after the stored setting and has the final say in both directions, without touching the saved preference. This replaces the `ATMOSPHERE_DISABLE_OUTGOING_REACTIONS` constant originally proposed in #197 — the filter matches the per-feature helper pattern established in #203 (`is_comment_publishing_enabled()`).
* Enforce the effective control in direct Publisher calls, queued cron callbacks, and the post-deletion cascades, so stale queued work cannot bypass it. A queued comment-only delete cascade becomes a silent no-op instead of a logged error.
* Preserve existing remote replies and their local metadata while comment publishing is disabled.
* Housekeeping on the settings screen: `rest_sanitize_boolean` sanitize callbacks and `aria-describedby` wiring for all four boolean toggles.

### Other information:

- The settings field uses native WordPress settings markup; no before/after screenshot is included.

## Testing instructions:

* Run `composer lint` and `npm run env-test` (full suite green).
* Go to Settings → ATmosphere → Reactions, uncheck **Publish eligible WordPress comments as Bluesky replies**, and save.
* Add, edit, and delete an otherwise eligible local comment; run due cron events and verify no outgoing Bluesky reply write is made while incoming reactions keep syncing.
* Re-enable the setting and verify eligible comments publish again.
* Add `add_filter( 'atmosphere_should_publish_comments', '__return_false' );` in a small mu-plugin: outgoing writes stop while the checkbox still shows (and edits) the saved preference; remove the filter and the saved preference applies again.

### Changelog entry

A validated manual changelog entry is included in `.github/changelog/add-outgoing-reactions-controls`.

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Add a setting — and a filter for plugins — to keep WordPress comment changes from being published to Bluesky while leaving incoming reactions enabled.

</details>


jeherve (approved):
This makes sense to me. 🚢

PR #203, opened 2026-07-15
Add Connectors API integration and a connection layer other plugins can consume

### LINKED ISSUE #190: Feature Request: extract the Bluesky connection layer into a standalone 'Connector for Bluesky' plugin (WP 7.0 Connectors API)
### What

Consider splitting ATmosphere's network-connection layer into a standalone **Connector for Bluesky** plugin that any WordPress plugin can build on, with ATmosphere consuming it as its flagship user:

- **Connector plugin**: the AT Protocol plumbing — OAuth 2.1 + PKCE + DPoP flow, PDS/handle resolution, session management, a public XRPC client surface (`applyWrites`, blob upload, record reads) — with the account connection registered through the **WordPress 7.0 Connectors API** (`wp_connectors_init`), so it lives on the core Settings → Connectors screen alongside the AI providers and Akismet.
- **ATmosphere**: everything product-shaped stays — automatic cross-posting of posts/comments, `standard.site` documents and publication records, reaction sync into comments, the editor panel and admin UI — now talking to Bluesky through the shared connector.

### Why

WP 7.0's Connectors API established the pattern of *one site-level connection, many consumers* (the `ai` plugin + `ai-provider-for-*` split is the precedent). Multiple plugins now want to talk to Bluesky for different purposes — cross-posters, publishing UIs, comment/backfeed tools. I maintain one ([Moment](https://github.com/jeffpaul/moment), a phone-first publishing prototype) and today the ecosystem reality is: every such plugin ships its own AT Protocol client and its own credential storage, users connect the same Bluesky account N times, and N half-clients drift apart on protocol details (DPoP nonces, session refresh, service auth).

ATmosphere has the best open implementation of that connection layer in the WordPress ecosystem. Spun out, it becomes shared infrastructure: users connect Bluesky once on Settings → Connectors; ATmosphere and third-party plugins all use that connection; protocol fixes land in one place.

### How

Possible shape, following the `ai`-plugin precedent:

- New plugin (e.g. `connector-for-bluesky`) providing the Connectors API registration, the OAuth connection flow, and a public client service + hooks for consumers.
- ATmosphere declares it via `Requires Plugins` (or bundles/feature-flags it during a transition).
- Consumers get a documented surface roughly equivalent to what ATmosphere's internal OAuth/XRPC classes do today.

No feature changes requested — purely packaging/layering. Related: #189 documents the current programmatic publishing surface, which is the near-term integration path for plugins like mine; this issue is the longer-term architectural version of the same need. Happy to help test from the consumer side.

🤖 Drafted with [Claude Code](https://claude.com/claude-code), human-reviewed.

jeherve:
Fixes #190

## Proposed changes:

#190 asked whether the Bluesky connection layer should be spun out into a standalone "Connector for Bluesky" plugin. This PR takes the alternative @pfefferle  suggested instead: keep the connection layer inside ATmosphere, surface it on the core Connectors screen, and give a host plugin a way to consume just the connection while hiding ATmosphere's own UI.

Four changes, from user-facing to developer-facing:

* **Handle typeahead on the connect field.** When connecting an account on Settings → ATmosphere, the handle field now suggests AT Protocol handles as you type instead of requiring you to know the exact handle. It defaults to Bluesky's public appview and is filterable (`atmosphere_handle_typeahead_url`) so you can point it at a network-wide index or disable it entirely.

Screenshot 2026-07-15 at 13 23 02

* **A connector on Settings → Connectors (WordPress 7.0+).** ATmosphere registers an `atmosphere` connector via `wp_connectors_init`, so the account connection lives on the core Connectors screen alongside the AI providers and Akismet. The card drives the existing OAuth flow; publishing and OAuth are untouched. On older WordPress the whole thing is inert (guarded by `class_exists( 'WP_Connector_Registry' )`), and the plugin's own settings page stays the baseline.

Screenshot 2026-07-15 at 13 19 05

Screenshot 2026-07-15 at 13 19 08

* **A connection-only mode for host plugins (`atmosphere_connection_only_mode`).** A plugin that embeds ATmosphere purely as its connection layer can return `true` to make ATmosphere stop acting on its own: automatic cross-posting, likes/reposts import, reply import, and publishing comments as Bluesky replies all switch off, and Settings → ATmosphere is hidden. The settings page stays *registered* (the OAuth callback lands on its own URL and core's access gate would otherwise `wp_die` before the callback runs); it's just removed from the menu, the reauth notice is suppressed, and a direct visit shows a short "managed by another plugin" notice. Each behavioural lane keeps its own filter (`atmosphere_should_auto_publish`, etc.), so a host can re-enable a single one while suppressing the rest.

* **A clearer settings link on the Plugins page.** When the settings page is hidden by connection-only mode, the plugin-row "Settings" shortcut becomes a plain "Settings hidden by another plugin" label instead of a dead link.

Screenshot 2026-07-15 at 13 22 47
Screenshot 2026-07-15 at 13 19 20

### Other information:

## Testing instructions:

**Typeahead (any WordPress):**
1. Go to **Settings → ATmosphere** while disconnected.
2. Start typing a handle (e.g. `jeher`) in the connect field — suggestions should appear; picking one fills the field.
3. Manual entry of a full handle still works, and `add_filter( 'atmosphere_handle_typeahead_url', '__return_empty_string' )` disables the suggestions.

**Connectors screen (WordPress 7.0+):**
1. Go to **Settings → Connectors**. An **ATmosphere** card should appear.
2. Connect from the card; you should complete the same OAuth flow and land back on the Connectors screen, now showing the connected state.
3. Disconnect from the card and confirm the account is cleared.

**Connection-only mode (hides the settings page):**
1. Add `add_filter( 'atmosphere_connection_only_mode', '__return_true' );` (mu-plugin or snippet).
2. **Settings → ATmosphere** disappears from the menu; the Plugins-page link reads "Settings hidden by another plugin"; the reauth notice no longer shows.
3. Automatic cross-posting, likes/reposts, and reply importing stop; connecting from the Connectors card still works and redirects correctly.
4. Visiting `options-general.php?page=atmosphere` directly shows the "managed by another plugin" notice.

### Changelog entry

Changelog entries are already committed under `.github/changelog/` (connectors integration, connection-only mode, settings handle typeahead) — no need to generate one from the box below.

- [ ] Automatically create a changelog entry from the details below.


jest-unit.config.js
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Internal dependencies
+ */
+const defaultConfig = require( '@wordpress/scripts/config/jest-unit.config.js' );
+
+module.exports = {
+	...defaultConfig,
+	setupFiles: [ '<rootDir>/jest.setup.js' ],
+};
```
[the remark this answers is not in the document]
jeherve: Good catch — fixed in 5a94f4a. `setupFiles` now spreads `defaultConfig.setupFiles` (falling back to `[]`) before appending our `jest.setup.js`, so any setup files the wp-scripts config declares now or in the future keep running. Updated the header comment to match. (The default config currently declares none, so nothing was actively broken, but the merge future-proofs it.)

webpack.config.js
```diff
@@ -0,0 +1,54 @@
...
+
+	return {
+		...config,
+		plugins: config.plugins.map( ( plugin ) => {
+			if (
+				plugin.constructor.name !==
+				'DependencyExtractionWebpackPlugin'
+			) {
+				return plugin;
+			}
```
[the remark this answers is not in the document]
jeherve: Agreed — fixed in 5a94f4a. Now matches via `plugin instanceof DependencyExtractionWebpackPlugin` (the class is already imported at the top of the file), which is minification-safe and expresses the intent directly. Verified the module build still emits `external "@wordpress/connectors"`.

pfefferle:
What about a more generic construct that does not only hide the settings, but also configures the plugin, to not auto cross post or handle incoming reactions!?

Or do we need the granularity?

pfefferle:
Replace the `fediverse` tag with `connector` in the README so the plugin appears in the connector tag search linked from the Connectors page footer: `/wp-admin/plugin-install.php?s=connector&tab=search&type=tag`.

jeherve:
> What about a more generic construct that does not only hide the settings, but also configures the plugin, to not auto cross post or handle incoming reactions!?
> 
> Or do we need the granularity?

That's a good call. I've implemented it, but will wait until #199 is merged and rebase my branch on top of it to leverage it.

> Replace the `fediverse` tag with `connector` in the README so the plugin appears in the connector tag search linked from the Connectors page footer: `/wp-admin/plugin-install.php?s=connector&tab=search&type=tag`.

Done in e7efeaaa9515e8a5c9e06daa63226e673bbb450c

src/connectors-card/index.js
```diff
@@ -0,0 +1,331 @@
...
+		} else {
+			actionArea = el(
+				Button,
+				{
+					variant: 'secondary',
+					size: 'compact',
+					onClick: () => connect(),
+					isBusy: busy,
+					disabled: busy || ! handle,
+					accessibleWhenDisabled: true,
```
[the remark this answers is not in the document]
jeherve: Good catch — fixed in 3a56d4a. The Connect button now disables on `busy || ! handle.trim()`, matching `connect()`'s own trim precondition, so a whitespace-only handle no longer leaves it clickable-but-inert. Rebuilt the card module bundle.

docs/developer-docs.md
```diff
@@ -45,7 +45,11 @@ ATmosphere exposes a small set of filters and actions for plugins to extend beha
...
 | `atmosphere_syncable_post_types` | filter | Add or remove post types eligible for cross-posting. |
+| `atmosphere_connection_only_mode` | filter | Return `true` to embed ATmosphere purely as a connection layer: auto cross-posting, reaction/reply import, comment publishing, and the settings screen all default off. |
+| `atmosphere_show_settings_page` | filter | Return `false` to hide the Settings → ATmosphere screen when another plugin drives the connection from Settings → Connectors. |
+| `atmosphere_should_auto_publish` | filter | Effective on/off for automatic post cross-posting; runs after the stored setting and connection-only mode, and has the final say. |
 | `atmosphere_should_publish_comment` | filter | Customise which approved comments from users allowed to publish posts are mirrored as Bluesky replies. |
+| `atmosphere_should_sync_reactions` | filter | Effective on/off for importing Bluesky likes and reposts; runs after the stored setting and connection-only mode. |
 | `atmosphere_should_sync_reply` | filter | Customise which inbound Bluesky replies become WordPress comments. |
```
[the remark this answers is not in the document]
jeherve: Fixed in 3a56d4a. Added `atmosphere_should_sync_replies` (and its sibling `atmosphere_should_publish_comments`) to the developer-docs hooks table, each flagged as the effective on/off toggle that runs after the stored setting and connection-only mode — the lever a connection-only-mode consumer uses to re-enable a single import/publish lane — and noting it is distinct from the per-reply `_reply` / per-comment `_comment` filter. Also completed the full catalog in `php-coding-standards.md`, which was missing `atmosphere_connection_only_mode`, `atmosphere_should_auto_publish`, `atmosphere_should_sync_reactions`, and `atmosphere_should_sync_replies`.

pfefferle:
**The connector card never appears on WordPress.com (Atomic), despite registering fine** — found while testing the branch on a live .com site.

**Symptom:** `WP_Connector_Registry::get_instance()->get_all_registered()` includes `atmosphere`, `build/connectors-card/index.asset.php` is present, WP is 7.0.1 — but Settings → Connectors shows no ATmosphere card and no console errors.

**Root cause:** on .com the Connectors screen isn't served as core's `options-connectors.php` — nav unification re-registers it as a settings submenu (`options-general.php?page=options-connectors-wp-admin`), so `admin_enqueue_scripts` fires with the hook suffix `settings_page_options-connectors-wp-admin` (confirmed via the page's body class). `Connectors::maybe_enqueue()` compares against `self::SCREEN` (`'options-connectors.php'`) exactly and silently bails, and since the connector registers with `method => 'none'`, no module = no card at all — registered but invisible.

**Fix suggestion** — recognize re-registered variants of the screen:

```php
$is_connectors_screen = self::SCREEN === $hook_suffix
	|| \str_contains( $hook_suffix, 'options-connectors' );

if ( ! $is_connectors_screen ) {
	return;
}
```

That covers stock core, .com's `settings_page_options-connectors-wp-admin`, and any similar `{parent}_page_options-connectors-*` remapping, while staying scoped enough not to enqueue elsewhere. Happy to push this to the branch if you prefer.

Verified evidence from the .com site: `bodyClass` containing `settings_page_options-connectors-wp-admin`, `importmapHasAtmosphere: false`, `atmosphereModuleTags: []`, registry listing `atmosphere` at index 5.

src/shared/handle-typeahead.js
```diff
@@ -0,0 +1,240 @@
...
+ * @return {Promise<Array>} Matching actors ({ did, handle, displayName, avatar }).
+ */
+export async function searchHandles( baseUrl, query ) {
+	try {
+		const url = `${ baseUrl }?q=${ encodeURIComponent(
+			query
+		) }&limit=${ TYPEAHEAD_LIMIT }`;
+		const res = await fetch( url, {
+			headers: { Accept: 'application/json' },
+		} );
```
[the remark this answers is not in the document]
jeherve: Fixed in 5e03b6a. `searchHandles()` now builds the request with the `URL` API (`url.searchParams.set('q', ...)` / `set('limit', ...)`), so an endpoint that already carries a query string (or a trailing `?`) no longer produces a double `?` / malformed query, and encoding stays consistent. Added regression tests for both the bare-endpoint and existing-query-string cases.

src/shared/handle-typeahead.js
```diff
@@ -0,0 +1,240 @@
...
+	};
+
+	const onKeyDown = ( e ) => {
+		if ( 'ArrowDown' === e.key ) {
+			e.preventDefault();
+			setOpen( suggestions.length > 0 );
+			setActive( ( i ) => Math.min( i + 1, suggestions.length - 1 ) );
+		} else if ( 'ArrowUp' === e.key ) {
+			e.preventDefault();
+			setActive( ( i ) => Math.max( i - 1, 0 ) );
```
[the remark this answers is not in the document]
jeherve: Good catch — fixed in 5e03b6a. The ArrowUp handler now holds `active` at -1 while the list is empty (`suggestions.length ? Math.max( i - 1, 0 ) : -1`) so `aria-activedescendant` never points at a non-existent option id, and it toggles `open` the same way ArrowDown does.

jeherve:
> **The connector card never appears on WordPress.com (Atomic), despite registering fine** — found while testing the branch on a live .com site.

That's a good one, thank you. Turns out this happens on any site using the Gutenberg plugin; it registers its own route-based submenu, removing Core's entry to avoid duplication. I've updated this PR to support both.

includes/wp-admin/class-admin.php
```diff
@@ -45,6 +51,27 @@ public static function register(): void {
...
+	 * connection through the Settings → Connectors screen and its own UI — has no
+	 * use for Settings → ATmosphere. The connection layer, publishing, and the
+	 * Connectors card all keep working; only this plugin's own settings surface is
+	 * hidden.
+	 *
+	 * @return bool True to show the settings screen (default), false to hide it.
+	 */
+	public static function is_settings_page_visible(): bool {
+		return ! is_connection_only_mode();
 	}
```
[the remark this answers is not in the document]
jeherve: Good catch on the inconsistency — but it's the PR description that was stale, not the code. The `atmosphere_show_settings_page` filter was intentionally removed earlier in this branch and folded into `atmosphere_connection_only_mode`, so settings-page visibility now follows connection-only mode directly (`! is_connection_only_mode()`). This is deliberate: it prevents the footgun of hiding the UI while the plugin keeps cross-posting and importing reactions in the background. A host that wants ATmosphere as a pure connection layer reaches for connection-only mode, which turns off those behaviours *and* hides the settings page in one switch.

I've updated the PR description to drop the `atmosphere_show_settings_page` references and document `atmosphere_connection_only_mode` instead. No code change needed here.

includes/rest/admin/class-connection-controller.php
```diff
@@ -0,0 +1,184 @@
...
+					'permission_callback' => array( $this, 'check_permission' ),
+					'show_in_index'       => false,
+					'args'                => array(
+						'handle' => array(
+							'description'       => \__( 'The AT Protocol handle to connect, e.g. alice.bsky.social.', 'atmosphere' ),
+							'type'              => 'string',
+							'required'          => true,
+							'sanitize_callback' => 'sanitize_text_field',
+						),
+					),
```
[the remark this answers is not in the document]
jeherve: Fixed in 9d06247 — the authorize route's `handle` arg now uses `'\sanitize_text_field'` to match the project convention for global functions in namespaced code.

(For reference: string callables like this always resolve in the global namespace when WordPress invokes them, so there was no actual resolution risk — but the backslash keeps it consistent with the AGENTS.md rule. The pre-existing `sanitize_callback` strings in `pre-publish-controller`/`reactions-controller` still use bare strings; normalizing those is a separate, out-of-scope cleanup.)

includes/functions.php
```diff
@@ -718,6 +756,152 @@ function is_sharing_enabled( \WP_Post $post ): bool {
...
+ * it off), forced off in {@see is_connection_only_mode()}, and finally the
+ * `atmosphere_should_auto_publish` filter, which has the last word so a host in
+ * connection-only mode can re-enable cross-posting on its own terms.
+ *
+ * @since unreleased
+ *
+ * @return bool
+ */
+function is_auto_publish_enabled(): bool {
+	$enabled = '1' === \get_option( 'atmosphere_auto_publish', '1' );
```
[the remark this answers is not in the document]
jeherve: Fixed in 48536b2 — cast `get_option()` to string before the `'1' ===` comparison, matching what `is_comment_publishing_enabled()` already does. Applied the same fix to all three gates (auto-publish, reaction sync, reply sync) and added a regression test that stores each option as the integer `1` and asserts the feature still reads as enabled (verified it fails without the cast).

includes/functions.php
```diff
@@ -718,6 +756,152 @@ function is_sharing_enabled( \WP_Post $post ): bool {
...
+ * `atmosphere_sync_reactions` option, forced off in
+ * {@see is_connection_only_mode()}, then the `atmosphere_should_sync_reactions`
+ * filter as the final say.
+ *
+ * @since unreleased
+ *
+ * @return bool
+ */
+function is_reaction_sync_enabled(): bool {
+	$enabled = '1' === \get_option( 'atmosphere_sync_reactions', '1' );
```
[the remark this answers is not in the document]
jeherve: Fixed in 48536b2 — cast `get_option()` to string before the `'1' ===` comparison, matching what `is_comment_publishing_enabled()` already does. Applied the same fix to all three gates (auto-publish, reaction sync, reply sync) and added a regression test that stores each option as the integer `1` and asserts the feature still reads as enabled (verified it fails without the cast).

includes/functions.php
```diff
@@ -718,6 +756,152 @@ function is_sharing_enabled( \WP_Post $post ): bool {
...
+ * `atmosphere_sync_replies` option, forced off in
+ * {@see is_connection_only_mode()}, then the `atmosphere_should_sync_replies`
+ * filter as the final say.
+ *
+ * @since unreleased
+ *
+ * @return bool
+ */
+function is_reply_sync_enabled(): bool {
+	$enabled = '1' === \get_option( 'atmosphere_sync_replies', '1' );
```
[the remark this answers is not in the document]
jeherve: Fixed in 48536b2 — cast `get_option()` to string before the `'1' ===` comparison, matching what `is_comment_publishing_enabled()` already does. Applied the same fix to all three gates (auto-publish, reaction sync, reply sync) and added a regression test that stores each option as the integer `1` and asserts the feature still reads as enabled (verified it fails without the cast).

pfefferle:
@jeherve there is a merge conflict!

jeherve:
Should be solved now!

includes/wp-admin/class-admin.php
```diff
@@ -74,6 +117,36 @@ public static function add_menu(): void {
...
+	 * so the copy is intentionally generic.
+	 *
+	 * @param string[] $links Existing plugin action links.
+	 * @return string[] Filtered plugin action links.
+	 */
+	public static function filter_action_links( array $links ): array {
+		if ( self::is_settings_page_visible() ) {
+			$settings_link = \sprintf(
+				'<a href="%s">%s</a>',
+				\esc_url( \admin_url( 'options-general.php?page=atmosphere' ) ),
```
pfefferle: `settings_url()` is already imported (line 27) and used elsewhere in this file:

```suggestion
				\esc_url( settings_url() ),
```

includes/class-connectors.php
```diff
@@ -0,0 +1,305 @@
...
+	 * `{parent}_page_options-connectors-*` variant). Match both by looking for the
+	 * shared {@see self::SCREEN_SLUG} marker, which stays scoped to this one screen so
+	 * the card never enqueues elsewhere.
+	 *
+	 * @param string $hook_suffix Current admin page hook suffix.
+	 * @return bool True on the Connectors screen in any of its registered forms.
+	 */
+	public static function is_connectors_screen( string $hook_suffix ): bool {
+		return self::SCREEN === $hook_suffix
+			|| \str_contains( $hook_suffix, self::SCREEN_SLUG );
```
pfefferle: The exact-match clause is subsumed: `'options-connectors.php'` contains `'options-connectors'`, so the `str_contains()` already matches core's hook suffix on its own.

```suggestion
		return \str_contains( $hook_suffix, self::SCREEN_SLUG );
```

includes/class-connectors.php
```diff
@@ -0,0 +1,305 @@
...
+							'type' => 'profile',
+							'did'  => $did,
+						)
+					)
+					: '',
+				'restRoot'       => \esc_url_raw( \rest_url() ),
+				'restNonce'      => \wp_create_nonce( 'wp_rest' ),
+				'authorizePath'  => $base . '/authorize',
+				'disconnectPath' => $base . '/disconnect',
+				'typeaheadUrl'   => self::typeahead_url(),
```
pfefferle: `typeahead_url()` is a new wrapper whose only caller is this line — the "existing callers" it is kept for don't exist yet. The payload can call the helper directly:

```suggestion
				'typeaheadUrl'   => handle_typeahead_url(),
```

With this applied, the wrapper below and its two dedicated tests can go too (suggestions follow).

includes/class-connectors.php
```diff
@@ -0,0 +1,305 @@
...
+	 * The handle typeahead endpoint the card queries as the user types.
+	 *
+	 * Thin wrapper over {@see handle_typeahead_url()}, kept so the card's
+	 * hydration payload and existing callers have a stable entry point.
+	 *
+	 * @return string The typeahead XRPC endpoint, or '' to disable typeahead.
+	 */
+	public static function typeahead_url(): string {
+		return handle_typeahead_url();
+	}
```
pfefferle: Pairs with the suggestion above — with the payload calling `handle_typeahead_url()` directly, the wrapper has no remaining callers, and the filter behavior stays covered by the `handle_typeahead_url()` tests in `class-test-functions.php`:

```suggestion
	}
```

includes/functions.php
```diff
@@ -156,6 +156,39 @@ function appview_base_url( string $base ): string {
...
+	 *
+	 * @param string $url Default typeahead XRPC endpoint. Return '' to disable
+	 *                    typeahead and require manual handle entry.
+	 */
+	$url = (string) \apply_filters(
+		'atmosphere_handle_typeahead_url',
+		'https://public.api.bsky.app/xrpc/app.bsky.actor.searchActorsTypeahead'
+	);
+
+	return '' === $url ? '' : \esc_url_raw( $url );
```
pfefferle: `\esc_url_raw( '' )` already returns `''`, so the guard is a no-op:

```suggestion
	return \esc_url_raw( $url );
```

tests/phpunit/tests/class-test-connectors.php
```diff
@@ -0,0 +1,332 @@
...
+		$this->assertSame( 'kept', $data['existing'] );
+		$this->assertArrayHasKey( 'isConnected', $data );
+	}
+
+	/**
+	 * The payload ships a default handle-typeahead endpoint for the card,
+	 * defaulting to Bluesky's official public appview.
+	 *
+	 * @covers ::get_connector_data
+	 * @covers ::typeahead_url
```
pfefferle: If the `typeahead_url()` wrapper goes, this `@covers` tag goes with it:

```suggestion
	 * @covers ::get_connector_data
```

tests/phpunit/tests/class-test-connectors.php
```diff
@@ -0,0 +1,332 @@
...
+	 */
+	public function test_typeahead_url_can_be_disabled() {
+		\add_filter( 'atmosphere_handle_typeahead_url', '__return_empty_string' );
+
+		$this->assertSame( '', Connectors::typeahead_url() );
+
+		\remove_all_filters( 'atmosphere_handle_typeahead_url' );
+	}
+
```
pfefferle: These two duplicate the `atmosphere_handle_typeahead_url` filter coverage already in `class-test-functions.php` (`test_handle_typeahead_url_is_filterable` / `..._can_be_disabled`) and only exercise the pass-through — and they fatal once the wrapper is removed:

```suggestion
```

includes/class-reaction-sync.php
```diff
@@ -373,6 +373,20 @@ public static function filter_avatar_data( array $args, $id_or_email ): array {
...
+		 * Connection-only mode: another plugin owns the connection and does
+		 * not want ATmosphere reaching out to the PDS on its own. Skip the
+		 * whole poll — the per-item gates already drop any writes, but bailing
+		 * here also spares the hourly `listNotifications` + `listRecords`
+		 * calls. Deliberately return *before* the watermarks advance, unlike
+		 * the per-setting toggles: if the host later leaves connection-only
+		 * mode, the interactions that arrived meanwhile are picked up rather
+		 * than skipped for good.
+		 */
+		if ( is_connection_only_mode() ) {
```
pfefferle: This early return checks raw `is_connection_only_mode()` rather than the per-feature helpers, so the documented contract — `atmosphere_should_sync_reactions` / `atmosphere_should_sync_replies` run last and can re-enable a lane even in connection-only mode — can't actually happen for sync: the filters are never reached. Auto-publish and comment publishing honor the override; the two sync lanes don't.

Keeping the watermark semantics the comment above describes, the bail could defer to the helpers:

```php
if ( is_connection_only_mode() && ! is_reaction_sync_enabled() && ! is_reply_sync_enabled() ) {
	return;
}
```

(The cron scheduling gate in `class-atmosphere.php` has the same shape — noted there.)

includes/class-atmosphere.php
```diff
@@ -327,7 +330,10 @@ public function init(): void {
 		\add_action( 'atmosphere_backfill_replies', array( Reaction_Sync::class, 'backfill_scheduled_replies' ) );
 		Reaction_Sync::register();
 
-		if ( ! \wp_next_scheduled( 'atmosphere_sync_reactions' ) && is_connected() ) {
+		// Skip scheduling in connection-only mode: the sync would short-circuit
+		// anyway (see Reaction_Sync::sync()), so there's no point running the
+		// hourly event against a connection another plugin is managing.
+		if ( ! \wp_next_scheduled( 'atmosphere_sync_reactions' ) && is_connected() && ! is_connection_only_mode() ) {
```
pfefferle: Counterpart to the note in `Reaction_Sync::sync()`: gating scheduling on raw `is_connection_only_mode()` means a site that re-enables a sync lane via the `atmosphere_should_sync_*` filters never gets the hourly event scheduled at all. Something like `( ! is_connection_only_mode() || is_reaction_sync_enabled() || is_reply_sync_enabled() )` keeps the skip for the common case while honoring the override.

includes/rest/admin/class-connection-controller.php
```diff
@@ -0,0 +1,184 @@
...
+
+		if ( '' === $handle ) {
+			return new WP_Error(
+				'atmosphere_missing_handle',
+				\__( 'Enter your AT Protocol handle to connect.', 'atmosphere' ),
+				array( 'status' => 400 )
+			);
+		}
+
+		$url = Client::authorize( $handle );
```
pfefferle: `Sanitize::handle()` re-validates the URL `Client::authorize()` returns before redirecting (scheme + host defence-in-depth, per the comment there). This path skips that check and hands the URL to the card, which navigates client-side — where a `javascript:` URI slipping through a misconfigured `atmosphere_*` filter would be worse than a server-side redirect. Worth running the same scheme/host validation here before returning `{ url }`, or extracting it so both entry points share one normalize-validate-authorize step.

src/shared/handle-typeahead.js
```diff
@@ -0,0 +1,248 @@
...
+ * Used by both the Settings → Connectors card (a script module) and the
+ * Settings → ATmosphere connect field (a classic script). Reads `@wordpress/*`
+ * from the `window.wp.*` globals rather than importing them, because a script
+ * module can't `import` those packages — the same constraint the card was
+ * written under — which is also why this uses `createElement` directly, no JSX.
+ */
+
+const { createElement: el, useState, useRef, useEffect } = window.wp.element;
+const { __ } = window.wp.i18n;
+const { Spinner } = window.wp.components;
```
pfefferle: This is the only thing pulling `wp-components` (JS bundle + its stylesheet, declared as a dependency of `atmosphere-handle-typeahead` at enqueue) onto the classic settings page — all for one loading spinner. Core's spinner is free on every wp-admin page:

```js
el( 'span', { className: 'spinner is-active' } )
```

would drop the dependency on both the script and the style side.

pfefferle (commented):
Ran a cleanup pass over the branch — no correctness issues, these are all simplification/consistency notes. The inline suggestions cover the mechanical ones (each is one-click applicable); three broader notes are anchored where they apply: the connection-only short-circuits vs. the documented `atmosphere_should_sync_*` filter contract, the authorize-URL validation the REST path skips, and the `wp-components` weight behind the typeahead spinner.

A few minor ones not worth their own thread:

- `assets/css/handle-typeahead.css` is registered with the same arguments in two places (`Connectors::maybe_enqueue()` and `Admin`) — could be one shared helper.
- The card round-trips `connectorId` through the hydration payload and falls back to a hardcoded `'atmosphere'` in JS (`src/connectors-card/index.js:330`) — the id is a constant both sides already know, so the payload key and the fallback are redundant with each other.
- `submenu_screen_url()` walks every submenu parent; the Gutenberg screen only ever registers under `options-general.php`, so the scan could be bounded to that parent.
- `stub_resolver_chain()` is duplicated between `class-test-connection-controller.php` and `class-test-sanitize-handle.php` — trait candidate if it grows a third copy.

includes/wp-admin/class-admin.php
jeherve: Done in 9688eff — the plugin-row link now uses `\esc_url( settings_url() )`.

includes/class-connectors.php
jeherve: Done in 9688eff. Dropped the exact-match clause; `\str_contains( $hook_suffix, self::SCREEN_SLUG )` already matches core's `options-connectors.php` suffix on its own.

includes/class-connectors.php
jeherve: Done in 9688eff — the payload calls `handle_typeahead_url()` directly.

includes/class-connectors.php
jeherve: Done in 9688eff — the wrapper is gone, along with its two dedicated tests. The filter behaviour stays covered by the `handle_typeahead_url()` tests in `class-test-functions.php`.

includes/functions.php
jeherve: Done in 9688eff — `handle_typeahead_url()` now just returns `\esc_url_raw( $url )`.

tests/phpunit/tests/class-test-connectors.php
jeherve: Done in 9688eff — the `@covers ::typeahead_url` tag is gone with the wrapper.

tests/phpunit/tests/class-test-connectors.php
jeherve: Done in 9688eff — both tests removed.

includes/class-reaction-sync.php
jeherve: Fixed in 8306e58. `sync()` now bails only when both `is_reaction_sync_enabled()` and `is_reply_sync_enabled()` are false, so a lane re-enabled through the `atmosphere_should_sync_*` filter reaches the poll and the per-item gates keep the disabled lane's writes out. The watermark-deferral semantics the comment describes are preserved (the bail still returns before the watermarks advance). Added a regression test — `test_connection_only_mode_reenabled_lane_still_polls`.

includes/class-atmosphere.php
jeherve: Fixed in 8306e58 alongside `sync()`. Scheduling now gates on `! is_connection_only_mode() || is_reaction_sync_enabled() || is_reply_sync_enabled()`, so a site that re-enables a sync lane via the filters still gets the hourly event while the common connection-only case keeps the skip.

includes/rest/admin/class-connection-controller.php
jeherve: Fixed in 4696cad. Extracted the scheme/host check into `Sanitize::is_safe_authorize_url()` and run it from both entry points, so they share one normalize-validate step. The REST route now returns a 400 (`atmosphere_unsafe_authorize_url`) rather than handing an unsafe `{ url }` to the card. Added unit coverage for the validator (`javascript:` / `data:` / `http:` / scheme-relative / empty all rejected, absolute `https` accepted).

src/shared/handle-typeahead.js
jeherve: Done in 9688eff. The shared component renders core's `.spinner` (`el( 'span', { className: 'spinner is-active' } )`), which is free on every wp-admin page, and the classic settings page drops `wp-components` from both its script and style dependencies. The card keeps the `wp-components` style dependency, since its own `Button` / `Notice` still need that stylesheet.

jeherve:
On the four minor notes from the review body:

- **`assets/css/handle-typeahead.css` registered in two places** — after the spinner change (9688eff) the two enqueues no longer share arguments: the card keeps the `wp-components` style dependency (its `Button`/`Notice` need that stylesheet) while the classic settings page drops it, so folding them into one helper would paper over a real per-screen difference. Left them separate.
- **`connectorId` round-trips through the payload + hardcoded fallback** — done in 9688eff. The card hardcodes the id like `MODULE_ID` (`const CONNECTOR_ID = 'atmosphere'`), and both the payload key and the JS fallback are gone.
- **`submenu_screen_url()` walks every submenu parent** — left as-is. The screen is matched by the `SCREEN_SLUG` marker everywhere else (`is_connectors_screen()`) rather than by a fixed parent, so bounding this one spot to `options-general.php` would be the only place that hardcodes the Gutenberg menu placement — and the scan is over a handful of menu entries. Happy to bound it if you'd prefer the tighter assumption.
- **`stub_resolver_chain()` duplicated across two test files** — agreed it's a trait candidate; leaving the two copies until a third appears, per your note.

includes/class-connectors.php
```diff
@@ -0,0 +1,305 @@
...
+		$registry->register(
+			self::CONNECTOR_ID,
+			array(
+				'name'           => \__( 'ATmosphere', 'atmosphere' ),
+				'description'    => \__( 'Connect your account to publish to the AT Protocol network with ATmosphere.', 'atmosphere' ),
+				'type'           => 'cloud_service',
+				'logo_url'       => ATMOSPHERE_PLUGIN_URL . 'assets/images/atmosphere.svg',
+				'authentication' => array(
+					'method' => 'none',
+				),
```
pfefferle, line 120: Core's registry accepts an optional `plugin` arg (`file` + `is_active`) and feeds it to the screen's script-module payload as `isInstalled`/`isActivated` for the card UI. Omitting it is harmless today — registration only happens while the plugin is active, so the `__return_true` default stays truthful — but passing the file is strictly more complete and future-proofs against core using it for "manage plugin" affordances:

```suggestion
				'logo_url'       => ATMOSPHERE_PLUGIN_URL . 'assets/images/atmosphere.svg',
				'plugin'         => array(
					'file' => \plugin_basename( ATMOSPHERE_PLUGIN_FILE ),
				),
				'authentication' => array(
					'method' => 'none',
				),
```

includes/wp-admin/class-admin.php
```diff
@@ -128,9 +285,18 @@ public static function handle_oauth_callback(): void {
 			\sanitize_text_field( $state )
 		);
 
+		// Return the browser to wherever the flow started, on success *or*
+		// failure. Consume the flag before branching so a failed callback still
+		// lands somewhere sane instead of stranding the admin on a hidden
+		// settings page (see render_page()), and so a stale flag can't leak.
+		$destination = self::oauth_return_destination();
+
```
pfefferle, line 320: With the Connectors card, connections now start and finish outside the settings page — but nothing observable fires when that happens. Core's Connectors API deliberately has no connection-state concept for `method => 'none'` connectors, so a plugin building on this connection can only poll `is_connected()`; a card-driven connect or disconnect is invisible to it.

Worth adding a small lifecycle trio while this surface is still `@since unreleased`:

- `atmosphere_connected` — here, after `Client::handle_callback()` succeeds;
- `atmosphere_disconnected` — in `Client::disconnect()`;
- `atmosphere_reauth_required` — where the `needs_reauth` flag gets set.

That gives host plugins cache invalidation and their own reconnect notices without polling, and it's much cheaper to add now than to retrofit once consumers exist.

includes/class-connectors.php
jeherve: Done in c765405 — added `'plugin' => array( 'file' => \plugin_basename( ATMOSPHERE_PLUGIN_FILE ) )`, letting core's default `is_active` (`is_plugin_active()`) resolve activation state. Added a registration-arg assertion to `test_register_declares_none_auth_connector`.

includes/wp-admin/class-admin.php
jeherve: Good call — added in 32c2e5d, since it's much cheaper while this surface is still `@since unreleased`. Added the lifecycle trio at the three sites you pointed to:

- `atmosphere_connected( $did, $handle )` — in `Client::handle_callback()` after the connection persists.
- `atmosphere_disconnected( $did )` — at the end of `Client::disconnect()`, once the local state is wiped.
- `atmosphere_reauth_required( $did, $reason )` — in `Client::mark_needs_reauth()`, guarded to fire only on the first false→true transition (the method is idempotent and reachable from several failure paths, so firing on every mark would be noisy).

Host plugins get cache invalidation / their own reconnect notices without polling. Documented in the public-hooks catalog (`developer-docs.md` + `php-coding-standards.md`) and covered by tests (`test_handle_callback_fires_connected_action`, `test_disconnect_fires_disconnected_action`, `test_permanent_failure_fires_reauth_required_action`).

pfefferle (commented):
Second pass over the branch (the earlier suggestions, the lifecycle actions, the `plugin.file` arg, and the authorize-URL re-validation are all addressed meanwhile). The remaining findings are anchored inline — the two I'd fix before merge are the invisible-error regression in `handle_oauth_callback()` and the WP < 7.0 return-destination 404.

Smaller notes not worth their own threads:

- Several new multiline comments use consecutive `//` lines (`class-atmosphere.php` ~333, `oauth/class-client.php` ~1152, `wp-admin/class-admin.php` ~160/173/186/287, `class-connectors.php` ~242) — house style elsewhere in the repo is `/* */` blocks.
- `use function Atmosphere\handle_typeahead_url;` in `class-connectors.php` is a same-namespace no-op import.
- `filter_action_links()` uses inline `style="color:#646970;"` for the hidden-state label — a class in `assets/css/admin.css` would respect admin color schemes and slightly better contrast on striped table rows.
- No CI job rebuilds and diffs the committed `build/` output, and the new `webpack.config.js` override makes drift between `src/` and the shipped `build/connectors-card/index.js` more likely; deploy ships whatever was committed.
- `build/connectors-card/block.json` / `build/settings-connect/block.json` declare blocks that are never registered and ship in release zips — Plugin Check or future scanners may trip on them.
- The shipped bundle minified two `__()` calls into `__( isExpanded ? 'Close' : 'Details', 'atmosphere' )` — fine if POT extraction runs against `src/`, unextractable if it ever runs against release artifacts (which now export-ignore `/src`).
- Card polish: the Details toggle has `aria-expanded` but no `aria-controls`/id on the revealed panel, and `disconnect()` reloads the page, dumping keyboard focus to the top.
- Tests: the `$submenu`/`set_current_screen` fixtures in `class-test-connectors.php` are restored inline after assertions (a failing assertion leaks state to later tests — `Test_Settings_Page_Visibility` shows the snapshot/restore-in-`set_up`/`tear_down` pattern), and `add_filter_tracked()` in `class-test-sanitize-handle.php` passes `PHP_INT_MAX` as `$accepted_args` (looks like a priority-slot mixup; harmless but confusing).

includes/wp-admin/class-admin.php
```diff
@@ -128,9 +284,18 @@ public static function handle_oauth_callback(): void {
...
 		if ( \is_wp_error( $result ) ) {
 			\add_settings_error( 'atmosphere', 'callback_failed', $result->get_error_message() );
-			return;
+			\set_transient( 'settings_errors', \get_settings_errors(), 30 );
+
+			\wp_safe_redirect( \add_query_arg( 'atmosphere_error', '1', $destination ) );
```
pfefferle: A failed connect is now silent — a behavior regression vs. trunk. There, the failure branch ran `add_settings_error()` and returned, so `options-head.php` rendered the error inline in the same request. On this branch the error goes into the `settings_errors` transient and the browser is redirected — but core only merges that transient back when `$_GET['settings-updated']` is set (`wp-admin/includes/template.php`, `get_settings_errors()`), neither redirect adds it, the Connectors screen never calls `settings_errors()` at all, and nothing PHP- or card-side reads `atmosphere_error`. So an expired state or failed handle resolution lands the admin back on the origin screen with no feedback; as a bonus, the orphaned 30-second transient can surface the stale error on an unrelated `options.php` save.

Fix shape: append `settings-updated=1` to the settings-page destination so core displays the transient, and give the Connectors destination a reader — an `admin_notices` handler for `atmosphere_error`, or hand the flag to the card via the hydration payload. (The success-path `connected=1` notice has the same non-display property, but that part is pre-existing trunk behavior.)

includes/class-connectors.php
```diff
@@ -0,0 +1,296 @@
...
+	 * The URL is resolved from the server-side admin menu, never from request
+	 * input, so it stays a trusted, hardcoded-shape destination for
+	 * `wp_safe_redirect()`.
+	 *
+	 * @return string Admin URL of the Connectors screen.
+	 */
+	public static function screen_url(): string {
+		$submenu_url = self::submenu_screen_url();
+
+		return null !== $submenu_url ? $submenu_url : \admin_url( self::SCREEN );
```
pfefferle: The REST `authorize()` route registers on WP ≥ 6.5 and always sets the Connectors-origin flag, but this fallback points at `options-connectors.php` — a file that only exists on WP 7.0. A host plugin driving connect through the route on WP 6.5–6.9 (the embedding use case) gets a 404 after the OAuth callback:

```suggestion
		if ( null !== $submenu_url ) {
			return $submenu_url;
		}

		/*
		 * On WP < 7.0 core's top-level Connectors page doesn't exist, so
		 * falling back to it would 404. Land on the settings page instead.
		 */
		if ( ! \class_exists( 'WP_Connector_Registry' ) ) {
			return settings_url();
		}

		return \admin_url( self::SCREEN );
```

includes/class-connectors.php
```diff
@@ -0,0 +1,296 @@
...
+	 *
+	 * @param \WP_Connector_Registry $registry The core connector registry.
+	 * @return void
+	 */
+	public static function register( $registry ): void {
+		$registry->register(
+			self::CONNECTOR_ID,
+			array(
+				'name'           => \__( 'ATmosphere', 'atmosphere' ),
+				'description'    => \__( 'Connect your account to publish to the AT Protocol network with ATmosphere.', 'atmosphere' ),
```
pfefferle, line 112: Copy note, per the plugin's end-user-string convention (lead with "Bluesky", plain language — the recent reauth copy says "saved Bluesky login" for the same reason): several new user-facing strings lead with "AT Protocol" instead —

- this connector description ("publish to the AT Protocol network");
- three strings in `Connection_Controller` ("The AT Protocol handle to connect…", "You are not allowed to manage the AT Protocol connection.", "Enter your AT Protocol handle to connect." — the latter two surface directly in the card);
- the card's session-expired notice in `src/connectors-card/index.js`;
- the visible field label "AT Protocol handle" in `src/shared/handle-typeahead.js`;
- the `add-settings-handle-typeahead` changelog entry ("suggest AT Protocol handles as you type").

Since this card deliberately isn't branded to a single app, something like "your Bluesky (AT Protocol) account" keeps both audiences.

includes/wp-admin/class-admin.php
```diff
@@ -84,12 +157,61 @@ public static function enqueue_assets( string $hook_suffix ): void {
...
+			\array_merge(
+				$asset['dependencies'] ?? array(),
+				array( 'wp-element', 'wp-i18n' )
+			),
+			$asset['version'] ?? ATMOSPHERE_VERSION,
+			true
+		);
+
+		\wp_localize_script(
+			'atmosphere-settings-connect',
```
pfefferle, line 232: The `atmosphere-settings-connect` script declares `wp-i18n` as a dependency and its sources translate everything with `__( …, 'atmosphere' )`, but nothing loads the locale data — so all of it renders English-only at runtime. `class-block-editor.php` sets it for the editor scripts; same one-liner here:

```php
\wp_set_script_translations( 'atmosphere-settings-connect', 'atmosphere' );
```

The `@atmosphere/connectors-card` script module has the same problem with the extra wrinkle that core's script-module i18n story is still limited — worth at least a code comment / follow-up issue so the untranslated card strings are a known gap rather than an accident.

assets/css/handle-typeahead.css
```diff
@@ -0,0 +1,81 @@
...
+	display: flex;
+	align-items: center;
+	gap: 8px;
+	padding: 6px 10px;
+	cursor: pointer;
+}
+
+.atmosphere-handle-field__option.is-active {
+	background: #f0f0f0;
+}
```
pfefferle, line 76: During keyboard navigation DOM focus stays on the input, so this background is the **only** visual indicator of which suggestion <kbd>Enter</kbd> will pick — and `#f0f0f0` on white is ~1.07:1, far below the 3:1 WCAG 1.4.11 needs for non-text indicators. Admin-blue with inverted text is the usual fix:

```suggestion
.atmosphere-handle-field__option.is-active {
	background: #2271b1;
	color: #fff;
}
```

(If the option renders secondary text in a fixed gray, that needs an `.is-active` override to `inherit` too.)

src/shared/handle-typeahead.js
```diff
@@ -0,0 +1,250 @@
...
+				name,
+				type: 'text',
+				className: inputClassName,
+				role: 'combobox',
+				autoComplete: 'off',
+				placeholder: 'alice.bsky.social',
+				value,
+				disabled,
+				'aria-expanded': open,
+				'aria-controls': suggestionsId,
```
pfefferle: Two screen-reader gaps in the combobox:

1. **Nothing is announced.** The spinner is a purely visual span, and there's no `aria-live` status region — an SR user types a handle and never learns that suggestions appeared, how many, or that none matched. A visually-hidden `role="status"` element updated with "N suggestions" / "No matches" (or `wp.a11y.speak()`) covers it; the card's error `Notice` on failed connect/disconnect deserves a `speak()` call for the same reason.
2. **`aria-controls` dangles while closed.** It always points at `suggestionsId`, but the listbox is only rendered while `open` — an invalid idref the rest of the time. Either render the listbox unconditionally (empty/hidden) or set the attribute only while open.

tests/phpunit/tests/rest/admin/class-test-connection-controller.php
```diff
@@ -0,0 +1,228 @@
...
+		\wp_set_current_user( self::factory()->user->create( array( 'role' => 'administrator' ) ) );
+		$this->stub_resolver_chain();
+
+		$request = new WP_REST_Request( 'POST', self::AUTHORIZE );
+		$request->set_param( 'handle', 'alice.atmosphere-test.io' );
+
+		$response = \rest_do_request( $request );
+
+		if ( 200 !== $response->get_status() ) {
+			$this->markTestSkipped( 'Resolver chain rejected the stubbed handle.' );
```
pfefferle: This `markTestSkipped()` means any future change that breaks the stubbed resolver chain silently stops verifying the Connectors-flag contract (the security-relevant bit of this route) without ever failing CI — same for its model, `test_authorize_persists_encrypted_jwk_transient`. A hard assertion on the 200 would turn drift into a red build. Related: the resolver path runs a real `dns_get_record( '_atproto.…', DNS_TXT )` before the stubbed well-known fallback, so these tests also depend on DNS egress from the test environment.

includes/class-connectors.php
jeherve: I'd like to push back on that one. Don't you think this should be an opportunity to move away from the Bluesky language?

includes/wp-admin/class-admin.php
jeherve: Fixed in fcbd5fc. Replaced the `settings_errors`-transient approach with a plugin-owned `atmosphere_oauth_notice` transient rendered by a dedicated `admin_notices` handler (`Admin::maybe_render_oauth_notice()`), which fires on whichever screen the callback redirected to — settings page or Connectors screen alike. That also drops the core-transient pollution you flagged, so a stale error can't surface on an unrelated `options.php` save. Both outcomes now display, and the notice is deleted on first render. Added tests for the render-once + `manage_options` gate.

includes/class-connectors.php
jeherve: Fixed in fcbd5fc, applying your suggestion: `screen_url()` returns `settings_url()` when `WP_Connector_Registry` is absent, so a WP 6.5–6.9 host driving connect through the REST route no longer returns to a non-existent `options-connectors.php`.

One caveat on coverage: the wp-env test environment defines `WP_Connector_Registry`, and a class can't be un-defined mid-run, so the registry-absent branch isn't directly unit-tested — the existing `screen_url()` tests cover the registry-present paths and I documented the gap in the test.

includes/wp-admin/class-admin.php
jeherve: Fixed in fcbd5fc — added `\wp_set_script_translations( 'atmosphere-settings-connect', 'atmosphere' )` after the enqueue. For the `@atmosphere/connectors-card` script module I left a code comment marking its strings as a known untranslated gap (core's script-module i18n story is still limited) rather than wiring something that doesn't yet work. That ties into the release-artifact POT-extraction note in the review body — replied there.

assets/css/handle-typeahead.css
jeherve: Fixed in b0ed2f4, applying your suggestion (admin-blue `#2271b1`, white text) — plus an `.is-active` override so the secondary handle/name text (`#757575`) stays legible on the highlighted row rather than sitting at ~2:1 against the blue.

src/shared/handle-typeahead.js
jeherve: Both fixed in b0ed2f4:

1. Added a visually-hidden `role="status"` `aria-live="polite"` region that announces the suggestion count (`_n( '%d handle found.', … )`) or "No matching handles found." as results settle, and routed the card's connect/disconnect errors through `window.wp.a11y.speak()` — guarded, since a script module can't declare a `wp-a11y` dependency, so it degrades to a no-op when the global is absent.
2. `aria-controls` now references the listbox only while it's open (rendered), so the idref never dangles; same treatment for the card's Details disclosure toggle.

tests/phpunit/tests/rest/admin/class-test-connection-controller.php
jeherve: Fixed in fcbd5fc — both `test_authorize_flags_connectors_initiated_connect` and `test_authorize_persists_encrypted_jwk_transient` now hard-assert (`assertSame( 200, … )` / `assertNotWPError()`) instead of `markTestSkipped()`, so a stubbed-chain regression turns the build red.

On the DNS dependency: the resolver's `_atproto` TXT lookup returns nothing for these fake test domains, so resolution deterministically falls through to the stubbed well-known — the hard assertion passes without DNS egress (verified in wp-env). Called that out in the test comments.

jeherve:
On the smaller notes from the review body:

- **Consecutive `//` comment blocks** — converted the flagged ones (`class-atmosphere.php`, `oauth/class-client.php`, `wp-admin/class-admin.php`, `class-connectors.php`) to `/* */` in fcbd5fc.
- **Same-namespace `use function` import** — removed in fcbd5fc.
- **`build/*/block.json` artifacts** — the connectors-card and settings-connect entries are enqueued by path (a script module / classic script), never registered as blocks, so I `export-ignore`d their `build/*/block.json` in fcbd5fc to keep them out of release zips.
- **Test fixtures + `PHP_INT_MAX` accepted-args** — fcbd5fc: the `$submenu` / current-screen fixtures now snapshot in `set_up()` and restore in `tear_down()` so a failed assertion can't leak them into later tests; `add_filter_tracked()` takes a proper `$accepted_args` (default 1) instead of `PHP_INT_MAX`.
- **Card Details disclosure a11y** — added the `aria-controls`/id pairing in b0ed2f4.

Three I'd rather handle as follow-ups than fold into this PR:

- **`filter_action_links()` inline style** — kept as-is. `assets/css/admin.css` is only enqueued on the settings page, not the Plugins screen where this label renders, so moving to a class there wouldn't apply without also loading a stylesheet on `plugins.php` for one muted label. `#646970` is core's standard secondary-text token, so it already tracks the admin palette. Happy to add a dedicated Plugins-screen stylesheet if you think it's worth it.
- **No CI rebuild/diff of `build/`, and POT extraction against release artifacts** — both real, and related: with `/src` export-ignored and no `make-pot` step, WP.org scans the minified `build/` for JS strings generally (not just the `Close`/`Details` ternary). The runtime piece (`wp_set_script_translations`) is in this PR; the proper fix is a build/release step that (a) rebuilds-and-diffs `build/` in CI and (b) ships a generated POT so translation never depends on scanning minified JS. That's its own change — I'll open a follow-up covering it plus the card-module i18n gap.
- **`disconnect()` reloads the page, dropping keyboard focus to the top** — real, but fixing it properly means re-rendering the card in place from client-side connection state instead of `window.location.reload()`, which is a bigger rework of the card's state model than fits here. Follow-up.

(The "AT Protocol" vs "Bluesky" copy note is being discussed on its own thread.)

includes/class-connectors.php
pfefferle: Ooops! Totally! AI is hallucinating!

includes/class-reaction-sync.php
```diff
@@ -373,6 +373,27 @@ public static function filter_avatar_data( array $args, $id_or_email ): array {
...
+		 * than skipped for good.
+		 *
+		 * Defer to the per-feature helpers rather than raw
+		 * is_connection_only_mode() so the documented contract holds: the
+		 * `atmosphere_should_sync_reactions` / `atmosphere_should_sync_replies`
+		 * filters run last and can re-enable a lane even in connection-only
+		 * mode. When either lane is on, the poll proceeds and the per-item
+		 * gates keep the disabled lane's writes out.
+		 */
+		if ( ! is_reaction_sync_enabled() && ! is_reply_sync_enabled() ) {
```
pfefferle: This bail is broader than connection-only mode: the two helpers also return `false` when a regular site simply unchecks both sync toggles in settings. In that case the watermarks stop advancing — trunk kept advancing them, so the off period was "skipped for good", as the docblock above still promises. Re-enable a toggle weeks later and the next cron run walks the whole off-period backlog and imports every old like/repost/reply as brand-new comments (plus the matching notification emails).

```suggestion
		if ( is_connection_only_mode() && ! is_reaction_sync_enabled() && ! is_reply_sync_enabled() ) {
```

Related edge, maybe for a follow-up: even in connection-only mode, if a host re-enables just one lane the poll advances the shared pagination cursor past the *other* (forced-off) lane's items, so those are dropped for good — the "picked up later" promise only holds while both lanes are off.

includes/wp-admin/class-admin.php
```diff
@@ -128,26 +307,112 @@ public static function handle_oauth_callback(): void {
...
-			\add_settings_error( 'atmosphere', 'callback_failed', $result->get_error_message() );
-			return;
+			self::store_oauth_notice( 'error', $result->get_error_message() );
+
+			\wp_safe_redirect( \add_query_arg( 'atmosphere_error', '1', $destination ) );
+			exit;
 		}
 
 		// Auto-create publication on first connect.
 		Publisher::sync_publication();
```
pfefferle: `sync_publication()` runs on every successful callback regardless of mode, so a host that disabled every lane still gets a publicly visible `site.standard.publication` record written to the connected repo the moment the user connects through the Connectors card — and there's no helper or filter that can prevent it. Should this be gated behind a per-feature helper (or at least a filter) like the other lanes?

src/shared/handle-typeahead.js
```diff
@@ -0,0 +1,283 @@
...
+			setOpen( suggestions.length > 0 );
+			// Never point aria-activedescendant at a non-existent option: hold
+			// active at -1 while the list is empty, mirroring ArrowDown.
+			setActive( ( i ) =>
+				suggestions.length ? Math.max( i - 1, 0 ) : -1
+			);
+		} else if ( 'Enter' === e.key ) {
+			e.preventDefault();
+			if ( open && suggestions.length ) {
+				choose( suggestions[ active >= 0 ? active : 0 ].handle );
```
pfefferle: `setActive( actors.length ? 0 : -1 )` above auto-highlights the first suggestion after every search, so while the list is open Enter always submits `suggestions[ 0 ]` — never the typed text. Someone who fully types a handle the search index doesn't know (say a self-hosted PDS handle, `alice.selfhosted.example`) and hits Enter starts the OAuth flow for a different account (`alice.bsky.social`). Stale results also linger through the debounce window, so the open list may not even match the current input.

Enter should probably only pick a suggestion the user explicitly navigated to (arrows/hover) and otherwise submit the typed value; typing could also close/invalidate the open list.

One more nit in the same area: `choose()` never clears `timer.current`, so picking a suggestion within the 250 ms debounce lets the stale search run afterwards and reopen the dropdown over the busy field.

includes/wp-admin/class-admin.php
```diff
@@ -128,26 +307,112 @@ public static function handle_oauth_callback(): void {
...
+	 * Connectors screen alike — unlike the `settings_errors` transient, which
+	 * either screen may never surface. The notice is deleted on first render so
+	 * it shows exactly once, on the screen the callback redirected to.
+	 */
+	public static function maybe_render_oauth_notice(): void {
+		if ( ! \current_user_can( 'manage_options' ) ) {
+			return;
+		}
+
+		$notice = \get_transient( self::OAUTH_NOTICE_TRANSIENT );
```
pfefferle: Two problems with this mechanism:

1. The transient is site-wide and deleted on first render — the first `manage_options` user whose `admin_notices` fires within the 60-second TTL consumes it. Another admin (or your own second wp-admin tab) can eat the message, showing them a context-free notice and the connecting admin nothing; a failed connect then goes completely silent for the person who initiated it. Scoping the key by `get_current_user_id()` fixes this half.
2. The Connectors screen renders as a React app, and classic `admin_notices` output isn't reliably visible there — yet that first render still deletes the transient.

Since the redirect already carries `connected=1` / `atmosphere_error=1` query args, having the card surface the outcome from those (with the message fetched per-user) would fix both halves.

includes/wp-admin/class-admin.php
```diff
@@ -237,6 +502,12 @@ public static function maybe_render_reauth_notice(): void {
 			return;
 		}
 
+		// The notice's only call to action links to the settings page; skip it
+		// entirely when that page is hidden so we never point at a dead URL.
+		if ( ! self::is_settings_page_visible() ) {
```
pfefferle: In connection-only mode this removes the only site-wide signal that the stored session died. With a lane re-enabled by the host (e.g. `atmosphere_should_auto_publish`), publishing then short-circuits silently and no notice appears anywhere in wp-admin — `atmosphere_reauth_required` fires once at the transition, so a host activated after that never hears it either. Instead of skipping, could the CTA point at the Connectors screen when it exists (`Connectors::screen_url()`), and only be dropped when there's genuinely no screen to link?

includes/rest/admin/class-pre-publish-controller.php
```diff
@@ -298,10 +300,12 @@ private function publish_decision( WP_Post $post, string $status, string $passwo
...
 		}
 
-		if ( '1' !== \get_option( 'atmosphere_auto_publish', '1' ) ) {
+		if ( ! is_auto_publish_enabled() ) {
 			return array(
 				'will_publish' => false,
-				'reason'       => \__( 'Automatic publishing to Bluesky is turned off in settings.', 'atmosphere' ),
+				'reason'       => is_connection_only_mode()
+					? \__( 'Automatic publishing to Bluesky is turned off by another plugin on this site.', 'atmosphere' )
+					: \__( 'Automatic publishing to Bluesky is turned off in settings.', 'atmosphere' ),
```
pfefferle, line 314: This attributes the off state to settings whenever connection-only mode is off — but `atmosphere_should_auto_publish` can force it off on its own, in which case the editor is told "turned off in settings", finds the checkbox already checked, and has no path to the real cause. Comparing the stored option to the effective value would let this branch attribute filter overrides to "another plugin" as well (the reply-sync error in `Reaction_Sync` has the same pattern).

includes/class-atmosphere.php
```diff
@@ -327,7 +330,18 @@ public function init(): void {
...
-		if ( ! \wp_next_scheduled( 'atmosphere_sync_reactions' ) && is_connected() ) {
+		/*
+		 * Skip scheduling in connection-only mode: the sync would short-circuit
+		 * anyway (see Reaction_Sync::sync()), so there's no point running the
+		 * hourly event against a connection another plugin is managing. Mirror
+		 * that method's contract, though — a host that re-enables a sync lane
+		 * through the `atmosphere_should_sync_*` filters still needs the event,
+		 * so honor the per-feature helpers rather than raw connection-only mode.
+		 */
+		$sync_scheduling_wanted = ! is_connection_only_mode()
```
pfefferle: Three gaps around this gate:

1. It's evaluated at `plugins_loaded` 10, before a host that re-enables a lane from its own `plugins_loaded`/`init` callback has attached its filter — so `$sync_scheduling_wanted` is `false` on every request and the event is never scheduled, even though `is_reaction_sync_enabled()` returns `true` by cron-dispatch time. Reactions then silently never import.
2. `atmosphere_backfill_replies` below isn't gated at all.
3. Nothing unschedules a pre-existing event, so a site that connected before entering connection-only mode keeps the hourly no-op cron forever (the teardown pattern already exists in `functions.php`).

Deferring the whole check to a later hook (e.g. `init`) and adding an unschedule branch when `$sync_scheduling_wanted` is false would cover all three.

includes/rest/admin/class-connection-controller.php
```diff
@@ -0,0 +1,201 @@
...
+		 */
+		if ( ! Sanitize::is_safe_authorize_url( $url ) ) {
+			return new WP_Error(
+				'atmosphere_unsafe_authorize_url',
+				\__( 'Authorization URL is not a safe HTTPS target.', 'atmosphere' ),
+				array( 'status' => 400 )
+			);
+		}
+
+		\set_transient( 'atmosphere_oauth_from_connectors', 1, HOUR_IN_SECONDS );
```
pfefferle: This flag is a site-wide transient with a 1-hour TTL and isn't tied to the OAuth `state` — and `Sanitize::handle()` deletes it unconditionally before every settings-page connect. Two overlapping flows from the two entry points therefore steal/clobber each other's return destination, and each admin lands on the wrong screen after their callback. Since `atmosphere_oauth_state` is itself a single site-wide transient (the flow is one-at-a-time by construction), the simplest fix is probably storing the origin alongside/inside the state transient so the two can't diverge — the callback then derives the destination from the same record it validates.

pfefferle (commented):
Automated high-effort review of `985803a` (multi-agent, adversarially verified) — nine inline comments on correctness, plus smaller cleanups that didn't warrant their own thread:

- **"AT Protocol" in new end-user strings** — the connector description (`class-connectors.php:111`), "Enter your AT Protocol handle to connect." / "You are not allowed to manage the AT Protocol connection." (`class-connection-controller.php`), the "AT Protocol handle" field label (`handle-typeahead.js`), and "Your AT Protocol session expired." (`connectors-card`). The class docblock argues the card is deliberately app-neutral, so maybe intentional — but it diverges from the lead-with-Bluesky wording used everywhere else in the admin UI.
- **`atmosphere-handle-typeahead` stylesheet enqueued twice with contradictory deps** — `Connectors::maybe_enqueue()` registers it with `array( 'wp-components' )`, `Admin::enqueue_assets()` with `array()` plus a comment explaining why wp-components is *not* needed. A shared enqueue helper (or matching arrays) would keep the two screens honest.
- **Four copies of the option → connection-only → filter resolver** — `is_comment_publishing_enabled()`, `is_auto_publish_enabled()`, `is_reaction_sync_enabled()`, `is_reply_sync_enabled()` share the same body modulo names; one private resolver (filter tag kept literal per lane) would prevent drift.
- **Handle normalization duplicated** between the REST authorize route and `Sanitize::handle()` (sanitize + @-strip); a shared `normalize_handle()` keeps the two connect flows identical by construction.

includes/wp-admin/class-health-check.php
```diff
@@ -254,7 +255,7 @@ public static function debug_information( $info ) {
 				),
 				'auto_publish'      => array(
 					'label'   => \__( 'Auto-Publish', 'atmosphere' ),
-					'value'   => '1' === \get_option( 'atmosphere_auto_publish', '1' )
+					'value'   => is_auto_publish_enabled()
```
pfefferle: Trunk reported the stored `atmosphere_auto_publish` option here; switching to the effective helper means that once connection-only mode or the filter forces publishing off, debug info shows "Disabled" with no trace of the user's actual setting — exactly the stored-vs-effective discrepancy a support thread needs to see. Showing both (e.g. "Enabled (overridden by another plugin)") would keep it diagnosable.

includes/class-reaction-sync.php
jeherve: Good catch — this was a real regression I introduced. Fixed in 34d02f2 with exactly your suggestion: the early bail is now `is_connection_only_mode() && ! is_reaction_sync_enabled() && ! is_reply_sync_enabled()`, so a regular site that unchecks both toggles still polls and the per-item gates skip writes *while the watermarks advance* — the off period stays "skipped for good" instead of being replayed on re-enable. Added a regression test (`test_sync_still_polls_with_both_toggles_off_off_connection_only`).

On the related edge (a single re-enabled lane in connection-only mode advancing the shared cursor past the other, forced-off lane's items): agreed that's a real subtlety of the shared pagination cursor. It's narrower — it only bites a connection-only host that re-enables exactly one lane via filter — so I've left it as a follow-up rather than reworking the cursor here.

includes/wp-admin/class-admin.php
jeherve: Gated in 34d02f2 / ad8b6f0. Added `is_publication_sync_enabled()` (defaults on, forced off in connection-only mode, final say via a new `atmosphere_should_sync_publication` filter) and wrapped both automatic write paths — the connect callback (`Admin::handle_oauth_callback`) and the `atmosphere_sync_publication` cron worker — so a connection-layer host no longer gets a public `site.standard.publication` record written to the connected repo unless it opts back in.

src/shared/handle-typeahead.js
jeherve: Fixed in 7ab9b3a. The list no longer auto-highlights the first result (`setActive( -1 )` after each search), and Enter only picks a suggestion the user explicitly navigated to (`open && active >= 0`); otherwise it submits the typed value — so fully typing `alice.selfhosted.example` and hitting Enter connects that account, not `suggestions[0]`. Editing the field resets the highlight (stale results can't be picked mid-debounce), and `choose()` now clears `timer.current` so a pending search can't reopen the dropdown after a selection.

includes/wp-admin/class-admin.php
jeherve: Both halves fixed.

1. Per-user scoping (ad8b6f0): the notice transient is now keyed by `get_current_user_id()`, so a second admin or tab can't consume the message meant for the admin who initiated the connect.
2. Connectors screen (ad8b6f0 + 7ab9b3a): rather than rely on classic `admin_notices` on the React screen, the card now renders the outcome from its hydration payload (`Connectors::get_connector_data` calls `Admin::consume_oauth_notice()`), and `maybe_render_oauth_notice()` skips the Connectors screen so the two don't double-render or race on consuming the transient. The settings page keeps the `admin_notices` path.

includes/wp-admin/class-admin.php
jeherve: Fixed in ad8b6f0. In connection-only mode the reauth notice no longer disappears: it resolves the reconnect link to `Connectors::screen_url()` when a Connectors screen exists (WP 7.0+), and only bails when there's genuinely no screen to link to. The link text is now destination-agnostic ("reconnect your account"). The test in `Test_Settings_Page_Visibility` now asserts the notice points at the Connectors screen when the API is present, and is suppressed only without it.

includes/rest/admin/class-pre-publish-controller.php
jeherve: Fixed in 34d02f2. Both the editor pre-publish panel and the reply-backfill error now compare the stored option to the effective value: when auto-publish / reply sync is forced off by the `atmosphere_should_*` filter (not just connection-only mode) while the saved checkbox is still on, they attribute it to "another plugin" rather than "settings". So the editor no longer tells an author "turned off in settings" while their checkbox is checked.

includes/class-atmosphere.php
jeherve: All three fixed in 34d02f2.

1. The gate now runs on `init` (priority 20), not `plugins_loaded`, so a host that re-enables a lane from its own plugins_loaded/init callback has its filter attached before the gate is read.
2. `atmosphere_backfill_replies` is now gated too (on reply sync being wanted).
3. Added a reconcile step that *unschedules* an event when it's no longer wanted, so a site that enters connection-only mode after connecting doesn't keep an hourly no-op cron forever. Both events flow through one `reconcile_cron_event()` helper.

includes/rest/admin/class-connection-controller.php
jeherve: Fixed in ad8b6f0 along the lines you suggested. The origin now rides inside the flow's own `atmosphere_oauth_resolved` record: `Client::authorize( $handle, $origin )` writes it and `Client::pending_origin()` reads it back, and the callback derives the destination from that same record it validates. The separate site-wide `atmosphere_oauth_from_connectors` transient is gone, along with the unconditional `delete_transient()` in `Sanitize::handle()` — since the flow is one-at-a-time by construction (state/verifier/DPoP are all single transients), the origin can no longer diverge from the flow that ran.

includes/wp-admin/class-health-check.php
jeherve: Fixed in 34d02f2. Site Health now reports both: when the stored option and the effective value agree it shows "Enabled"/"Disabled" as before, but when connection-only mode or the filter overrides the saved preference it shows e.g. "Enabled in settings, overridden off by another plugin" — so a support thread sees the user's actual setting and the override.

jeherve:
On the four smaller notes from the review body:

- **Four copies of the option → connection-only → filter resolver** — extracted in 34d02f2. `is_auto_publish_enabled()`, `is_comment_publishing_enabled()`, `is_reaction_sync_enabled()`, and `is_reply_sync_enabled()` now share one `feature_option_enabled( $option )` for the option-read + connection-only override, while each keeps its own literal `apply_filters()` (and its docblock) at the call site so the hooks stay greppable.
- **Handle normalization duplicated** — extracted in ad8b6f0. Both connect entry points now call one `normalize_handle()` (sanitize + strip leading `@`); the REST route previously only cast + stripped, so it now sanitizes too, and the two flows are identical by construction.
- **`atmosphere-handle-typeahead` enqueued with contradictory deps** — left as-is intentionally, and I should have flagged this more clearly last round. The two enqueues *should* differ: the Connectors card is a script module that renders `Button`/`Notice`/`HStack`, whose styling comes from the `wp-components` stylesheet — and the only thing pulling that stylesheet onto the Connectors screen is this style's `wp-components` dependency, so the card must keep it. The classic settings page renders just a plain input (core `.spinner`, no wp-components), so declaring `wp-components` there would load a bundle it doesn't use. A shared helper would force them equal and reintroduce the dead dependency on the settings page. I added the explaining comment on the settings-page side; happy to mirror one on the card side if that reads clearer.
- **"AT Protocol" in new end-user strings** — this is intentional and there's an open thread about it: the card is deliberately app-neutral (it connects an AT Protocol account used for Bluesky *and* standard.site), and the preference here is to move *toward* AT Protocol wording rather than lead with Bluesky. Leaving the new strings as they are pending that discussion.

pfefferle (approved):

PR #206, opened 2026-07-20
Add custom publication theme color settings and filter

### LINKED ISSUE #205: Feature Request: ability to set custom colours for site.standard.publication basicTheme
### What

Ability to set the basic theme colours for standard.site publication without having to modify `theme.json` settings. 

It could be cool to have a preview in the admin of what the change looks like I well.

### Why

There doesn't seem to be any way to set the colours without changing your `theme.json` file, essentially hardcoding the colour choices to decisions made there.

### How

Initially I was thinking that I could add a filter with an array of values, or a settings page (or both, since the second approach probably needs the first anyway). Assumed I could just pass in an array of hex values somewhere, but there didn't seem to be a clean place to do that.

--- 

#### Experiment 1

I made a few attempts with open code and big pickle, this was where i stopped ([commit compare](https://github.com/Automattic/wordpress-atmosphere/compare/trunk...davekellam:wordpress-atmosphere:feature/add-publication-theme-colors-filter)). It's simple enough to read but ends up in a pretty gnarly nested array filter that people would need to create. Could probably just pass hexcodes, but it would need a utility function for that. 

```php
add_filter( 'atmosphere_publication_basic_theme_colours', function () {
    return array(
        'background' => array( 'r' => 255, 'g' => 255, 'b' => 255 ),
        'foreground' => array( 'r' => 0, 'g' => 0, 'b' => 0 ),
        'accent'     => array( 'r' => 0, 'g' => 102, 'b' => 204 ),
    );
} );
```

#### Experiment 2

Made another attempt with codex-5.3 i think, was just on auto in vs code ([commit compare](https://github.com/Automattic/wordpress-atmosphere/compare/trunk...davekellam:wordpress-atmosphere:feature/publication-theme-colors)). Trying to see if I could add in a settings page, but it's touching a lot and I haven't reviewed it, but adding it here. Got stuck a bit trying to test it visually and needing to connect to an account first.

pfefferle:
Lets site owners set the `site.standard.publication` `basicTheme` colors from Settings → ATmosphere, instead of the colors being pinned to whatever `theme.json` derives. Grew out of "Experiment 2" in #205.

Fixes #205

## Proposed changes:

* Adds three options — background, foreground, and accent — each stored as a hex value and sanitized to `#rrggbb` (blank means "keep the color derived from the active theme").
* Renders them as a **Publication theme** section on the settings screen using `wp-color-picker`.
* Applies the stored values in the Publication transformer, recomputing `accentForeground` so a custom accent keeps its contrast guarantee.
* Adds an `atmosphere_publication_basic_theme` filter so the whole theme object can be replaced (or omitted) in code — the filter form floated in #205's Experiment 1, kept alongside the settings UI.
* Re-syncs the publication record when any of the three settings change.

### Other information:

Note that a theme object is only published when all four colors resolve. If the active theme exposes no usable colors (a classic theme, or a theme that styles links with a gradient), setting only one or two colors here still results in no `basicTheme` — all three have to be set. The section copy says so.

## Testing instructions:

No Bluesky connection is needed to verify the record shape:

1. Go to **Settings → ATmosphere → Publication theme**. Confirm the three color fields render, that each label focuses its input, and that the picker offers your active theme's palette as swatches.
2. Leave all three blank and save, then inspect the derived record:
   ```
   wp eval 'print_r( ( new \Atmosphere\Transformer\Publication( null ) )->transform()["basicTheme"] );'
   ```
   The colors should match your theme.
3. Set just the **accent** color and save. Re-run the command: only `accent` changes, `background` / `foreground` stay derived, and `accentForeground` flips to black or white for contrast.
4. Set all three and save. Re-run: all three are yours.
5. Clear a field and save. Re-run: that color reverts to the derived value.
6. Paste something invalid (`rgb(1,2,3)`, `javascript:alert(1)`) into a field and save — it stores empty rather than being persisted, and the derived color is used.
7. With a connection active, saving any of the three should queue a publication sync (`wp cron event list | grep atmosphere`).

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Add color settings for your publication theme, so you can choose the background, text, and accent colors apps use when they display your site.

</details>

### DIFF
--- a/assets/css/admin.css
+++ b/assets/css/admin.css
@@ -10,3 +10,11 @@
 	display: inline-block;
 	margin-top: 2px;
 }
+
+.atmosphere-color-input {
+	width: 8rem;
+}
+
+.wp-picker-container {
+	vertical-align: middle;
+}
--- a/includes/class-atmosphere.php
+++ b/includes/class-atmosphere.php
@@ -282,6 +282,9 @@ class Atmosphere {
 		\add_action( 'update_option_site_icon', array( $this, 'schedule_publication_sync' ) );
 		\add_action( 'update_option_home', array( $this, 'schedule_publication_sync' ) );
 		\add_action( 'update_option_siteurl', array( $this, 'schedule_publication_sync' ) );
+		\add_action( 'update_option_atmosphere_publication_theme_background', array( $this, 'schedule_publication_sync' ) );
+		\add_action( 'update_option_atmosphere_publication_theme_foreground', array( $this, 'schedule_publication_sync' ) );
+		\add_action( 'update_option_atmosphere_publication_theme_accent', array( $this, 'schedule_publication_sync' ) );
 		\add_action( 'switch_theme', array( $this, 'schedule_publication_sync' ) );
 		\add_action( 'save_post_wp_global_styles', array( $this, 'schedule_publication_sync' ) );
 		\add_action( 'customize_save_after', array( $this, 'schedule_publication_sync' ) );
--- a/includes/class-options.php
+++ b/includes/class-options.php
@@ -10,6 +10,7 @@ namespace Atmosphere;
 \defined( 'ABSPATH' ) || exit;
 
 use Atmosphere\Content_Parser\Registry;
+use Atmosphere\Transformer\Publication;
 
 /**
  * Registers every stored plugin option with the Settings API.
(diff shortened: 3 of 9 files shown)


pfefferle:
Superseded by #207, which keeps @davekellam's commit as-is and adds the review follow-ups on top. Moved because this branch lives on a fork with maintainer edits disabled, so the fixes could not be pushed here.

#207 also fixes a bug in trunk that this feature surfaced: on Twenty Twenty-Five no `basicTheme` was ever published, because TT5 styles links with `currentColor` and names its palette slugs `accent-1` through `accent-6`, while the fallback only looked for a slug named exactly `accent`. Thanks for the experiments in #205 that led here.

PR #207, opened 2026-07-20
Add custom publication theme colors, and fix accent derivation on default themes

### LINKED ISSUE #205: Feature Request: ability to set custom colours for site.standard.publication basicTheme

pfefferle:
Lets site owners set the `site.standard.publication` `basicTheme` colors from Settings → ATmosphere, instead of the colors being pinned to whatever `theme.json` derives.

Supersedes #206, which could not be pushed to (fork branch with maintainer edits disabled). @davekellam's original commit is preserved here as-is, with review follow-ups on top.

Fixes #205

## Proposed changes:

**The feature** (@davekellam's commit, unchanged):

* Three options, background, foreground, and accent, each stored as a hex value and sanitized to `#rrggbb`. Blank means "keep the color from the active theme".
* A **Theme** section on the settings screen using `wp-color-picker`.
* The stored values applied in the Publication transformer, recomputing `accentForeground` so a custom accent keeps its contrast guarantee.
* An `atmosphere_publication_basic_theme` filter so the whole theme object can be replaced or omitted in code. This was the filter shape floated as Experiment 1 in #205, kept alongside the settings UI.
* A publication re-sync when any of the three settings change.

**Fixes a bug in trunk that this feature made visible.** On Twenty Twenty-Five, the default theme, no `basicTheme` was ever published: TT5 styles links with `currentColor`, which `resolve_color()` cannot resolve, and the accent fallback only looked for a palette slug named exactly `accent` while TT5 ships `accent-1` through `accent-6`. Accent came back null, and a single null color drops the whole theme object. The accent fallback now takes the lowest-numbered resolvable `accent-N` slug, skipping entries that are not plain colors such as `color-mix(...)`. TT5 now yields `#ffffff` / `#111111` / `#ffee58` with a black accent foreground. This part stands on its own and is worth reviewing independently of the settings UI.

**Review follow-ups on top of the feature commit:**

* Overrides are resolved *before* derivation and passed into `build_basic_theme()`, so the record is assembled and `accentForeground` computed in one place. This drops `apply_theme_option_overrides()` and `color_object_to_rgb()`, the latter of which existed only to unwrap what `color_object()` had just wrapped.
* Overrides and the filter keep working when global styles are unavailable, rather than an early return skipping both.
* Sanitizing goes through core's `sanitize_hex_color()`, keeping the shorthand-to-`#rrggbb` normalization, so the settings layer no longer imports the transformer.
* One field callback instead of three near-identical ones, each row label tied to its input via `label_for`.
* The color picker is seeded with the active theme's palette as swatches, and each field's `data-default-color` is the color derived from the active theme, so adopting it is one click. Core renders its Default button in place of Clear when a default color is set, so the section copy tells you to empty a field to go back to the theme's color.
* Each field now states the color the active theme provides, and says so plainly when the theme provides none, because empty inputs gave no clue what a blank field publishes.

### Other information:

A theme object is only published when all four colors resolve. If the active theme exposes no usable colors, setting only one or two here still results in no `basicTheme`. All three have to be set, and the section copy says so.

Fields are intentionally *not* pre-filled with the derived colors: an empty field means "follow my theme" and re-derives at publish time, so pre-filling would silently convert that to a pinned override on the next save. The description text under each field states the theme's color, and the picker's Default button applies it deliberately.

## Testing instructions:

No Bluesky connection is needed to verify the record shape.

1. On a stock Twenty Twenty-Five install, before this branch, confirm the bug: `wp eval 'var_dump( array_key_exists( "basicTheme", ( new \Atmosphere\Transformer\Publication( null ) )->transform() ) );'` prints `false`. On this branch it prints `true`.
2. Go to **Settings → ATmosphere → Theme**. Each field should render, state the color your theme provides, focus its input when you click the row label, and offer your theme's palette as picker swatches.
3. Leave all three blank and save, then inspect the derived record:
   ```
   wp eval 'print_r( ( new \Atmosphere\Transformer\Publication( null ) )->transform()["basicTheme"] );'
   ```
   The colors should match your theme.
4. Set just the **accent** color and save. Re-run: only `accent` changes, `background` and `foreground` stay derived, and `accentForeground` flips to black or white for contrast.
5. Set all three and save. Re-run: all three are yours.
6. Clear a field and save. Re-run: that color reverts to the derived value.
7. Paste something invalid such as `rgb(1,2,3)` or `javascript:alert(1)` and save. It stores empty and the derived color is used.
8. With a connection active, saving any of the three should queue a publication sync: `wp cron event list | grep atmosphere`.

### Changelog entry

<details>

Committed in `.github/changelog/add-publication-theme-colors`:

> Significance: minor
> Type: added
>
> Add color settings for your publication theme, so you can choose the background, text, and accent colors apps use when they display your site.

</details>

### DIFF
--- a/.github/changelog/add-publication-theme-colors
+++ b/.github/changelog/add-publication-theme-colors
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Add color settings for your publication theme, so you can choose the background, text, and accent colors apps use when they display your site.
--- a/assets/css/admin.css
+++ b/assets/css/admin.css
@@ -20,11 +20,3 @@
 .atmosphere-handle-typeahead-mount .atmosphere-handle-field {
 	max-width: 25em;
 }
-
-.atmosphere-color-input {
-	width: 8rem;
-}
-
-.wp-picker-container {
-	vertical-align: middle;
-}
--- a/assets/css/admin.css
+++ b/assets/css/admin.css
@@ -20,3 +20,11 @@
 .atmosphere-handle-typeahead-mount .atmosphere-handle-field {
 	max-width: 25em;
 }
+
+.atmosphere-color-input {
+	width: 8rem;
+}
+
+.wp-picker-container {
+	vertical-align: middle;
+}
--- a/includes/class-atmosphere.php
+++ b/includes/class-atmosphere.php
@@ -285,6 +285,9 @@ class Atmosphere {
 		\add_action( 'update_option_site_icon', array( $this, 'schedule_publication_sync' ) );
 		\add_action( 'update_option_home', array( $this, 'schedule_publication_sync' ) );
 		\add_action( 'update_option_siteurl', array( $this, 'schedule_publication_sync' ) );
+		\add_action( 'update_option_atmosphere_publication_theme_background', array( $this, 'schedule_publication_sync' ) );
+		\add_action( 'update_option_atmosphere_publication_theme_foreground', array( $this, 'schedule_publication_sync' ) );
+		\add_action( 'update_option_atmosphere_publication_theme_accent', array( $this, 'schedule_publication_sync' ) );
 		\add_action( 'switch_theme', array( $this, 'schedule_publication_sync' ) );
 		\add_action( 'save_post_wp_global_styles', array( $this, 'schedule_publication_sync' ) );
 		\add_action( 'customize_save_after', array( $this, 'schedule_publication_sync' ) );
--- a/includes/class-atmosphere.php
+++ b/includes/class-atmosphere.php
@@ -279,30 +279,22 @@ class Atmosphere {
 		 * theme colours from: classic-theme Customizer saves
 		 * (`customize_save_after`) and block-theme Site Editor saves
 		 * (the `wp_global_styles` post update).
-		 */
-		\add_action( 'update_option_blogname', array( $this, 'schedule_publication_sync' ) );
-		\add_action( 'update_option_blogdescription', array( $this, 'schedule_publication_sync' ) );
-		\add_action( 'update_option_site_icon', array( $this, 'schedule_publication_sync' ) );
-		\add_action( 'update_option_home', array( $this, 'schedule_publication_sync' ) );
-		\add_action( 'update_option_siteurl', array( $this, 'schedule_publication_sync' ) );
-
-		/*
-		 * Both `add_option_*` and `update_option_*` for the theme colours:
-		 * they have no option row until a colour is first picked, and
-		 * WordPress routes that first save through `add_option()`, where
-		 * `update_option_*` never fires. Without the add hook the opening
-		 * colour choice would not reach the publication record. The core
-		 * options above always exist, so they need the update hook only.
+		 *
+		 * Each option gets both `add_option_*` and `update_option_*`: an
+		 * option with no row yet is written by `add_option()`, where
+		 * `update_option_*` never fires, which is the case a plugin
+		 * option hits the first time it is saved. Registering both for
+		 * every option keeps the list uniform, and the add hook is
+		 * simply never reached for the core options, which always exist.
 		 */
 		foreach (
-			array(
-				Publication::OPTION_THEME_BACKGROUND,
-				Publication::OPTION_THEME_FOREGROUND,
-				Publication::OPTION_THEME_ACCENT,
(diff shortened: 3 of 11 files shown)


kraftbj (approved):
This is clean, defensively-written work — no blocking issues. The per-channel override union, the accent-palette fallback ordering, and the `hex_color()` sanitizer all check out, and the settings/color-picker output is fully escaped. Approving.

## Verified: blank fields stay "follow theme" across a no-op save

I wanted to confirm the color inputs don't silently pin the derived color when someone saves Settings without touching them (which would quietly convert "follow theme" into a hard override that goes stale on the next theme switch). Traced it through core `wp-color-picker.js` `_create()`:

- It captures `initialValue = el.val()` (empty for a blank field), inits Iris, then **restores `el.val(self.initialValue)` right after** — so any color Iris seeds during its own init is overwritten back to empty.
- `$.extend(self.options, el.data())` does read `data-default-color`, but it only drives the Default button and the toggler swatch, not the submitted field value.
- The field value only changes on explicit user action (picking a color, or clicking Default/Clear).

So a "Save Changes" with untouched fields posts three empty strings and the overrides stay unset. The contract holds. No change needed — just noting it so it's on the record.

## Optional: pin the numbered-accent ordering in the test

`test_build_basic_theme_falls_back_to_numbered_palette_accent` lists `accent-1` before `accent-2` in the fixture, so the test would still pass if `ksort()` in `resolve_palette_accent()` were removed — plain iteration order already produces the same result. Since the "lowest-numbered `accent-N` wins" behavior is the core fix this PR ships, it'd be nice to make the test actually exercise the sort:

- Put the palette slugs out of order in the fixture (e.g. `accent-3` before `accent-1`) so iteration order and sorted order differ.
- Add a case where an exact `accent` slug is present alongside numbered ones, to lock in the `-1`-ranked precedence.

Not blocking — happy to see this land as-is and follow up separately.

A couple of even smaller notes, purely FYI (no action expected): the `atmosphere_publication_basic_theme` filter is only array-vs-not validated, so a malformed-but-array return would publish verbatim; and the derived colors the settings UI shows ("your theme provides #ffffff", plus the "set all three or nothing publishes" precondition) aren't exposed over REST, so an API/agent caller setting these is a bit blind compared to the UI. Both are fine to leave.

jeherve (approved):
This should be a nice improvement, and tests well. 🚢

davekellam:
Thanks so much! You guys are moving quick. Had been messing around Friday and mostly just wanting to get the idea out of my head before heading off to visit family. Think I would've needed to actually open proper PRs to set the maintainer edits, opted for lazy branch links 😅 Also wasn't sure about just opening a bunch of PRs for experiments. Appreciate the plugin in general 🙏🏻

PR #211, opened 2026-07-22
Add a filter to publish standard.site documents without a companion Bluesky post

### LINKED ISSUE #210: Feature Request: Publish to standard.site only, without a companion Bluesky post
### What

Add a filter that lets a site publish its posts as `site.standard.document` records only, skipping the companion `app.bsky.feed.post` that ATmosphere normally creates alongside each document.

```php
add_filter( 'atmosphere_should_publish_bluesky_post', '__return_false' );
```

With the filter on, every publish path — the `wp atmosphere backfill` command, auto-publishing of new posts, and edits to existing ones — writes only the document record. No UI changes; it's purely a developer-facing filter.

This is a follow-up to #91, which added the `wp atmosphere backfill` command. That command is what makes seeding an existing site's back catalogue practical; this filter is what makes doing so without flooding a Bluesky account practical.

### Why

Today ATmosphere always pairs each document with a Bluesky post. That's the right default for people who want to cross-post, but it doesn't fit everyone.

Some folks want their content available as standard.site documents — so it's fully readable in tools like Standard Reader — without turning their Bluesky account into a feed of every blog post they've ever written. That's especially true when backfilling an existing site: nobody wants to dump hundreds of old posts into their followers' timelines just to seed the document records.

This filter gives them a clean way to run ATmosphere as a standard.site publishing layer and keep Bluesky quiet.

### How

Gated by a new `is_bluesky_post_enabled()` helper backed by the `atmosphere_should_publish_bluesky_post` filter (default `true`). When it returns `false`, `Publisher::publish_post()` and `Publisher::update_post()` route to document-only create/update paths. It's forward-only — it governs new writes and doesn't remove Bluesky posts published before the filter was enabled.

jeherve:
Fixes #210

## Proposed changes:

Today ATmosphere writes two records for every post: a `site.standard.document` and a companion `app.bsky.feed.post`. That's the right default for cross-posting, but it forces a Bluesky post on people who only want the standard.site document — and it makes seeding an existing site with `wp atmosphere backfill` (#91) impractical, since backfilling would dump every old post into followers' timelines.

This adds a filter, `atmosphere_should_publish_bluesky_post` (default `true`). Return `false` and the plugin writes only the document record — no companion Bluesky post — across every publish path: the backfill command, auto-publishing of new posts, and edits to existing ones.

Under the hood it's a new `is_bluesky_post_enabled()` helper backing the filter. When it's off, `Publisher::publish_post()` and `Publisher::update_post()` route to document-only create/update paths.

It's deliberately **forward-only**: it governs new writes and does not delete Bluesky posts published before the filter was turned on. It makes **no UI changes** — it's a developer-facing filter, so someone running ATmosphere as a standard.site publishing layer can wire it up without touching the settings screen. Reaction sync, reply sync, and comment publishing simply stay inert, since there's no Bluesky post to attach them to.

### Other information:

## Testing instructions:

1. Connect the plugin to an AT Protocol account as usual.
2. Add the filter to a mu-plugin or your theme's `functions.php`:
   ```php
   add_filter( 'atmosphere_should_publish_bluesky_post', '__return_false' );
   ```
3. Publish a new post. Confirm in the AT Protocol repo that a `site.standard.document` record was created and **no** `app.bsky.feed.post` record was.
4. Backfill existing posts, previewing first:
   ```
   wp atmosphere backfill --limit=5 --dry-run
   wp atmosphere backfill --limit=5
   ```
   Confirm each processed post gets a document record only, and that a second `wp atmosphere backfill` run reports them skipped (already synced).
5. Edit one of those posts and confirm the document is updated in place, still with no Bluesky post.
6. Remove the filter, publish another post, and confirm the default returns: both a document and a Bluesky post are written.

Automated coverage: `composer lint` is clean and the full PHPUnit suite passes (1059 tests). New tests cover the publish, update, delete-cleanup, and backfill paths in both filter states.

### Changelog entry

A changelog entry is already committed on the branch (`.github/changelog/add-document-only-publishing`), so the auto-create box below is intentionally left unchecked.


includes/class-publisher.php
```diff
@@ -862,6 +928,62 @@ public static function update_post( \WP_Post $post ): array|\WP_Error {
...
+			// Never seeded (or a prior create failed before storing a URI):
+			// create the document fresh, reusing any reserved TID via get_rkey().
+			//
+			// Rare edge: if a prior #create actually committed on the PDS but
+			// its URI was never stored (a malformed applyWrites response), this
+			// re-issues a #create against the same reserved TID, which the PDS
+			// rejects as already-existing — surfacing a WP_Error. That mirrors
+			// the dual-record update_post() tolerance for half-synced state.
+			return self::publish_document_only( $post );
+		}
```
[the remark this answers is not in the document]
jeherve: Good catch — fixed in e2f9b19. `update_document_only()` now only falls back to a fresh create when there is no stored URI at all. A stored URI with a missing TID is treated as the same corrupted half-synced state the dual-record `update_post()` path handles, returning an `atmosphere_missing_tid` error instead of minting a second document that would orphan the existing record. Added a regression test (`test_update_document_only_errors_when_uri_present_but_tid_missing`) covering it.

tests/phpunit/tests/cli/class-test-backfill-command-invoke.php
```diff
@@ -393,4 +393,54 @@ public function test_not_connected_midrun_aborts_remaining() {
...
 	}
+
+	/**
+	 * Backfill with the Bluesky companion disabled writes documents only and
+	 * marks each post synced so a second run skips it.
+	 */
+	public function test_backfill_document_only_marks_posts_synced() {
+		\add_filter( 'atmosphere_should_publish_bluesky_post', '__return_false' );
+
+		$collections = array();
```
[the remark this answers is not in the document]
jeherve: Fixed in 4d8a879. Moved the `atmosphere_should_publish_bluesky_post` cleanup into `tear_down()` (which already clears `atmosphere_pre_apply_writes`), so it always runs even if an assertion fails mid-test. Dropped the now-redundant manual `remove_all_filters()` calls at the end of the method.

includes/functions.php
```diff
@@ -954,6 +954,43 @@ function is_publication_sync_enabled(): bool {
...
+ *
+ * Forward-only: it governs new writes and does not remove Bluesky posts
+ * published before it was enabled.
+ *
+ * Unlike the other lane helpers (see {@see is_publication_sync_enabled()}), this
+ * one has no connection-only pass: it shapes *what* a publish writes, not
+ * *whether* the site publishes, and connection-only mode already disables
+ * publishing wholesale upstream — so the lane never runs there and there is
+ * nothing to force off. It is deliberately a pure filter, not the
+ * "option → force off in connection-only → filter last" contract.
```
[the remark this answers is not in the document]
jeherve: You're right — reworded in 286ef05. Confirmed the Backfill CLI calls `Publisher::publish_post()` / `update_post()` directly (not through `is_auto_publish_enabled()`), and `Publisher` gates only on `is_bluesky_post_enabled()`, so this lane does run in connection-only mode. The docblock now explains the real reason it has no connection-only pass: it shapes *what* a publish writes, not *whether* the site publishes, so a connection-layer host running a manual backfill can still choose document-only output — forcing it off there would remove that choice.

pfefferle:
**Both document-only paths skip `reconcile_post_after_write()`**

The doc-only lane reads cleanly and the create / update / delete lifecycle all line up. One gap I wanted to flag rather than assume is deliberate:

The document-only branches return without going through `reconcile_post_after_write()`:

- **publish** — `Publisher::publish_post()` returns at the `if ( ! is_bluesky_post_enabled() )` branch (right after `publish_document_only()` + the result action), while the dual-lane path calls `reconcile_post_after_write()` before returning.
- **update** — `Publisher::update_post()` returns `update_document_only( $post )` directly, while `update_single()` and `update_thread_in_place()` both end with `reconcile_post_after_write()`.

`reconcile_post_after_write()` is the post-write race guard: after `applyWrites` commits, it invalidates the object cache, re-reads the post, and issues a compensating `delete_post()` if the post became private or password-protected while the write was in flight. Because the doc-only paths skip it, a document-only publish or edit that races a visibility change would leave the `site.standard.document` record live on the PDS for content that is no longer publishable.

The branch comment enumerates what it intends to skip (bsky post, strongRef precompute, thread machinery) but doesn't mention reconciliation, so this reads as an oversight rather than an intentional exclusion. `reconcile_post_after_write()` looks safe to call here — it delegates to the meta-driven `delete_post()`, which already deletes doc-only records correctly (empty thread records + present document TID → one document delete).

The fix would be to route the doc-only paths through the same reconcile tail instead of early-returning: compute `$result` from whichever lane ran, then fall through the shared `reconcile_post_after_write()` + result-action tail so a mode can't miss it.

pfefferle:
Did a deeper high-effort pass (following up on the reconcile comment above). The recurring theme is that the two document-only paths are modeled as early-return special cases rather than composing with the shared publish/update lifecycle, so each hardening the dual-record paths earned is individually missing on the doc-only side. Findings below, most consequential first — I think 1–3 are merge-blockers.

### 1. Document-only posts are undiscoverable via HTML (`output_document_link()` keys on the wrong meta)

`publish_document_only()` → `store_document_meta()` writes `Document::META_URI`/`META_CID` but never `Post::META_URI`. The front-end discovery tag gate reads the **Bluesky post** URI:

```php
// Atmosphere::output_document_link(), class-atmosphere.php:446-449
$bsky_uri = get_post_meta( $post->ID, Post::META_URI, true );
if ( empty( $bsky_uri ) ) {
    return;
}
```

So on a document-only site (the intended standard.site-publication use case) the `<link rel="site.standard.document">` tag is never printed, and resolvers/visitors loading the article can't discover the record. The document is published but effectively undiscoverable via HTML — which defeats the feature's purpose.

### 2. Editing a legacy (never-synced) post silently mints a fresh record

With doc-only mode on and auto-publish enabled, editing a post that predates the ATmosphere connection routes `update_post()` → `update_document_only()`; `$doc_uri` is empty, so it falls through to `publish_document_only()` and creates a brand-new `site.standard.document` on the PDS.

The same edit in normal dual mode is a deliberate no-op: `update_post()` checks `Post::META_TID`, finds no prior publish, fires `atmosphere_update_skipped_unsynced_post`, and returns `array()` — the documented protection against turning routine edits of old content into fresh records (retro-sync is meant to require the backfill command). Doc-only mode loses that guard and auto-publishes content the author never intended to share.

### 3. Posts created in document-only mode can never gain a Bluesky companion afterward

Once a post has `Document::META_URI` but no `Post::META_TID`, re-enabling cross-posting (`atmosphere_should_publish_bluesky_post => true`) doesn't recover it:

- **Edit path:** `has_post_records()` (class-atmosphere.php:1023-1026) sees `Document::META_URI` and classifies the change as an update → `update_post()` → normal path where `stored_thread_records()` is empty and `Post::META_TID` is unset → fires `atmosphere_update_skipped_unsynced_post` and returns without publishing.
- **Backfill:** `class-backfill.php:139` treats a present `Document::META_URI` as already-synced and skips it.

So every post created during document-only mode is permanently stuck with no Bluesky post and no recovery path.

### 4. Both doc-only paths skip `reconcile_post_after_write()`

(Already covered in the comment above — flagging here only so the full picture is in one place.)

### 5. Seeding a doc-only post via an edit skips `atmosphere_publish_post_result`

When a never-published post is first written by an edit, `update_document_only()` sees no `Document::META_URI` and calls the private `publish_document_only()` directly, which fires no action. In normal mode the analogous transition delegates to `publish_post()`, which fires `atmosphere_publish_post_result` exactly once. Metrics/notification subscribers therefore miss the first publish of doc-only posts seeded through an edit.

### Cleanup

- **Duplication:** `publish_document_only()` and `update_document_only()` repeat the same `$writes` array + `apply_writes` → `is_wp_error` → `store_document_meta( …, 0 )` tail, differing only in the op and rkey source. Worth one shared single-document writer so future changes don't have to be made in two places. (I have a small `write_document_only()` extraction ready if useful.)
- **Comment style:** the `if ( ! $doc_uri )` and `if ( ! $doc_tid )` blocks in `update_document_only()` stack consecutive `//` lines for multi-line paragraphs; `docs/php-coding-standards.md:311` asks for `/* */` block comments there.
- **Changelog:** "Allow sites to publish posts to standard.site only, without a companion Bluesky post." leads with `standard.site` — this surfaces on the WP update screen, where the convention is plain language / lead-with-Bluesky.

The deeper fix for 1–5 is the same one: rather than five separate patches, route the document-only paths through the shared publish/update machinery (the legacy-post guard, the reconcile tail, the result action, the `Post::META_URI`/discovery contract, and backfill's synced-check) so "one lane off" composes with the lifecycle instead of re-implementing a thinner copy of it.

includes/class-publisher.php
```diff
@@ -862,6 +928,76 @@ public static function update_post( \WP_Post $post ): array|\WP_Error {
...
+
+		if ( ! $doc_uri ) {
+			// Never seeded (or a prior create failed before storing a URI):
+			// create the document fresh, reusing any reserved TID via get_rkey().
+			//
+			// Rare edge: if a prior #create actually committed on the PDS but
+			// its URI was never stored (a malformed applyWrites response), this
+			// re-issues a #create against the same reserved TID, which the PDS
+			// rejects as already-existing — surfacing a WP_Error. That mirrors
+			// the dual-record update_post() tolerance for half-synced state.
```
pfefferle: These multi-line paragraphs should be `/* */` block comments per `docs/php-coding-standards.md`.

```suggestion
			/*
			 * Never seeded (or a prior create failed before storing a URI):
			 * create the document fresh, reusing any reserved TID via get_rkey().
			 *
			 * Rare edge: if a prior #create actually committed on the PDS but
			 * its URI was never stored (a malformed applyWrites response), this
			 * re-issues a #create against the same reserved TID, which the PDS
			 * rejects as already-existing — surfacing a WP_Error. That mirrors
			 * the dual-record update_post() tolerance for half-synced state.
			 */
```

includes/class-publisher.php
```diff
@@ -862,6 +928,76 @@ public static function update_post( \WP_Post $post ): array|\WP_Error {
...
+			return self::publish_document_only( $post );
+		}
+
+		if ( ! $doc_tid ) {
+			// A stored URI with no TID is a corrupted record: falling back to a
+			// fresh create would mint a second document and orphan the existing
+			// one. Surface the same error the dual-record update_post() raises.
```
pfefferle: Same here — `/* */` block comment.

```suggestion
			/*
			 * A stored URI with no TID is a corrupted record: falling back to a
			 * fresh create would mint a second document and orphan the existing
			 * one. Surface the same error the dual-record update_post() raises.
			 */
```

includes/class-publisher.php
jeherve: Converted to a `/* */` block comment in 7ec706d (the surrounding `! $doc_uri` branch was rewritten there for finding 2 anyway).

includes/class-publisher.php
jeherve: Converted to a `/* */` block comment in 7ec706d.

jeherve:
Thanks for the high-effort pass — this was exactly the kind of lifecycle-composition review the doc-only lane needed. Went through targeted fixes rather than the full unification (kept the blast radius off the well-tested dual path); rundown below.

**1. Document-only posts undiscoverable via HTML — fixed in bbe7e3f.**
`output_document_link()` now gates on the document record’s own `Document::META_URI` instead of the Bluesky post’s `Post::META_URI`. That URI is present after any real document publish (dual or doc-only) and is exactly what the tag advertises, so standard.site-only publications now emit `<link rel="site.standard.document">`. Bonus: the earlier return means unpublished posts no longer reach the lazy-mint path at all. Added a doc-only regression test and rounded out the dual test’s fixture.

**2 & 5 collapse into one fix — fixed in 7ec706d.** They were the same buggy call: `update_document_only()` falling to `publish_document_only()` on `! $doc_uri`. It now mirrors the dual guard — fires `atmosphere_update_skipped_unsynced_post` and no-ops — so a routine edit of legacy content never retro-syncs (that stays the backfill command’s job). One nuance worth flagging: I deliberately did *not* use `Document::META_TID` as the "was published" marker the way the dual path uses `Post::META_TID`, because `output_document_link()` mints it speculatively on a pageview — only a stored URI proves a real publish. Genuine new posts still seed through `publish_post()` (which fires the result action, resolving #5), and failed creates retry through the publish ladder, so nothing is lost by skipping. Updated the test that had encoded the old mint-on-edit behavior.

**4. Skipped `reconcile_post_after_write()` — fixed in 9e2ee19.** Both doc-only paths now run the same reconcile tail the dual paths end with, so a document publish/edit that races a visibility change is cleaned up. Added a regression test (doc-only publish racing a mid-flight password-protect).

**3. No recovery to a Bluesky companion — not fixed here; proposing a follow-up.** This is the one I couldn’t address cleanly as a targeted patch. A real recovery path needs `has_post_records()`/backfill to stop treating a lone `Document::META_URI` as fully synced *and* the publish path to add a companion to an existing document without re-`#create`-ing it (which would duplicate the record). That’s exactly the shared-machinery composition you sketched — doing it piecemeal means touching the dual path’s `has_post_records()` and the publish flow with real duplicate-record risk. Given the filter is explicitly forward-only, I’d rather treat "posts published while doc-only stay doc-only" as documented semantics and track the recover-on-re-enable path as a dedicated follow-up (happy to take the `write_document_only()` seed you offered as the starting point). Let me know if you’d prefer to block on it instead.

**Cleanup:**
- Duplication → extracted `write_document_only( $post, $op, $rkey )` in 51b591d; `publish_document_only()`/`update_document_only()` are now thin wrappers over it.
- Comment style → the two `//` paragraphs are now `/* */` blocks (7ec706d).
- Changelog → reworded to lead with plain language in 51b591d.

Full suite green (1062 tests). Rebased onto the trunk merge before pushing.

pfefferle:
Thanks for the fixes. The discovery link, the reconcile guard, the legacy-edit skip and the shared writer all look right to me, and the suite stays green.

One from the earlier list is still open, and I think it's more a question than a clear bug: a post published while doc-only mode was on can't get a Bluesky post later. Backfill counts it as synced because `Document::META_URI` is set (`class-backfill.php:139`), and an edit routes to the dual path, finds no `Post::META_TID`, and skips. So once someone re-enables Bluesky there's no way to fill in the missing post.

Might be fine, doc-only is doc-only, and "add a Bluesky post to an already-published document" is arguably a different job than backfill. I just want to make sure it's a decision and not a surprise later.

.github/changelog/add-document-only-publishing
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: added
+
+Allow sites to publish to Bluesky as documents only, without also adding a post to the Bluesky feed.
```
pfefferle: This reads like it posts to Bluesky, but doc-only mode skips the Bluesky feed and only writes the standard.site document. Maybe something like:

```suggestion
Allow sites to publish posts as standard.site documents without also posting to Bluesky.
```

pfefferle (approved):

PR #214, opened 2026-07-24
Distinguish upstream resolver failures and add a set_identity() helper

### LINKED ISSUE #86: Resolver and identity-write follow-ups from FOSSE bidirectional verification work
Two small follow-ups surfaced while working on the FOSSE-side recovery flow for the `atmosphere_identity`-wipe regression (paired with PR 85). Filing them together since they share the same code area but they're independent.

## 1. `Resolver::resolve_did()` doesn't check the HTTP status code before parsing the body

In `includes/oauth/class-resolver.php` (around lines 95-112), `resolve_did()` calls `wp_safe_remote_get()`, checks `is_wp_error()` for transport failures, then immediately `json_decode`s the body without first inspecting `wp_remote_retrieve_response_code()`. The same gap exists in `discover_auth_server()` (around lines 186-234).

Consequence: a 500 / 503 / 429 from plc.directory or from a `did:web` host returns a non-null body (HTML error page, rate-limit JSON without an `id` key, etc.), which falls through to the `atmosphere_invalid_did_doc` / `Invalid DID document received` branch. The caller surfaces "Could not fetch the DID document: Invalid DID document received" — indistinguishable from a genuinely malformed document — with no signal that the upstream service was down or rate-limiting.

Suggested fix: after the `is_wp_error()` guard, read `$status = wp_remote_retrieve_response_code( \$response )` and return a distinct `WP_Error` for non-2xx before decoding the body. For 429 include the `Retry-After` header in the message. For 5xx use a code like `atmosphere_plc_upstream_error` so callers can distinguish transient from structural failures. The 10s timeout is correctly set; only the status-code discrimination is missing.

Same fix in `discover_auth_server()` for both the `/.well-known/oauth-protected-resource` and `/.well-known/oauth-authorization-server` fetches.

## 2. No `Atmosphere\set_identity()` helper

`includes/functions.php` exposes `get_identity()`, `has_identity()`, `get_did()`, and `get_pds_endpoint()` as the canonical read surface for `atmosphere_identity`. There's no public write helper.

Consequence: any consumer that needs to write identity from outside the OAuth token-exchange path (FOSSE's identity-recovery escape hatch is the immediate case) ends up calling `update_option( 'atmosphere_identity', \$identity, true )` directly. The `autoload=true` parity with `get_identity()`'s lazy-migration write is load-bearing — if Atmosphere ever changes the option shape, autoload flag, or migration semantics, every external caller has to be re-audited.

Suggested API: `Atmosphere\set_identity( array \$identity ): bool` that:
- validates `did`/`handle`/`pds_endpoint` are present and well-formed
- writes the option with the same autoload/shape contract as the lazy-migration path
- returns true on persisted, false on validation or write failure
- fires a `do_action( 'atmosphere_identity_set', \$identity )` action so consumers can react

This keeps the option's storage shape an Atmosphere implementation detail rather than something every domain-handle wrapper has to mirror by hand.

---

Both are low-priority. Filing as a single issue since they affect the same caller flow and surfaced together. Happy to PR either separately if useful.

Context links:
- FOSSE PR 175 (the recovery escape hatch that exposed both gaps): https://github.com/Automattic/fosse/pull/175
- Upstream PR 85 (the disconnect-preserves-identity fix, paired with this work): https://github.com/Automattic/wordpress-atmosphere/pull/85

pfefferle:
Two small, independent follow-ups from #86, sharing the same code area.

## 1. Distinguish upstream fetch failures in the resolver

`resolve_did()` and `discover_auth_server()` parsed the response body without ever checking the HTTP status, so a `500`/`503`/`429` from plc.directory or an auth server fell through to the generic "Invalid DID document" / "PDS did not advertise an authorization server" branch. A user seeing "Could not fetch the DID document: Invalid DID document received" during a plc.directory outage has no way to tell a temporary blip from a genuinely broken account.

A shared `upstream_error()` helper now rejects any non-2xx before the body is parsed, at all four fetch sites, including `handle_to_did()`'s well-known lookup. That one points at the user's own domain, so it is the hop most likely to sit behind a CDN answering 503 or a WAF answering 429, and it is reached before plc.directory. A 404 there still falls through to the existing "could not resolve handle" message, since a host with no well-known file has simply not set handle verification up, which is a different problem from an outage.

* `atmosphere_upstream_rate_limited` for `429`, with the `Retry-After` value in the message when it is a plain number of seconds.
* `atmosphere_upstream_error` for other non-2xx.
* Both carry the status as `get_error_data()['upstream_status']`. Deliberately not `status`: that key is what `rest_convert_error_to_response()` reads to set the REST response's own code, and an upstream `401` from an attacker-chosen host must not become the site's API status. No caller branches on it today, so the improvement users actually see is the message.

## 2. `set_identity()` write helper

`get_identity()` / `has_identity()` / `get_did()` / `get_pds_endpoint()` are the read surface for `atmosphere_identity`, but there was no write helper, so a consumer writing identity from outside the OAuth token exchange had to call `update_option( 'atmosphere_identity', ..., true )` by hand and mirror the shape and autoload flag.

`set_identity( array $identity ): bool` is a thin wrapper that normalizes to `{did, handle, pds_endpoint}` and writes with `autoload=true`, so that contract lives in one place. `Client` and `Handle` both route through it now, so it has real consumers rather than only the lazy migration.

It replaces rather than merges, and the docblock leads with that: passing only `handle` clears the DID, which takes `has_identity()` false and stops `/.well-known/atproto-did` answering. Non-scalar values are dropped rather than cast, so nothing can store the literal `"Array"` and have it read as a live identity.

## Testing instructions

* `composer lint && npm run env-test` — full suite passes (1130 tests).
* Resolver: `npm run env-test -- --filter=Test_Resolver` — covers a `5xx` on `resolve_did`, a `429` with `Retry-After`, `5xx` on both `discover_auth_server` fetches, the three `handle_to_did()` paths (`5xx`, `404` fall-through, and a `did:`-shaped error body), and the `Retry-After` values that must never reach the message.
* `set_identity`: `npm run env-test -- --filter=Test_Functions`.

To see the escaping fix by hand, connect to a handle whose host answers `429` with `Retry-After: <script>alert(1)</script>`. The settings notice should say it was rate limited with no value and no markup, since `settings_errors()` prints that message unescaped.

Closes #86.

### Changelog entry

Two entries. `.github/changelog/fix-resolver-upstream-status` (Patch / Fixed) for the clearer message, now worded to include the handle verification hop against the user's own domain.

`.github/changelog/change-mention-resolution-error-codes` (Patch / Changed) for extenders: the `handle_to_did()` gate changes which error code `atmosphere_mention_didweb_fallback` receives for an HTTP-level failure. Default behavior does not change, but the filter has been published since 2.0.0.

Item 2 is a developer helper with no user-facing behavior, so it has no separate line.


includes/oauth/class-resolver.php
```diff
@@ -110,6 +117,61 @@ public static function resolve_did( string $did ): array|\WP_Error {
...
+	 */
+	private static function upstream_error( array $response, string $label ): ?\WP_Error {
+		$status = (int) \wp_remote_retrieve_response_code( $response );
+
+		if ( is_success_status( $status ) ) {
+			return null;
+		}
+
+		if ( 429 === $status ) {
+			$retry_after = \wp_remote_retrieve_header( $response, 'retry-after' );
```
kraftbj, line 186: The raw header value goes straight into the message on the next line, and the message ends up somewhere core prints unescaped (`settings_errors()`, via `Sanitize::handle()`'s `add_settings_error()` call). Since the handle picks the host, whoever runs that host controls this string, and it lands in an admin screen as HTML.

Same line also has the array case Copilot raised: `wp_remote_retrieve_header()` returns an array for a repeated header, so `sprintf()` gets an array, PHP warns "Array to string conversion", and the message reads "Retry after Array."

Both go away with one normalization:

```php
$retry_after = \wp_remote_retrieve_header( $response, 'retry-after' );
$retry_after = \is_numeric( $retry_after ) ? (int) $retry_after : 0;
```

then `%2$d` in the format string, and `0` falls through to the no-value variant you already have. That also fixes the cosmetic `Retry-After: 0` case, where the string `'0'` is falsy today and silently drops the value.

If HTTP-date values should still display, `Atmosphere\sanitize_text()` plus a length cap works too, which is what `class-atmosphere.php` line 2601 already does for other remote-sourced error text.

Whichever way it goes, the test wants a hostile value rather than `'30'`, since the current `assertStringContainsString( '30', ... )` would pass on `<script>30</script>`.

includes/oauth/class-resolver.php
```diff
@@ -110,6 +117,61 @@ public static function resolve_did( string $did ): array|\WP_Error {
...
+	 * `get_error_data()['status']` to tell a transient 5xx / 429 from a
+	 * structural 4xx. Returns null for a 2xx response so the caller parses.
+	 *
+	 * @since unreleased
+	 *
+	 * @param array  $response `wp_remote_*` response array.
+	 * @param string $label    Human label for the fetched resource.
+	 * @return \WP_Error|null Error for a non-2xx status, null otherwise.
+	 */
+	private static function upstream_error( array $response, string $label ): ?\WP_Error {
```
kraftbj, line 161: The helper is applied at three call sites, but the resolver makes four HTTP calls. `handle_to_did()` line 48 fetches `https://{$handle}/.well-known/atproto-did` and still goes straight to the body.

That's the hop I'd have expected to want this most: it's aimed at the user's own domain, so it's the one sitting behind a CDN that throws 502/503 or a WAF that throws 429. A custom-domain user hits the misleading "Could not resolve handle" message before ever reaching plc.directory, so for that population the fix doesn't land. There's also a smaller correctness wrinkle: a non-2xx body that happens to start with `did:` is currently accepted as a DID.

Same shape as the others, after the `is_wp_error()` check at line 55:

```php
$upstream_error = self::upstream_error( $response, \__( 'handle verification document', 'atmosphere' ) );
if ( $upstream_error ) {
	return $upstream_error;
}
```

One caveat worth deciding on: a host with no well-known file returns 404, and "handle not configured" is genuinely a different message than "upstream is down." So you may want 404 to keep falling through to `atmosphere_resolve_handle` rather than becoming "could not be fetched (HTTP 404)."

Happy for this to be a follow-up, but then the PR body's "all three fetch sites" is describing the two functions in scope rather than the class of bug, and it'd be good to say so.

includes/oauth/class-resolver.php
```diff
@@ -110,6 +117,61 @@ public static function resolve_did( string $did ): array|\WP_Error {
...
 
+	/**
+	 * Reject a non-2xx HTTP response before its body is parsed.
+	 *
+	 * A fetch to plc.directory, a did:web host, or an auth server can answer
+	 * with a 500 / 503 / 429 whose body still decodes — an HTML error page,
+	 * or rate-limit JSON with no `id` — which would otherwise fall through
+	 * to the generic "invalid document" branch and hide a transient upstream
+	 * failure behind a malformed-document message. Callers can read
+	 * `get_error_data()['status']` to tell a transient 5xx / 429 from a
```
kraftbj: This claim doesn't hold on the one path that can act on it today.

`Resolver::resolve()` is reachable from exactly one place, `Client::authorize()`, which has two callers: `Sanitize::handle()` line 65 and `Connection_Controller::authorize()` line 169. Neither reads `get_error_data()['status']` and neither branches on the new codes. On the REST one it's worse than unused: line 172 unconditionally does

```php
$url->add_data( array( 'status' => 400 ) );
```

and `WP_Error::add_data()` replaces the primary data for that code (the old value gets pushed to `additional_data`, which `get_error_data()` doesn't return). So the 429 or 503 you just attached is gone by the time it reaches the client.

The upside is that this clamp is also what prevents the nastier version of this: `status` is the key `rest_convert_error_to_response()` reads to set the response's own HTTP status, so without the clamp an upstream 401 or 403 from an attacker-chosen host would become the site's REST API status. Which is the real argument for either naming the key `upstream_status`, or keeping `status` and being deliberate about it.

Two ways out, either is fine:

- Have `Connection_Controller::authorize()` pass a resolver-supplied 429/503 through instead of hardcoding 400 (and `Sanitize::handle()` pick a distinct settings-error class), so the discriminator actually does something.
- Or drop the "callers can read `get_error_data()['status']`" sentence and the PR body's transient-vs-structural framing, since right now only the message text differs.

The changelog line is accurate either way, the message improvement genuinely reaches users through both surfaces.

includes/functions.php
```diff
@@ -460,6 +460,36 @@ function has_identity(): bool {
...
+ *                        `pds_endpoint` are persisted; a missing key is
+ *                        stored as an empty string and any other keys are
+ *                        dropped.
+ * @return bool Whether the option was updated (see `update_option()`).
+ */
+function set_identity( array $identity ): bool {
+	return \update_option(
+		'atmosphere_identity',
+		array(
+			'did'          => (string) ( $identity['did'] ?? '' ),
```
kraftbj: `?? ''` catches null and missing keys, which means a partial call quietly blanks whatever it didn't mention. `set_identity( array( 'handle' => 'new.example.com' ) )` reads a lot like "update the handle" given the name and the way this mirrors `get_identity()`, but it zeroes the DID. Then `has_identity()` goes false, `/.well-known/atproto-did` starts 404ing, and the publication link tag stops advertising, which is exactly the bidirectional-verification anchor PR 85 went to the trouble of preserving across disconnect.

That matters more than usual here because the docblock invites third-party callers in, and a recovery flow is precisely the caller most likely to have a partial identity in hand.

Either merge against the stored value for keys that are absent, or make the docblock summary say "replaces" rather than leaving it to the param description. The one-line summary is what a third-party caller reads.

Separately on this line: `(string)` on an array or object warns and stores the literal `"Array"`, which `has_identity()` then treats as a live identity and `class-atmosphere.php` will happily serve from the well-known endpoint. No current caller can do that, but "no validation" plus "documented for external callers" is the combination that makes it worth a `\is_scalar()` guard.

includes/functions.php
```diff
@@ -460,6 +460,36 @@ function has_identity(): bool {
...
+ *
+ * @since unreleased
+ *
+ * @param array $identity Identity to store. Only `did`, `handle`, and
+ *                        `pds_endpoint` are persisted; a missing key is
+ *                        stored as an empty string and any other keys are
+ *                        dropped.
+ * @return bool Whether the option was updated (see `update_option()`).
+ */
+function set_identity( array $identity ): bool {
```
kraftbj, line 488: Minor, and I see the PR body already calls this out as a follow-up, so mostly registering the cost of deferring it.

As of this PR the only caller is `get_identity()`'s own migration, while the two writers the docblock is describing still hand-roll it: `class-client.php` line 650 and `class-handle.php` line 315. So "the option's shape and its autoload flag live in one place" isn't true yet, and anything added here later (validation, an action, a normalization) silently won't apply to the two paths that actually write identity.

The migration is small enough that I'd lean toward doing it here: Handle's is a literal function-name swap, and Client's collapses a 9-line `update_option()` into one call. Nothing diverges functionally today since both already write the same three keys, so it's cheap now and gets more expensive the moment someone adds a rule to the helper.

Also worth a docblock tweak while you're in here: the `bool` return is a straight `update_option()` pass-through, so it's `false` both when the write failed and when the stored value was already identical. "Whether the option was updated" is accurate but easy to read as success/failure, and a future `if ( ! set_identity( $x ) ) { bail; }` would bail on a perfectly good no-op write. The new test only covers the first insert after `delete_option()`, so nothing pins that today.

kraftbj (commented):
Ran a multi-reviewer pass over this branch (correctness, security, adversarial, reliability, testing, maintainability, API-contract, plus a Codex second opinion). Two independent things came out of it: one that I think should land before merge, and three that decide whether the PR delivers what it says on the tin.

**The one to fix first: the `Retry-After` value reaches wp-admin unescaped.**

This is the first resolver message to interpolate remote-controlled *text* into a `WP_Error` (the generic branch is fine, it formats the status with `%2$d`). I traced the sink hop by hop against core rather than assuming:

`Resolver::upstream_error()` -> `Client::authorize()` returns it verbatim -> `Sanitize::handle()` line 68 `add_settings_error( 'atmosphere', 'auth_failed', $auth_url->get_error_message() )` -> `wp-admin/options.php` line 370 stores it in the `settings_errors` transient -> `options-head.php` calls `settings_errors()` -> `wp-admin/includes/template.php` line 2013:

```php
$output .= "<p><strong>{$details['message']}</strong></p>";
```

No escaping. The host whose header this is gets chosen by the submitted handle, so a did:web host, a PDS, or an auth server can put markup into an authenticated `manage_options` screen. It needs an admin to connect to a hostile handle and the host to answer 429, so it's not trivially reachable, but the payoff is script execution in wp-admin.

Worth noting: Copilot flagged the array-to-string half of this same line, and that thread is marked resolved, but the 429 block is byte-identical from `9b91de8` through `d117cb2`. No fixing commit ever landed. One normalization closes both.

**The two that decide whether the fix lands where it's aimed.**

The `['status' => N]` payload gets thrown away on the one path that could use it, and nothing anywhere branches on the new codes, so today only the message string changed. And `handle_to_did()`'s well-known fetch is a fourth fetch site that didn't get the gate. That's the hop pointed at the user's own domain, which is the one most likely to be sitting behind a CDN throwing 503 or a WAF throwing 429. So for custom-domain users the changelog's promise doesn't land at the first thing they hit. Details inline.

**What came back clean**, so nobody re-checks it: nothing in `includes`, `src`, `templates`, or `integrations` switches on `atmosphere_invalid_did_doc` / `atmosphere_no_auth_server` / `atmosphere_no_pds`, so no caller contract broke and no retry or negative-caching logic moved. No cron path reaches the resolver. `is_success_status()` is 200-299 and `wp_safe_remote_get()` follows redirects, so 3xx is a non-issue. No new SSRF or oracle surface (both entry points need `manage_options`). `set_identity()`'s unvalidated `pds_endpoint` is not the API request target, `class-api.php` line 68 reads `atmosphere_connection`. The Connectors card renders messages as React text children, so that path is genuinely safe. And the new tests are load-bearing: removing any of the three gate calls fails its matching test, and `tear_down()` clears the stubs.

There are a few smaller things (the bool return being `false` for both "failed" and "unchanged", the `(string)` cast on a non-scalar, 429 being detected and then discarded with no backoff) that I left off since they're your call and none of them block.

Testing gap worth calling out either way: `test_resolve_did_surfaces_rate_limit_with_retry_after` asserts `assertStringContainsString( '30', ... )`, which passes just as happily on a raw `<script>30</script>` header. That assertion can't tell a sanitized value from a raw one.

includes/oauth/class-resolver.php
pfefferle: Fixed with your normalization, thanks for tracing the sink. `is_numeric()` to an int, `%2$d`, and 0 falls through to the no-value variant, so the array case and `Retry-After: 0` go with it. Test now feeds `<script>alert(1)</script>` and an HTTP-date and asserts neither reaches the message.

includes/oauth/class-resolver.php
pfefferle: Done here rather than as a follow-up, since you are right that it is the hop custom-domain users hit first. 404 falls through to `atmosphere_resolve_handle` as you suggested. Also covers the wrinkle you spotted: a 500 body starting with `did:` is no longer accepted as a DID.

includes/oauth/class-resolver.php
pfefferle: Went with the rename rather than passing the status through, because of the hazard you pointed out: `status` is what `rest_convert_error_to_response()` reads, so an upstream 401 from a host the handle picked would become our own REST status. It is `upstream_status` now, with a test asserting `status` is absent, and the docblock says plainly that nothing branches on it today and the message is what users get. Dropped the transient-vs-structural framing from the PR body too.

includes/functions.php
pfefferle: Kept replace semantics rather than inventing merge, but the summary line leads with it now and spells out the consequence, since that is what a third-party caller reads. Added the `is_scalar()` guard, and both hazards have tests.

includes/functions.php
pfefferle: Migrated both in this PR. Client's nine-line `update_option()` is now one call and Handle's was the function-name swap you expected. Also fixed the `@return` to say `false` means both failure and no-op.

includes/functions.php
```diff
@@ -460,6 +460,53 @@ function has_identity(): bool {
...
+ * The canonical write surface for `atmosphere_identity`, mirroring the
+ * read helpers ({@see get_identity()} and friends). A consumer that writes
+ * identity from outside the OAuth token exchange — a recovery or
+ * escape-hatch flow — should call this rather than `update_option()`
+ * directly, so the option's shape and its autoload flag (which
+ * {@see get_identity()}'s lazy migration also relies on) live in one place.
+ *
+ * @since unreleased
+ *
+ * Replaces the stored identity outright. It is not a partial update: a key
```
kraftbj: This paragraph is doing the right work but it's below the tag block, so it doesn't land where you meant it to.

PHPDoc stops treating text as description once tags start, so everything from here down reads as a continuation of the `since` tag's value rather than as the function description. IDEs, phpDocumentor, and PHPStan will all attribute it there. PHPCS doesn't check for this, which is why lint stayed green.

The practical cost is the thing the paragraph exists to prevent. A third-party caller skimming the summary reads "Persist the AT Protocol identity" and then "the canonical write surface... should call this rather than `update_option()` directly" and stops, because that's the pitch. The replace warning is past the tags, which is where people stop reading.

Moving it up so it sits right after the summary line, before the `since` tag, fixes both. If it's the first thing after "Persist the AT Protocol identity (DID, handle, PDS endpoint)." then the caller most likely to get bitten reads it first.

includes/oauth/class-resolver.php
```diff
@@ -52,6 +54,25 @@ public static function handle_to_did( string $handle ): string|\WP_Error {
...
+		 * likely to sit behind a CDN answering 503 or a WAF answering 429,
+		 * and it is reached before plc.directory — without the gate a
+		 * custom-domain user gets "could not resolve handle" for an outage.
+		 *
+		 * 404 is excluded on purpose: a host with no well-known file has
+		 * simply not set up handle verification, which is a different
+		 * problem from an upstream being down, and the existing
+		 * `atmosphere_resolve_handle` message says so.
+		 */
+		if ( 404 !== (int) \wp_remote_retrieve_response_code( $response ) ) {
```
kraftbj, line 68: Heads-up on a second consumer of `handle_to_did()` that I don't think has come up: `Transformer\Facet::resolve_mention()` at `class-facet.php` line 645. It runs at publish time for every mention in a post, and it hands the resolver error to the `atmosphere_mention_didweb_fallback` filter, whose docblock says the error is passed so that "its code distinguishes a definitive miss from a transient network failure."

That claim was only half true until this commit. A connection-level failure gave `http_request_failed`, so that case worked, but an HTTP 503 from the handle's host fell through to `atmosphere_resolve_handle` and was indistinguishable from "this handle genuinely has no well-known file." Which is the case the filter's whole reason for existing (a `did:web` site keeping mentions through a transient blip) actually cares about.

So this gate is what makes the documented contract true, which is a nice second-order win. Two consequences though:

Default behavior doesn't change, since the filter defaults to false and both codes are still `WP_Error`, so nothing breaks for anyone who hasn't implemented it. But it *is* a behavior change for a filter published since 2.0.0: an implementer matching on `atmosphere_resolve_handle` to detect the definitive miss now gets a correctly narrower match, and one matching it to catch any failure will start missing the new codes. Worth a changelog line so it isn't a silent change under their feet.

And the filter's own docblock could now name the codes rather than gesturing at "its code," since there's finally a specific answer: `atmosphere_upstream_error` and `atmosphere_upstream_rate_limited` for transient, `atmosphere_resolve_handle` for a definitive miss.

includes/oauth/class-resolver.php
```diff
@@ -110,6 +136,76 @@ public static function resolve_did( string $did ): array|\WP_Error {
...
+			 * The header comes from a host the submitted handle chose, and
+			 * this message is rendered by `settings_errors()`, which prints
+			 * it without escaping. Coercing to an int is the whole defence:
+			 * a non-numeric value (an HTTP-date, markup, or the array PHP
+			 * hands back for a repeated header) becomes 0 and falls through
+			 * to the variant with no value in it. Also fixes `Retry-After: 0`,
+			 * whose string form is falsy and used to be dropped silently.
+			 */
+			$retry_after = \wp_remote_retrieve_header( $response, 'retry-after' );
+			$retry_after = \is_numeric( $retry_after ) ? (int) $retry_after : 0;
```
kraftbj, line 187: Cosmetic only, no injection, so purely take-it-or-leave-it.

`is_numeric()` accepts more than delta-seconds. A negative passes through as truthy, so a hostile host can render "Retry after -30 seconds", and scientific notation gets expanded, so `1e5` becomes "Retry after 100000 seconds" (about 27 hours). Both are just a silly-looking notice rather than a hazard, since the value is an int by the time it hits `%2$d`.

If it's worth tightening at all, `$retry_after > 0` instead of the truthiness check gets the negatives, and a cap sized to whatever a plausible wait looks like handles the rest. Equally reasonable to leave it: the value is display-only and nothing acts on it.

.github/changelog/fix-resolver-upstream-status
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Show a clearer error when Bluesky's directory or a login server is temporarily unavailable, instead of reporting the account as invalid.
```
kraftbj: This is now understating what shipped. It still scopes the fix to "Bluesky's directory or a login server", but the gate covers the handle verification hop too, and that one points at the user's own domain, which for a custom-domain user is the first thing they hit and the one most likely to be sitting behind a CDN or WAF. That's the population the clearer message helps most, and the entry doesn't mention them.

Something like:

> Show a clearer error when your own domain, Bluesky's directory, or a login server is temporarily unavailable, instead of reporting the account as invalid.

Separately, the `handle_to_did()` change alters which error code the `atmosphere_mention_didweb_fallback` filter receives for an HTTP-level failure (see the other comment). That's a published filter, so a second entry aimed at extenders would be fair. Changed rather than fixed, roughly: "Mention resolution now tells a temporarily unreachable handle host apart from a handle that has no verification set up."

Your call whether that second one clears the bar for a user-facing changelog. It doesn't change default behavior.

kraftbj (commented):
Went back over this after `415d625`. Everything from the last pass is addressed, and I verified the fixes rather than taking the replies at their word:

- The `Retry-After` coercion genuinely closes the sink. I ran the table: markup, an HTTP-date, and the array PHP hands back for a repeated header all land on `0` and fall through to the no-value variant. There's no path left where remote text reaches that string.
- The `Client` migration is equivalent. My one worry was that `Resolver::resolve()` returns no `handle` key, so `set_identity( $resolved )` would blank it, but at that point `$resolved` is the `atmosphere_oauth_resolved` transient, which does carry `handle`. Clean.
- `composer lint` passes locally, and CI is green on 8.2/8.3/8.5 across 6.5, latest, and trunk.

Doing the well-known hop here rather than as a follow-up was the right call, and `upstream_status` is a better answer than the one I suggested.

Three small things left, none of them blocking. Two are documentation, one is a heads-up about a consumer I don't think either of us had looked at.

The consumer is the interesting one: `Transformer\Facet::resolve_mention()` also calls `handle_to_did()`, at publish time, for every mention in a post. The filter it exposes documents the error code as the transient-versus-definitive discriminator, and this PR is what finally makes that true. Worth a changelog line, since it's a published extension point. Details inline.

Also still open from last time, and still fine to leave: the 429 is detected and then discarded. There's no backoff, and `Client::rate_limit_check()` runs only after a *successful* resolve, so a failed one isn't charged to the local bucket either. Repeated Connect clicks against a rate-limited upstream cost up to four outbound requests each. Not this PR's job, but the information is sitting right there now.

kraftbj (dismissed):
Approving. `415d625` addresses everything from the first pass, and I checked the fixes rather than just the replies: the `Retry-After` coercion leaves no path for remote text to reach the message, `set_identity( $resolved )` is equivalent to the old explicit write because the transient carries `handle`, and CI is green on 8.2/8.3/8.5 across 6.5, latest, and trunk.

The four things on my follow-up review are all non-blocking and shouldn't hold this up. The docblock placement and the changelog wording are documentation, the `is_numeric` edge is cosmetic with no hazard behind it, and the `Facet::resolve_mention()` note is a heads-up about a second consumer rather than a defect. Happy for any of them to land here, in a follow-up, or not at all.

The one I'd most like you to see before it ships is the mention-resolution filter, since it's a published extension point whose documented behavior this quietly makes true for the first time. But that's an argument for a changelog line, not for another round.

includes/functions.php
pfefferle: Moved it up. It sits right after the summary line now, before `@since`, so the replace warning is the first thing a caller reads instead of trailing off the end of the tag block.

includes/oauth/class-resolver.php
pfefferle: I had not looked at the Facet path at all, so that second consumer is new to me. The filter docblock now names the codes instead of gesturing at "its code": `atmosphere_resolve_handle` for the definitive miss, `http_request_failed` for a connection-level error, `atmosphere_upstream_error` and `atmosphere_upstream_rate_limited` for a non-2xx from the handle's host. Changelog entry added too, since an implementer matching on `atmosphere_resolve_handle` to catch any failure will start missing the HTTP cases.

includes/oauth/class-resolver.php
pfefferle: Tightened it. `$retry_after < 1 || $retry_after > DAY_IN_SECONDS` sends both to the no-value message, so `-30 seconds` and the 27 hour wait are gone, and `Retry-After: 0` now falls out of the range check rather than out of the truthiness check. Two tests for it, both fail if I drop the range check.

.github/changelog/fix-resolver-upstream-status
pfefferle: Took your wording. Added the second entry as `change-mention-resolution-error-codes` (patch / changed). I went back and forth on whether it clears the bar, but the filter has been published since 2.0.0, so a silent change of which code it hands out seems worth a line even though default behavior is the same.

jeherve (approved):
This should be good to go too. 🚢

PR #218, opened 2026-07-24
Share the rkey DID-provenance logic across transformers

pfefferle:
This came out of a cleanup review of #216, which is merged now, so the base is `trunk`.

## Proposed changes

The DID-provenance block in `get_rkey()` was byte-identical across the Post, Document, and Comment transformers: refresh `META_DID` to the current account, then reserve `META_TID`. #216 added the third copy (Comment), which made the triplication obvious.

This collapses all three onto a single `Base::reserve_rkey_with_provenance()`. Each transformer's `get_rkey()` is now a thin call that passes its own meta accessors (post meta vs comment meta); the DID/TID key set comes from the subclass `static::META_DID` / `static::META_TID` constants.

The point is that this is a security guard with two load-bearing invariants that were being maintained by hand in three places:

* the DID is written **before** the TID, so a partial failure lands in the safe "DID set, no TID" state rather than "TID set, no DID" (which reads as "origin unknown" and re-opens the wrong-repo delete the guard exists to stop);
* the DID is compared before writing, so republishing an unchanged post or comment does not touch the meta at all. Every caller is in the Publisher at publish time; the `wp_head` emitters read the stored AT-URI on purpose and never call `get_rkey()`.

Both now live once, in the helper's docblock and body.

No behavior change. This came out of a cleanup review of #216.

### Other information:

The existing transformer tests cover `get_rkey()` on all three (TID reserved and persisted, DID provenance written and refreshed). They pass unchanged, which is the point: the extraction is behavior-preserving.

## Testing instructions:

```
npm run env-test -- --filter='get_rkey|persists_origin_did|Test_Post|Test_Document|Test_Comment'
```

All green. No user-facing change, so this carries the Skip Changelog label.


includes/transformer/class-comment.php
```diff
@@ -169,32 +168,10 @@ public function get_collection(): string {
...
-		 * account and the mismatch guard in delete_comment() would later
-		 * block a legitimate cleanup. Written BEFORE META_TID so a
-		 * partial failure leaves the safe "DID set, no TID" state rather
-		 * than "TID set, no DID" (which lets the guard fall through to
-		 * get_did() and re-open the wrong-repo delete). Compared before
-		 * writing so a republish of an unchanged comment is a no-op.
-		 */
-		$current_did = get_did();
-		$stored_did  = (string) \get_comment_meta( $comment_id, self::META_DID, true );
-		if ( '' !== $current_did && $stored_did !== $current_did ) {
```
jeherve, line 186: It looks like you got rid of `'' !== $current_did` as part of moving this to `reserve_rkey_with_provenance`. Is there a specific reason for it? Could that create issues in some issues, like when a site is not connected?

jeherve (commented):
This looks good, I only have a minor comment.

includes/transformer/class-comment.php
pfefferle: No good reason, the Post and Document versions never had it and the helper followed those. But it matters: the delete guards read an empty origin as "unknown" and let the delete through, so blanking a real origin on a disconnected site would switch the guard off for that record. Today every caller sits behind `is_connected()`, so I could not find a path that hits it, but the helper is the one place the rule has to hold. Put the guard back there for all three and added tests for the disconnected case.

includes/transformer/class-base.php
```diff
@@ -174,6 +174,62 @@ public function get_uri(): string {
...
+	 * the meta accessors (post vs comment meta); the key set comes from its
+	 * `static::META_DID` / `static::META_TID` constants.
+	 *
+	 * Three invariants live here once, all load-bearing for the guard:
+	 *
+	 * 1. The DID is written BEFORE the TID, so a partial failure between
+	 *    the two writes leaves the safe "DID set, no TID" state. The
+	 *    inverse, "TID set, no DID", reads as "origin unknown" and lets the
+	 *    guard fall through to the current DID, re-opening the wrong-repo
+	 *    delete.
```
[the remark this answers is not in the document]
pfefferle: Right, the disconnected case reserves a TID and skips the DID on purpose. Reworded the invariant to say it applies when there is a DID to write, and noted why that state is the honest one there.

jeherve (approved):

PR #227, opened 2026-07-31
Improve handling of restricted content when publishing

jeherve:
## Proposed changes:

* Adds one visitor-independent seam that decides which part of a post's content is safe to publish to the AT Protocol network, and routes every record field through it. When a membership plugin keeps part of a post restricted, that content no longer shows up in what gets published.
* Jetpack support ships as its own self-contained integration (`integrations/class-jetpack.php`). Other membership plugins can hook the same `atmosphere_publishable_content` filter.

Previously, records were built straight from the stored post content, without checking whether any of it was restricted. Background and specifics are in CMA-43.

### Other information:

## Testing instructions:

* On a site with Jetpack active, create a few posts that restrict their content in different ways: a whole-post subscriber or paid audience, a paywall block partway through, and a paid-content block.
* Connect the plugin and publish or share each one.
* Confirm the restricted portions are absent from the resulting records. A fully restricted post shares only a title and link; a partly restricted post shares only its public portion.

### Changelog entry

Included in `.github/changelog/`.


integrations/class-jetpack.php
```diff
@@ -0,0 +1,245 @@
...
+	 *
+	 * @param string   $content The stored post content.
+	 * @param \WP_Post $post    The post being published.
+	 * @return string Publicly readable content ('' when fully gated).
+	 */
+	public static function filter_publishable_content( string $content, \WP_Post $post ): string {
+		// Name-based detection is robust to block attributes and whitespace that
+		// a serialized-string match would miss.
+		$has_split = \has_block( self::PAYWALL_BLOCK, $content );
+
```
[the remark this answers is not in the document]
jeherve: Addressed in 9fcaefa3c76a1b3c6792220f582268f3500ff482. `filter_publishable_content()` now calls `parse_blocks()` once and drives detection, splitting, and subscriber-view stripping off that single tree. Detection stays depth-aware (a recursive `blocks_contain()` walk) so a nested paywall marker still fails closed, and ungated content is returned verbatim without a serialize round-trip.

tests/phpunit/tests/class-test-functions.php
```diff
@@ -372,6 +373,36 @@ public function test_per_post_disable_gates_publishing() {
...
+
+		\add_filter(
+			'atmosphere_publishable_content',
+			static fn() => 'Public.',
+			10
+		);
+
+		$this->assertSame( 'Public.', get_publishable_content( $post ) );
+
+		\remove_all_filters( 'atmosphere_publishable_content' );
```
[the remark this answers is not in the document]
jeherve: Addressed in 9fcaefa3c76a1b3c6792220f582268f3500ff482. The assertion is now wrapped in `try/finally` so `remove_all_filters( 'atmosphere_publishable_content' )` always runs, even if the assertion fails.

pfefferle:
I read through this a few times. The choke point itself looks right to me, my findings are all about what happens after it: the callers were not really updated for the new "body can be empty or narrowed" state. All of this is from reading the code, I did not test it on a real Jetpack site, so please double check the ones that surprise you.

### I would fix these before merging

**1. A gated short-form post with a featured image publishes an empty record** (`includes/transformer/class-post.php:1253`)

`collect_image_attachment_ids()` now returns `[]` in two different cases: the body really has no image blocks, and the body was gated away. `build_images_embed()` and the `'' === $text && null === $embed` escape hatch in `transform()` both still read it the old way. So for a subscribers-only post that is short form (no title, or any post format like Aside/Status) and has a featured image: text is `''`, the featured-image fallback returns a non-null embed, the link-card fallback never runs, and we publish a record with no text, a bare image and no link back to the post.

**2. A nested paywall block empties a post that is not gated at all** (`integrations/class-jetpack.php:113`)

`blocks_contain()` recurses, but `blocks_above_paywall()` only walks the top level. If the `jetpack/paywall` block sits inside a group, `$above` is null and the filter returns `''`, no matter what `_jetpack_newsletter_access` says. The block's `"parent": ["core/post-content"]` is only an editor restriction, so imports, migrated content, WP-CLI and the REST API can all produce that. A fully public post then publishes an empty body and silently drops to a title and link teaser.

**3. `the_content` puts the paywall back in** (`includes/content-parser/class-parser-base.php:144`, same at `includes/transformer/class-base.php:302`)

We hand the narrowed string to `the_content` but leave the original post as the global post, and Jetpack's own `add_paywall()` runs there at priority 8. Publishing happens in cron, so there is no logged-in user and `user_can_view_post()` is false. Before this PR the marker was still in the string, so Jetpack returned intro + subscribe form. Now the marker is gone, its `strpos()` misses and it falls through to `return $paywalled_content`, so the rendered HTML is only the "Subscribe to keep reading" form. That is what ends up in the document `content`. The tests do not catch it because Jetpack is not loaded in the suite, only the meta key is faked.

**4. The split branch never looks at the access level** (`integrations/class-jetpack.php:111`)

The no-split branch checks `is_access_level_public()`, the `$has_split` branch does not. If someone inserts a Paywall block, then sets visibility back to "Everybody" without deleting the block, Jetpack renders the whole post publicly on the site but we truncate at the marker forever. I think gating that branch on `! is_access_level_public( $post )` keeps it fail-closed and stops punishing public posts.

**5. The pre-publish preview reads the saved access level** (`integrations/class-jetpack.php:185`)

`Pre_Publish_Controller::get_preview()` clones the post and overwrites content/title/excerpt with the unsaved editor state. The block checks honour that, but the whole-post gate reads `get_post_access_level( $post->ID )`, which comes from the last save, and `_jetpack_newsletter_access` is only written on save. So while drafting a new paid post the panel shows the full body and a long-form strategy, and the published record is a teaser. Jetpack memoizes that lookup per request too.

**6. `get_excerpt()` skips the gate** (`includes/transformer/class-base.php:230`)

The `! empty( $post->post_excerpt )` branch returns before it reaches `get_publishable_content()`, which does not match the docblock claim that every body-derived field goes through the helper. A manually written excerpt is fine, but SEO plugins and imports store a generated excerpt (the first 55 words of the body) in `post_excerpt`, and that lands in the document description and the link card. The new test only covers the hand-written case.

### Smaller things

**7. The registration guard** (`integrations/class-load.php:41`)

The comment above it says `Jetpack_Memberships` is often not loaded yet at `plugins_loaded`, which means the second arm of the `||` is almost never true and the guard is really just `defined( 'JETPACK__VERSION' )`. On installs where the gating blocks exist without that constant (Simple, jetpack-mu-wpcom style setups), the filter is never registered and gated posts federate in full. Keying off the meta or the blocks, or checking at publish time instead, would be safer.

**8. The helper is not cached** (`includes/functions.php:822`)

It is a plain `apply_filters()` and there are about eight call sites, two of them per transform. With the Jetpack callback attached, each call is a full `parse_blocks()` plus a `serialize_blocks()`, and `collect_image_attachment_ids()` parses the result again. The pre-publish preview endpoint runs on every keystroke, so on a long post this gets expensive. `Parser_Base` already caches the block tree per post, something similar here would do.

**9. The parser contract still says raw content** (`includes/content-parser/interface-content-parser.php`, `integrations/README.md`)

The interface still tells parser authors they get "raw post content (post_content)" and may re-run `the_content` themselves. A third-party parser doing exactly what the docs say puts the gated body straight back into the document. Worth updating both, since the docs are the whole point of that extension point.

**10. Changelog wording** (`.github/changelog/fix-paywalled-content-federation`)

Two em dashes and "AT Protocol network", and this line goes to the update screen. Something like "Content restricted to subscribers is no longer shared to Bluesky. This covers whole posts, paywalled sections, and paid-content blocks." would fit our rules better.

jeherve:
Thanks, this was a genuinely useful pass — all ten landed in 82028ce. I checked the two claims that hinged on Jetpack's runtime (#3, #4) against `add_paywall` and `get_post_access_level`/`user_can_view_post` in Jetpack itself before changing anything, and you were right on both.

**Fixed as described:**

- **#1** — a whole-post-gated short-form post with a featured image now falls back to the link card (title + link home) instead of shipping a bare image.
- **#2 / #4** — the whole-post *and* split gates now key off the stored access level, not block presence. A public post with a stray/nested paywall block is left whole; a nested block on a gated post still fails closed. (Jetpack defaults empty meta to `everybody` and `user_can_view_post()` returns true there, so the block alone was never the signal.)
- **#3** — renders now run through a `render_publishable_content()` seam that unhooks `add_paywall` for the duration, so it can't overwrite the narrowed body with the subscribe form. There was a third render site (`build_short_form_text`) with the same exposure — also covered.
- **#5** — the panel forwards the unsaved access level and a generic `atmosphere_pre/post_projection` seam applies it during projection, so the preview matches what will publish. Jetpack specifics stay in the integration.
- **#7** — registration now also keys off `IS_WPCOM` for Simple / mu-wpcom.
- **#8** — `get_publishable_content()` memoizes per post + content hash (the hash keeps the preview clone from colliding with the saved post).
- **#9 / #10** — parser contract docs and changelog updated.

**Two calls I want to flag:**

- **#4 changed a test.** `test_split_detects_paywall_block_with_attributes` asserted a split with no access meta, which contradicts Jetpack (an inert block on a public post). I switched it to a gated level and added public-post + nested-block regressions.
- **#6 I took as a docs fix, not a behavior change.** A hand-written `post_excerpt` is the public teaser Jetpack itself surfaces for a gated post, and a test already asserts it's preserved — so I corrected the overclaiming docblocks rather than gating it. If you'd rather fail closed on stored excerpts too, happy to add it.

Regression coverage added for #1, #3, #4, #5, #7. `composer lint` clean, full suite green (1142 tests).

includes/transformer/class-post.php
```diff
@@ -1140,6 +1148,24 @@ private function carry_mentions_into_teaser( array $texts ): array {
...
+	 * empty post from one that only looks empty because it is fully gated, so
+	 * the short-form path can fall back to a link card that still links home
+	 * instead of shipping a bare, contextless featured image. A split (partly
+	 * gated) post keeps a public portion, so it is not "body gated" here.
+	 *
+	 * @return bool
+	 */
+	private function is_body_gated(): bool {
+		return '' !== (string) $this->object->post_content
+			&& '' === get_publishable_content( $this->object );
```
pfefferle: This compares against `''`, but `Document::get_content()` a few files over trims first:

```php
if ( empty( \trim( $publishable ) ) ) {
```

That difference matters here. `blocks_above_paywall()` keeps every block above the split, and `parse_blocks()` emits a `blockName: null` block for leading whitespace, so a fully gated post does not always serialise to exactly `''`. I ran the real filter in wp-env:

```
paywall first    => ''    strict_empty=YES  trim_empty=YES
leading newline  => '\n'  strict_empty=NO   trim_empty=YES
leading space    => ' '   strict_empty=NO   trim_empty=YES
```

So with any whitespace before the paywall block `is_body_gated()` is false, and in `compose()`:

```php
if ( '' === $text && ( null === $embed || $this->is_body_gated() ) ) {
```

`$text` renders empty, `$embed` is not null when the post has a featured image, so the link-card fallback does not fire. The post goes out as a bare image, which is the thing the comment right above says this check exists to avoid.

Narrow, since it needs whitespace-only content above the paywall (imports, hand-edited posts, an intro paragraph deleted but its newline left). Nothing leaks, the share is just worse. `'' === \trim( get_publishable_content( $this->object ) )` fixes it and makes the two transformers agree.

pfefferle (commented):
Read the whole seam rather than just the diff, and traced every remaining `post_content` read in `includes/transformer/` and `includes/content-parser/`. What is left is comments, the deliberate `is_body_gated()` check, and paths that now go through `render_publishable_content()`. I could not find a way to get gated content into a record.

The shape is right: one visitor-independent decision, every body-derived field routed through it, fail-closed by contract, and the integration reading stored meta instead of viewer state. Suspending Jetpack's own `the_content` paywall around our render, with the depth counter and the `finally`, is the part I would most likely have gotten wrong myself.

Two things I would want before merge, both inline. Neither leaks anything.

1. `Post::is_body_gated()` compares against `''` while `Document::get_content()` trims, so the two halves of this PR disagree about what empty means. A fully gated split post with whitespace before the paywall block ships as a bare featured image instead of the link card.
2. The memo cache in `get_publishable_content()` is never evicted, which walks back the per-batch eviction `Backfill_Command` does on purpose.

There is a third, latent one in the same inline comment as 2.

### What I checked and found clean

`build/` is byte-identical to a fresh `npm run build`, including both asset hashes. PHPCS 40/40, `lint:js` clean, 23 JS tests, 1145 PHP tests / 3055 assertions. `accessLevel` is `sanitize_key`'d and any unrecognised value falls through to gated. `Jetpack_Memberships::get_post_access_level()` is guarded by `class_exists` and `method_exists`. `blocks_above_paywall()` returns null for a nested marker and the caller fails closed. The `serialize_blocks()` round-trip is correctly skipped for an untouched public post. The loader's three signals cover Simple, and both files ship in one versioned package behind the path autoloader, so there is no split-deploy window.

Two behaviour changes that I think are right but are worth being deliberate about, both already in the description: an author-written `post_excerpt` still federates on a fully gated post, and `Content_Parser::parse()` now gets narrowed content, which is a semantic change for any third-party parser even though the signature is the same.

Risk feels medium to me, only because the seam changes what every body-derived field publishes on every site, not because anything looks unsafe.

includes/functions.php
```diff
@@ -771,6 +771,130 @@ function is_sharing_enabled( \WP_Post $post ): bool {
...
+	/*
+	 * Per-request memoization. Integrations doing real work here (Jetpack
+	 * parses and re-serializes the block tree) are hit several times per
+	 * transform and again on every keystroke of the pre-publish preview, so
+	 * the repeated parse dominates. Key on the post ID plus a hash of the
+	 * stored content: the pre-publish endpoint projects a clone that keeps the
+	 * real ID but carries unsaved content, and the hash keeps those from
+	 * colliding with — or masking — the saved post.
+	 */
+	static $cache = array();
```
pfefferle: This cache is never evicted, and for a public post the value is the whole `post_content` verbatim.

`Backfill_Command` walks every post ID in one `foreach` in a single process, and it already goes out of its way to stop exactly this:

> Every $batch posts, drop the per-post object-cache entries accumulated since the last flush. Long runs (10k+ posts) would otherwise hold every visited WP_Post + meta + term row in the in-process cache for the entire process lifetime.

The new static brings that growth back, and `clean_post_cache()` cannot reach it. The memoisation itself is clearly worth keeping, the preview re-parses on every keystroke, but I think it wants a bound: last N entries, or just a single-entry cache for the current post, which is all the transform path needs.

Separate and smaller, on the key one line down. The comment says the hash "keeps those from colliding with, or masking, the saved post", but `Jetpack::$access_override` also decides the output and is not in the key. An author who changes only the access-level dropdown and previews has unsaved content that hashes the same as the saved content, so the projected answer takes the saved post's slot.

Not reachable today as far as I can tell: the preview endpoint projects once per request and nothing transforms that post again after. Worth noting because the guarantee is narrower than it reads, and if a second consumer ever shows up in the same request it fails open, not closed.

includes/transformer/class-post.php
jeherve: Fixed in c4aed9e. `is_body_gated()` now trims before comparing, so it agrees with `Document::get_content()`: a whitespace-only body above the paywall reads as gated, and a titleless gated post with a featured image falls back to the link card instead of shipping the bare image.

Writing the regression test sent me down a small rabbit hole worth flagging. The obvious version of it passed even with the bug still in place, because `set_post_thumbnail()` silently deletes the meta when the factory attachment has no real image file. So the featured image was never actually set, and the fallback fired through the `null === $embed` branch rather than `is_body_gated()`. The existing `test_gated_titleless_post_with_featured_image_uses_link_card` has the same blind spot. The new test sets `_thumbnail_id` directly so the image is really present, and it does fail without the trim (the post goes out as `short-form` with a bare image). :)

includes/functions.php
jeherve: Both fixed in c4aed9e.

The cache is now bounded: it drops its oldest entry once it hits a cap, so a long `Backfill_Command` run no longer holds every visited post's content for the life of the process. The transform path only ever reads back the post it's working on, so a small bound is enough.

On the key, you're right that it only hashes `post_content` while the access override also decides the output without touching the content, so the guarantee was narrower than the comment read. Rather than teach the generic helper about Jetpack, I added an `atmosphere_publishable_content_cache_key` filter and had the integration fold its override into the key. An overridden preview no longer lands in the saved post's slot, even if a second consumer shows up in the same request. The new regression test covers that exact collision: without the fix it returns the saved public body.

pfefferle:
I had a look at this, mostly because I am touching the same publish path in #248. Four things, one of them I think is a real problem.

**The access level override always fires** (`integrations/class-jetpack.php:126`)

`accessLevel` is registered with `'default' => ''`, and `WP_REST_Request::has_param()` counts defaults: `get_parameter_order()` includes `'defaults'`, and `dispatch_request()` calls `set_default_params()` before the callback runs. So `has_param( 'accessLevel' )` is true on every dispatched request, also when the client sent nothing. `set_access_override()` then stores `''`, `is_access_level_public()` looks at the override first, and `'' === $level` is true, so the preview says public no matter what is saved.

What that looks like: a post with `_jetpack_newsletter_access` set to `paid_subscribers`, and a request that does not send `accessLevel` (a cached older `build/pre-publish-panel/plugin.js`, or a post type where the meta is not in the editor's `meta` object). The panel shows the full body as what will be published, the actual publish sends only a link card. That is the mismatch the `atmosphere_pre_projection` docblock says this is there to avoid.

Checking `'' !== (string) $request['accessLevel']` instead of `has_param()` should do it. Worth a look at the `customText` guard in `includes/rest/admin/class-pre-publish-controller.php:270` too, it has the same shape, and its comment says it falls back to the saved meta for an older editor, which `has_param()` does not actually give you.

Three smaller ones:

* `includes/functions.php:829`, the memo key is the post ID plus `md5( post_content )`, but the answer also depends on `_jetpack_newsletter_access`, which is not in the key. Change the access level in the same process and you get the earlier, more permissive answer back. `Parser_Base` has `flush_block_cache()` for this, there is no equivalent here, so tests cannot reset it either.
* `includes/transformer/class-post.php:1165`, `is_body_gated()` compares raw `post_content` against `''` without trimming. A post whose content is only a newline, with a featured image and no gating at all, comes out as gated and moves from the short form image embed to a link card. Trimming the first clause too should keep the gated case and restore the old behavior.
* `integrations/class-jetpack.php:254`, the paywall suspension is a global `remove_filter`, so a nested `the_content` for a different post (a query loop in the body, or a shortcode that renders another post) loses its gate too. Probably rare, but worth a note at least.

Happy to push fixes for these directly if that is easier for you, just say the word.

jeherve:
All four addressed in c199e7e.

**The access-level override always firing.** You're right, and it's the one that could actually mislead. `has_param()` counts the `''` default once the request is dispatched, so the guard fired even when the client sent nothing, and a blank override read as "everybody". Both guards now key off a non-empty value (`'' !== (string) $request['accessLevel']`, same for `customText`) and fall back to the saved meta otherwise, which fails closed. The existing tests never caught it because they call the controller directly and skip the dispatch that fills defaults, so I added two that set the param to its dispatched `''` and assert the saved gate still wins.

One nuance on `customText`: the modern editor always sends the current field value, so a blank now falls back to the saved custom text instead of forcing the default composition. That matches the guard's original comment; the only case it changes is an author clearing the field before saving, where the preview shows the saved text until the save lands (fails safe). Shout if you'd rather it track the cleared field live.

**`is_body_gated()` trimming.** Fixed, both sides trim now. A whitespace-only ungated post with a featured image stays on the short-form image path instead of the link card. The regression test ships `short-form`, not `link-card`, without the trim.

**The memo key.** The Jetpack cache-key filter now folds in the effective access level (the override during a preview, the saved meta otherwise), so gating a post mid-process no longer hands back the earlier, more permissive answer. I also added `flush_publishable_content_cache()` for the reset you noted was missing, with a test through it.

**The global paywall suspension.** Noted in the docblock rather than fixed. Scoping it to our post needs a reliable global `$post` inside the nested `the_content`, which the logged-out cron render doesn't give us, and I didn't want to trade the primary suspension's correctness for the narrow nested-render case. Happy to revisit if you have a clean way to scope it.

I also brought the branch up to date with trunk and resolved the conflicts (build artifacts only) along the way.

pfefferle (approved):

PR #229, opened 2026-07-31
Restrict who can reply to Bluesky posts (threadgate)

### LINKED ISSUE #228: Restrict who can reply to Bluesky posts (threadgate)
Right now every post we crosspost to Bluesky is open to replies from anyone. Bluesky can restrict that with the `app.bsky.feed.threadgate` record: replies allowed from nobody, mentioned users, people you follow, your followers, or members of a list.

We don't write that record today, so there is no way to set reply rules from WordPress.

Idea:

- a `Threadgate` transformer that shares the post's rkey (collection `app.bsky.feed.threadgate`, its `post` field is the post AT-URI)
- emit it in the same `applyWrites` batch as the post, the way we already pair the doc record
- a per-post "who can reply" control in the share panel, default "everybody" (no record written)

Only the root post author can set a threadgate, and the record must share the post's rkey, so it slots into the existing transformer + batch-publish pattern.

Taking this one.

pfefferle:
Fixes #228

Screenshot 2026-07-31 at 18 33 33

## Proposed changes:

Bluesky lets you limit who can reply to a post with an `app.bsky.feed.threadgate` record. We didn't write that record, so every crossposted post was open to replies from anyone. This adds a per-post "who can reply" control.

* New `Threadgate` transformer. It shares the post's rkey (collection `app.bsky.feed.threadgate`) and builds the `allow` list from the chosen audiences.
* Publisher writes the threadgate in the same atomic `applyWrites` batch as the post, for both short-form and thread publishes. No record is written when the post is open to everyone.
* A per-post setting in the editor sidebar: Everybody, Nobody, or Specific people (mentioned / following / followers, combinable).
* New `atmosphere_reply_restriction` post meta, cleaned up on uninstall.

Left as follow-ups on purpose (the deferred "delete cleanup" work): changing the restriction after a post is published does not re-sync yet, deleting a post does not delete its threadgate record (they share an rkey but are separate records), and a single ↔ thread strategy change (`rewrite_thread`) doesn't delete the old threadgate. The failed-publish rollback path *does* clean up the threadgate, so a gated post never leaves a dangling record on a failed publish.

### Other information:

## Testing instructions:

You need a connected Bluesky account (or a simulated connection).

* Edit a post. In the ATmosphere panel in the document sidebar, find "Who can reply on Bluesky".
* Pick "Specific people" and tick one or more of the audience boxes, then save. The setting should persist (the `atmosphere_reply_restriction` post meta holds your choice).
* Publish the post. The `applyWrites` batch should include an `app.bsky.feed.threadgate` create that shares the post's rkey, with an `allow` list matching your choice.
* Set another post to "Everybody" and publish: no threadgate record is written.
* Set a post to "Nobody": the threadgate is written with an empty `allow` list.

### Changelog entry

A changelog entry is already committed at `.github/changelog/add-bluesky-reply-restrictions` (minor / added), so the checkbox workflow below isn't needed.


jeherve (commented):
This doesn't work for me. I think for all of this to work, we also need to update the `SCOPES` list in `class-client.php`. I think we'd need to add `repo:app.bsky.feed.threadgate`.

That said, that also means that sites that are already connected will need to reconnect.

pfefferle:
You are right, the scope was missing. Added in df82935.

The reconnect part I tried to handle in 59875c8, so existing sites do not just silently lose the feature:

* The granted `scope` from the token response is now stored on connect, and backfilled on the hourly refresh (same trick as the key fingerprint). So a site that connected before this knows within an hour whether it has the scope, without anyone reconnecting just to find out.
* When the scope is known to be missing, the publisher leaves the threadgate out of the batch. The gate shares the atomic `applyWrites` with the post, so a rejected gate would have failed the post too. The post still goes out, the restriction stays in meta, and the first update after a reconnect writes it.
* A warning next to the reply control in the editor, a notice on the settings page, and a Site Health entry all point at reconnecting. Recommended, not critical, since posts still publish.

An unknown scope (not refreshed yet, or a PDS that does not send it) is allowed through, so nothing is hidden on day one.

One thing I could not check: whether the PDS actually returns `scope` in the token response. If it does not, we never learn and the write is just attempted. Since you have a setup where it failed, could you have a look at what the token response contains when you reconnect? And what error the PDS returned for the rejected write, so I can classify it as a reconnect error.

jeherve (approved):
This is looking good now, should be good to merge!

PR #233, opened 2026-08-05
Warn in the editor when the Bluesky connection has expired

### LINKED ISSUE #245: The editor panel should tell you when your site isn't connected to Bluesky
Right now the ATmosphere panel in the block editor looks the same whether or not your site is connected to Bluesky. If your connection has lapsed (a token refresh failed and the site is sitting in a needs-reauth state, say), the "Share this post" toggle is still on and the help text still reads "This post will be shared via ATmosphere when published." So you write the post, publish it, and nothing goes out to Bluesky. You only find out later, which isn't a fun surprise :)

I'm wondering if the panel should say so up front: when the site isn't connected, show a small notice under the toggle explaining that the post won't be shared, with a link to reconnect on the settings page.

I considered just hiding the panel until you're connected, but the toggle stores a per-post preference that still matters once you reconnect, so hiding it would stop you from setting share intent on drafts while a connection is being sorted out. A notice keeps the panel useful and fills in the missing context.

The pre-publish panel is already connection-aware (it tells you "Your site isn't connected to Bluesky yet."), so this mostly brings the document sidebar panel in line with it.

What do you think?

pfefferle:
Fix #245

Screenshot 2026-08-05 at 16 33 30
Screenshot 2026-08-05 at 16 33 38

Reported here: https://wordpress.org/support/topic/push-single-post/#post-18984243. Two people published a post while the connection had expired, and it silently never reached Bluesky.

This is only the warning half of that thread. It does not let you push the old post afterwards, `wp atmosphere backfill` is still the only way to recover one, so the request for a manual per-post push stands.

## Proposed changes:

The editor now tells an author what will actually happen to their post, in one message, from one decision.

Before this, the only signal was the site-wide admin notice, which needs `manage_options` and is effectively invisible inside the block editor, so an author could publish a post that was silently never shared.

* **One decision, computed once.** `Atmosphere\share_status()` answers "can this site share right now, and if not why" in a single place, with the precedence written down. Both editor surfaces render that answer instead of each deriving it from a handful of flags, which is what let them contradict each other.
* **One message at a time.** The document panel shows at most one thing, by precedence: a dead connection outranks a failed share, which outranks a pending removal. A dead connection explains both of the others, so it speaks for them rather than stacking with them.
* **The panel follows site policy, not connection health.** The share toggle stays put while the connection is down, because it records a preference and `wp atmosphere backfill` reads that meta when the site recovers.
* **Removal is called what it is.** Making a shared post private, password-protecting it, or switching its toggle off deletes the Bluesky record. The reasons now say so, because "this post will not be shared" reads like a no-op when it is not one.
* **The admin notice reaches everyone.** It used to require `manage_options`, so the person hitting Publish never saw it. It now renders for any user, with the action swapped to "ask an administrator" for those who cannot reconnect, and its consequence sentence follows what the site actually uses the connection for.

## Scenarios

One post, already shared to Bluesky, in each state. Each shows at most one message.

**1. Healthy connection, sharing on.** No message. Toggle, custom text, and the link to the shared post.

1-healthy-shared

**2. A published post about to be removed from Bluesky.** Switching the toggle off, going private, or adding a password all delete the record when the change is saved, and that is not undoable: republishing mints a new record with a new URL, so the reactions on the old one are gone. The warning says that rather than "this post will not be shared", which reads like a no-op. Only shown for a post whose saved status is publish, since nothing is removed otherwise.

2-about-to-be-removed

**3. Last share failed, retry queued.** Error notice with the server's own message underneath. Red is right here: it is a genuine per-post failure, not a restatement of something else.

3-share-failed-retrying

**4. Connection needs a reconnect.** Warning with the cause and a reconnect link. The toggle stays and its help stops promising delivery.

4-needs-reconnect

**5. Automatic sharing turned off in settings.** A plain sentence, no toggle, link kept. The site owner turned it off, so they are told where.

5-sharing-off-in-settings

**6. Sharing turned off by another plugin (connection-only mode).** Nothing but the link, and the pre-publish panel does not render at all. A host plugin owns the sharing experience there, and narrating the arrangement to an author who cannot act on it is noise.

6-sharing-off-by-another-plugin

**7. Dead connection and a failed share together.** One warning. The failure is a consequence of the dead connection, so it does not get its own red box. This state used to render two notices and two help texts, 610px of panel; it is 512px and one notice now.

7-worst-case-dead-connection-and-failed-share

### Other information:

PHPUnit for `share_status()`, the editor config, and every branch of the pre-publish decision. Jest for the two pure helpers that own the copy and the precedence. The JSX itself is not covered; this repo's Jest setup tests helpers only, which is why the branchy parts live in `utils.js`.

## Testing instructions:

You need a site connected to Bluesky. Each step below matches a scenario above.

1. **Dead connection:** `wp-env run cli wp option patch update atmosphere_connection needs_reauth 1`. The panel shows one warning with the cause and a reconnect link; the toggle stays and its help says the post will not be shared while disconnected.
2. **Add a failed share on top:** set `_atmosphere_last_publish_error` on the post. Nothing changes on screen. That is the point.
3. **Clear the connection flag,** and the failure now shows on its own as a red error with the server message.
4. **Sharing off in settings:** `wp-env run cli wp option update atmosphere_auto_publish 0`. No toggle, one plain sentence, link kept.
5. **Sharing off by another plugin:** add `add_filter( 'atmosphere_should_auto_publish', '__return_false' );` in an mu-plugin. The panel says nothing, and the Bluesky pre-publish panel is gone entirely.
6. **Removal warning:** on a published post that is on Bluesky, switch its share toggle off, or set it to private. The panel should warn that saving will remove it from Bluesky and that this cannot be undone. On a draft it should stay quiet, because nothing is removed there.
7. **As a non-admin:** the same states, with "Ask an administrator to reconnect it." instead of a link, and a generic cause sentence rather than the specific one.
8. **Undo:** `wp-env run cli wp option patch delete atmosphere_connection needs_reauth` and `wp option delete atmosphere_auto_publish`.

### Changelog entry

Already committed as `.github/changelog/add-editor-reauth-reminder`.


src/editor-plugin/plugin.js
```diff
@@ -131,23 +126,25 @@ const EditorPlugin = () => {
...
+			     explains why the toggle's help text is hedged. The lead comes
+			     from PHP (`reauthLead`), which picks the cause sentence and
+			     already accounts for an operator-initiated disconnect. */ }
+			{ NEEDS_REAUTH && (
+				<BaseControl>
+					<Notice status="warning" isDismissible={ false }>
+						{ REAUTH_LEAD } <ReconnectAction />
+					</Notice>
+				</BaseControl>
+			) }
```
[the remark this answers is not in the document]
pfefferle: Good catch on the description, that line was stale and I have fixed it.

The behaviour itself is intentional though. The connection is not a per-post thing: ATmosphere can be embedded as a pure connection layer, where another plugin relies on the connection even when this site does not cross-post at all. Hiding a dead connection because the author switched sharing off for one post would hide a site level problem.

The split is: the sidebar panel only loads when automatic sharing is on, and then always shows the reconnect warning. The pre-publish panel answers the per-post question, so there the disabled check comes first and there is no reconnect prompt.

includes/rest/admin/class-pre-publish-controller.php
```diff
@@ -263,12 +265,19 @@ public function get_preview( WP_REST_Request $request ) {
...
-				'strategy'      => $projection['strategy'],
-				'limit'         => $projection['limit'],
-				'records'       => $projection['records'],
+				'will_publish'    => $decision['will_publish'],
+
+				/*
+				 * Only the expired-session branch sets this, so every other
+				 * "will not publish" reason defaults to false and stays an
+				 * info-level note in the panel.
+				 */
```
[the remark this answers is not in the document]
pfefferle: Right, the comment was wrong. That branch covers a deliberate disconnect too, not just an expired session. Reworded in 8f2d7fb.

includes/class-block-editor.php
```diff
@@ -45,7 +47,20 @@ public static function enqueue(): void {
...
+		 * The document panel is the cross-posting UI itself, so it stays out
+		 * of the way when the site does not cross-post. The pre-publish
+		 * panel's whole job is to answer whether this post will be shared,
+		 * and "automatic publishing is turned off" is that answer — so it
+		 * keeps loading either way.
+		 */
+		$auto_publish_enabled = is_auto_publish_enabled();
+
 		foreach ( self::SCRIPTS as $name ) {
+			if ( 'editor-plugin' === $name && ! $auto_publish_enabled ) {
```
kraftbj: This drops the entire panel, not just the cross-posting controls. Also going away: the "View on Bluesky" link, the "sharing is off but this post is still on Bluesky" warning, the publish-error notice, and the custom text field. Those describe state created *before* auto-publish was switched off, and the records aren't cleaned up either, since `on_status_change()` bails on the same helper. So the post stays on Bluesky with nothing in the editor acknowledging it.

The per-post toggle is the part that worries me most. `wp atmosphere backfill` gates every post on `is_post_publishable()`, which reads `atmosphere_disabled`. Turning auto-publish off hides the only control over a meta value that still decides whether the documented recovery path will ever share that post.

Could the gate move inside the component instead? Hide the toggle and textarea when auto-publish is off, keep the shared-URL link, the removal warning, and the error notice.

includes/class-block-editor.php
```diff
@@ -45,7 +47,20 @@ public static function enqueue(): void {
...
+				continue;
```
kraftbj: Second concern on the same gate, separate from the panel-contents one above.

The description says the reconnect warning isn't tied to the per-post toggle because a host plugin can rely on the connection even when this site doesn't cross-post. That reasoning holds, but this gate defeats it: connection-only mode forces `is_auto_publish_enabled()` false, so the panel never loads and the banner never renders in precisely that scenario. The settings page is hidden there too, which leaves the admin notice and Site Health as the only signals.

Would gating on `'' !== reauth_lead_for_current_user()` as an additional condition work? Load the panel when there's something to warn about, even with auto-publish off.

includes/class-block-editor.php
```diff
@@ -88,21 +103,34 @@ private static function enqueue_script( string $name ): void {
...
+		$can_manage = \current_user_can( 'manage_options' );
+
+		/*
+		 * `reauth_lead_for_current_user()` already folds in the
+		 * operator-disconnect suppression for non-admins (an empty
+		 * lead there), so `needsReauth` is derived from the lead
+		 * rather than re-deriving the same rule here.
+		 */
+		$reauth_lead  = reauth_lead_for_current_user();
+		$needs_reauth = '' !== $reauth_lead;
```
kraftbj: `'' !== $reauth_lead` isn't the same question as "does the connection need a reconnect." `reauth_lead_for_current_user()` returns `''` in two unrelated cases: no reconnect needed, and a reconnect *is* needed but it was an operator disconnect and the reader lacks `manage_options`. The second is a suppression of the cause copy, not of the state.

That matters because the flag feeds `shareHelpText()` as well as the banner. `Client::disconnect()` keeps `atmosphere_identity` (it only deletes `atmosphere_connection`), so `needs_reauth()` is true, and an author on a deliberately disconnected site reads "This post will be shared via ATmosphere when published." The pre-publish panel tells the same user the opposite in the same session.

I think this needs two signals: one for "is the connection dead" (drives the hedged help text) and one for "is there a cause sentence to show" (drives the banner).

includes/rest/admin/class-pre-publish-controller.php
```diff
@@ -320,6 +332,39 @@ private function publish_decision( WP_Post $post, string $status, string $passwo
...
 		}
 
+		if ( ! is_connected() ) {
+			/*
+			 * `is_connected()` is false for both a dead session and a site
+			 * that never connected. Only the first is fixable by an admin,
+			 * so it gets its own copy and lifts the panel's notice from
+			 * info to warning.
+			 */
+			if ( needs_reauth() ) {
```
kraftbj, line 363: This branches on raw `needs_reauth()` while `Block_Editor::script_data()` derives its flag from the lead, so the non-admin operator-disconnect suppression is enforced on one surface and bypassed on the other. An author gets a warning-level notice plus "Ask an administrator to reconnect it." for a disconnect the sidebar deliberately stayed quiet about.

Deriving `needs_reconnect` from a non-empty `reauth_lead_for_current_user()` would make both surfaces agree. That also resolves the contradiction I flagged on `class-block-editor.php:123`.

includes/rest/admin/class-pre-publish-controller.php
```diff
@@ -320,6 +332,39 @@ private function publish_decision( WP_Post $post, string $status, string $passwo
 			);
 		}
 
+		if ( ! is_connected() ) {
```
kraftbj, line 356: The reconnect branch sits above the password, private, and post-type checks, so an author writing a private post gets a warning with a live "Reconnect your Bluesky account" link. Reconnecting won't help. `is_post_publishable()` blocks those three unconditionally.

The comment above the auto-publish check argues it goes first because "reconnecting would not change whether this post publishes." Same argument applies to the three checks sitting below this one. Moving the `is_connected()` block down past them would make the ordering consistent with its own rule.

src/editor-plugin/utils.js
```diff
@@ -19,3 +20,34 @@ import { DISABLED_META_KEY } from '../config';
...
+	if ( ! enabled ) {
+		return __(
+			'This post will not be shared via ATmosphere.',
+			'atmosphere'
+		);
+	}
+
+	if ( needsReauth ) {
+		return __(
+			'Sharing is on for this post, but it will not be shared while your site is disconnected. An administrator can share it later once the connection is back.',
```
kraftbj: "An administrator can share it later once the connection is back" reads like there's a button somewhere. There isn't. `on_status_change()` bails at `! is_connected()` before `META_TID` is ever reserved, so a later re-save lands in `Publisher::update_post()`'s pristine-post branch and gets skipped on purpose. `wp atmosphere backfill` is the only recovery, which the description says too.

Something closer to "an administrator can re-share it later with WP-CLI" would match what the code does, though I don't love pointing authors at WP-CLI either. Open to better wording.

src/editor-plugin/plugin.js
```diff
@@ -89,24 +93,39 @@ const EditorPlugin = () => {
...
+				{ __(
+					'Sharing to Bluesky failed because your site is no longer connected to Bluesky.',
+					'atmosphere'
+				) }{ ' ' }
+				<a href={ RECONNECT_URL }>
+					{ __( 'Reconnect on the settings page.', 'atmosphere' ) }
+				</a>
+			</>
+		);
+	} else {
```
kraftbj, line 229: This `else` also catches an administrator who has `CAN_MANAGE` but an empty `RECONNECT_URL` (settings page hidden, no Connectors screen), so an admin gets told to ask an administrator. `ReconnectAction` handles that case correctly at `reconnect-notice.js:25`, which makes this the first drift between the shared component and the copy kept inline to preserve translations. Narrow to reach, but it's the failure mode the shared component exists to prevent.

src/shared/reconnect-notice.js
```diff
@@ -0,0 +1,37 @@
...
+
+import { __ } from '@wordpress/i18n';
+import { CAN_MANAGE, RECONNECT_URL } from '../config';
+
+/**
+ * What to do about a dead connection, addressed to the current user.
+ *
+ * @return {React.JSX.Element|string} A link for administrators, plain text for everyone else.
+ */
+export function ReconnectAction() {
```
kraftbj, line 17: Three copy branches rendering on two production surfaces, no test. There's already a `jest.mock`-based precedent for config-driven modules in `src/settings-connect/__tests__/index.test.js`. Could we get coverage for the three arms — admin with a URL, admin without one, and non-admin?

tests/phpunit/tests/wp-admin/class-test-reauth-notice.php
```diff
@@ -226,6 +226,33 @@ public function test_no_output_when_marker_present_but_identity_absent(): void {
...
+				'access_token'  => '',
+				'needs_reauth'  => true,
+				'reauth_reason' => 'key_changed',
+			),
+			false
+		);
+
+		$html = $this->capture_notice();
+
+		$this->assertStringContainsString( \Atmosphere\reauth_lead_for_current_user(), $html );
```
kraftbj: This computes its expectation by calling the same production helper the code under test calls, so it can't catch a bug in that helper. It's also vacuously true if the helper ever returns `''`. Every sibling test in this file asserts a literal string. Worth matching them.

tests/phpunit/tests/class-test-functions.php
```diff
@@ -1088,4 +1091,105 @@ public function test_is_bluesky_post_enabled_filter_can_disable() {
...
+		\delete_option( 'atmosphere_identity' );
+		\delete_option( Client::DISCONNECTED_OPTION );
+	}
+
+	/**
+	 * An operator-initiated disconnect is a state the administrator chose, not
+	 * a problem for every author to worry about: a non-admin gets no lead at
+	 * all for it (not even the generic "needs attention" sentence), matching
+	 * the suppression `Block_Editor::script_data()` relies on for the document
+	 * panel — without this, the pre-publish panel would still nag a
```
kraftbj, line 1330: The claim that "without this, the pre-publish panel would still nag a non-admin about a disconnect the administrator deliberately chose" isn't what happens. `publish_decision()` branches on `needs_reauth()`, not on the lead, so the panel still nags — it just drops the cause sentence. If the change I suggested on `class-pre-publish-controller.php:342` lands, this comment becomes true and can stay as written.

includes/class-block-editor.php
```diff
@@ -88,21 +103,34 @@ private static function enqueue_script( string $name ): void {
...
+
 		return array(
 			'previewPath'       => Pre_Publish_Controller::full_route(),
 			'disabledMetaKey'   => ATMOSPHERE_META_DISABLED,
 			'customTextMetaKey' => ATMOSPHERE_META_CUSTOM_TEXT,
 			'settingsUrl'       => settings_url(),
```
kraftbj: `settingsUrl` has no consumers left after this branch. `plugin.js` moved to `RECONNECT_URL`, and `SETTINGS_URL` in `src/config.js` is now unreferenced. Fine to drop both.

kraftbj (changes requested):
Real problem, and the support thread makes the case on its own. The admin notice needs `manage_options`, so the person actually hitting publish never sees it. Worth fixing.

Three things I'd want resolved before this lands, and one of them undercuts the stated design.

**The enqueue gate takes the whole panel with it.** `class-block-editor.php:60` skips `editor-plugin` when auto-publish is off, but that script is also the only home for the "View on Bluesky" link, the "still on Bluesky, will be removed on next sync" warning, the publish-error notice, the custom text field, and the per-post toggle. Turn off automatic sharing on a site that has been cross-posting, and every already-shared post silently loses its Bluesky link, and any post sitting on a publish error shows nothing. `trunk` enqueues both scripts unconditionally, so this is new here. Worse, `wp atmosphere backfill` still gates each post on `is_post_publishable()`, which reads `atmosphere_disabled`. The per-post toggle keeps deciding whether the documented recovery path will ever touch that post, while being invisible and unchangeable.

**The banner can't render in connection-only mode.** Same gate. The description says the warning isn't tied to the per-post toggle "because a host plugin can rely on the connection even when this site does not cross-post" — I agree with that, but connection-only mode forces auto-publish off, so the editor never loads the panel and the banner never appears. The pre-publish panel doesn't cover it either: `publish_decision()` returns on the auto-publish branch before it reaches the connection check, so `needs_reconnect` is false there too. The one deployment the reasoning names is the one with no signal. `test_needs_reauth_true_when_auto_publish_off` passes, but the value it pins can't reach a browser.

**The two panels contradict each other after a Disconnect.** `Client::disconnect()` drops `atmosphere_connection` and keeps `atmosphere_identity`, so `needs_reauth()` stays true. For an author, `reauth_lead_for_current_user()` returns `''`, `needsReauth` goes false, and the toggle help says "This post will be shared via ATmosphere when published." The pre-publish panel, branching on raw `needs_reauth()`, says "This post will not be shared until your site is reconnected." Same screen, opposite answers, and the sidebar is asserting the exact falsehood this is meant to kill.

Smaller stuff inline: the reconnect prompt fires on private and password-protected posts where reconnecting changes nothing, the "share it later" copy promises something only WP-CLI can do, and `ReconnectAction` has three copy branches on two surfaces with no test.

What I checked: PHPCS clean (40/40), `npm run test:unit:js` 27 passing, `lint:js` clean, and `build/` really is in sync — I rebuilt from `src/` into a scratch directory and the bundles plus both `plugin.asset.php` hashes came back byte-identical. I also chased the obvious security angles and came up empty: `reauthLead` is a hardcoded `__()` literal with no filter and React escapes it, `RECONNECT_URL` can't become a `javascript:` URL because every producer routes through `admin_url()`, and the `key_changed` sentence has no non-admin path.

Still unverified by anyone: what these panels actually look like. Happy to spin up wp-env and walk the states if that helps.

### Notes that didn't warrant their own thread

- `reconnect_url()` isn't filterable, unlike the `is_*_enabled()` family. A host plugin running its own connect UI in connection-only mode can't redirect the four reconnect surfaces to itself. Possible follow-up.
- `includes/functions.php` now imports `WP_Admin\Admin`. That file is `require_once`'d on every front-end request, and `Admin` imports back from it. Works today, but it's a two-way dependency between a global helper file and an admin class.
- `BaseControl` is used as a bare spacing wrapper with no `__nextHasNoMarginBottom`. `@wordpress/components` ^38 deprecates that bottom margin, so the spacing will shift when the default changes.
- If the session dies *after* page load and the author has "Always show pre-publish checks" off, there's still no signal at all — the banner is a page-load snapshot and `on_status_change()` bails before recording `atmosphere_publish_error`. Not a blocker here, but the original bug survives in that corner.
- `use function Atmosphere\...` imports in a file already namespaced `Atmosphere` are no-ops, and the same method calls `settings_url()` and `reauth_lead_for_current_user()` without them. Cosmetic.

includes/class-block-editor.php
pfefferle: Is this something we really want to show if auto-publish is false?

pfefferle:
Thanks, this was the review that made me stop patching and look at the shape of the thing.

All three blockers are addressed, but not in the way the individual threads describe, so I would read the diff fresh rather than thread by thread.

**The enqueue gate is gone.** You were right that it took the whole panel with it, and the backfill point settled it: hiding the toggle hid the only control over meta that `is_post_publishable()` still reads. The panel enqueues unconditionally again and the controls hide inside the component instead.

**The two-signal problem is fixed, then replaced.** `needsReauth` went back to the raw state, but the underlying issue was bigger: the sidebar was re-deriving in JavaScript what `publish_decision()` already computed in PHP, so the two drifted. There is now one `Atmosphere\share_status()` that both surfaces render, and the panel shows at most one message by precedence, so a dead connection no longer stacks with the failed share it caused.

**Private and password-protected posts** no longer get a reconnect prompt, and while I was in there: those states, plus the per-post toggle, delete an existing Bluesky record. The reasons now say so. "This post will not be shared" reads like a no-op when it is not one.

Two of your smaller ones I did not take. `BaseControl` without `__nextHasNoMarginBottom`: I checked `@wordpress/components` 38 and WP 7.0.2's bundled `components.js` and neither emits that deprecation, and the panel's existing controls already render without the flag. And the `reconnect_url()` filter is still a follow-up.

One thing you flagged that is still open and is not mine to decide: the panel sits at the bottom of the sidebar, under Move to trash, so the warning needs scrolling to see. Fixing that means a different slot, either `PluginPostStatusInfo` or an editor notice like the ActivityPub plugin uses.

Screenshots of all seven states are in the description now.

includes/class-block-editor.php
pfefferle: Gate removed. The panel enqueues unconditionally again and the controls hide inside the component, so the link, the removal warning, the error notice and the custom text field survive with sharing off. The backfill point was the decider: the toggle writes meta `is_post_publishable()` still reads, so hiding it hid the only control over whether recovery would ever touch the post.

includes/class-block-editor.php
pfefferle: Same fix covers it, though the answer changed shape. The panel loads in connection-only mode now, but it deliberately says nothing there: if ATmosphere is not the thing publishing, the connection has no bearing on the post being edited. The signal for that mode is the Connectors card, which already renders the needs-reauth state, plus the admin notice and Site Health. The pre-publish panel does not render there either, so the two agree.

includes/class-block-editor.php
pfefferle: Right, and this one went deeper than the fix. `needsReauth` is the raw state again, but the real problem was that the sidebar re-derived in JavaScript what `publish_decision()` already computed in PHP. Both now render one `Atmosphere\share_status()`, which separates site policy (is sharing on, drives whether controls render) from health (can a share succeed, drives the copy). Your disconnected-site case has a test.

includes/rest/admin/class-pre-publish-controller.php
pfefferle: Both surfaces read the same decision now, so they cannot disagree. `publish_decision()` calls `share_status()` for its site-level branches instead of keeping its own copies, and `needs_reconnect` follows whether a cause sentence is actually shown rather than the raw connection state. A non-admin on a deliberate disconnect gets no nag on either surface.

includes/rest/admin/class-pre-publish-controller.php
pfefferle: Moved below password, private and post type. While checking it I found those three do more than not-share: each one deletes an existing Bluesky record through `on_status_change()`'s removal branch. So the reasons now also say the post is on Bluesky now and will be removed, when there is a record to lose. Tests cover all four removal paths.

src/editor-plugin/utils.js
pfefferle: Dropped the promise rather than pointing authors at WP-CLI: "Sharing is on for this post, but it will not be shared while your site is disconnected from Bluesky." True without implying a button. Backfill is still the recovery, and the PR description says so.

src/editor-plugin/plugin.js
pfefferle: Gone, along with the branch that produced it. The publish-error notice only renders when the site can share, so it can never sit under a banner that already carries a call to action, and the `CAN_MANAGE` with no `RECONNECT_URL` case has its own self-contained sentence now.

src/shared/reconnect-notice.js
pfefferle: Added, following the `settings-connect` `jest.mock` precedent: administrator with a URL, administrator without one, and non-admin. The precedence that decides which message the panel shows is covered separately in `utils`, since that is where the branching moved.

tests/phpunit/tests/wp-admin/class-test-reauth-notice.php
pfefferle: Replaced with the literal, matching the siblings in that file.

tests/phpunit/tests/class-test-functions.php
pfefferle: It is true now: `needs_reconnect` follows the lead rather than the raw connection state, so the suppression reaches the pre-publish panel too. The comment stands as written.

includes/class-block-editor.php
pfefferle: Both dropped. Nothing reads `settingsUrl` or `SETTINGS_URL` any more, and the whole config block shrank: the panel now gets one `shareStatus` object rather than a set of loose flags.

src/editor-plugin/plugin.js
```diff
@@ -131,76 +182,91 @@ const EditorPlugin = () => {
...
+					</Notice>
+				</BaseControl>
+			) }
+
+			{ /* The controls, whose help text always states the outcome for
+			     this post. They follow the site's sharing policy, not the
+			     connection: with the connection down the toggle still records
+			     a preference, and `wp atmosphere backfill` reads that meta
+			     when the site recovers. */ }
+			{ SHARE_STATUS.sharing_enabled && (
```
kraftbj: The toggle and custom-text field are still gated on `SHARE_STATUS.sharing_enabled`, so with auto-publish off site-wide they're as unavailable as they were before the enqueue gate came out. The backfill point from round 1 was what closed that thread: the meta still feeds `is_post_publishable()`, which `wp atmosphere backfill` reads per post.

`share_status()`'s own reasoning keeps the toggle visible when `can_share` is false for exactly this reason. Why does `sharing_enabled` get treated differently?

Either is fine by me: render the toggle with help text saying it takes effect for backfill and if sharing is turned back on, or have backfill ignore the meta when auto-publish is off. What doesn't work is the meta staying load-bearing and invisible.

src/editor-plugin/plugin.js
```diff
@@ -83,30 +102,52 @@ const EditorPlugin = () => {
 	}
 
 	const enabled = isSharingEnabled( meta );
+	const message = panelMessage( SHARE_STATUS, {
+		enabled,
+		hasRecord: !! sharedUrl,
+		hasPublishError: !! publishError,
+		isPublished,
+		willBeUnpublished:
```
kraftbj, line 182: `willBeUnpublished` is a projection of edited attributes, which means the warning can only fire on paths that go through a save with a rendered edit in between. Two common paths don't:

**Move to Trash.** Gutenberg's trash action deletes via REST and redirects; the edited status never becomes `trash`, so nothing renders. Server-side that's the canonical cleanup transition — `test_publish_to_trash_schedules_delete` covers it. The most common unpublish gets WP's generic "Are you sure?" and nothing about Bluesky.

**Rescheduling.** `wp_insert_post()` rewrites `publish` to `future` when `post_date_gmt` is at least a minute ahead, on updates as well as inserts. Gutenberg's schedule control only edits the date. So picking a future date deletes the record and republishing later mints a new URI — likes, reposts and replies on the old one are gone. `willBeUnpublished` never looks at the date.

Folding `date_gmt` in is easy. Trash needs either a different surface or honest scoping in the copy and changelog.

src/editor-plugin/plugin.js
```diff
@@ -83,30 +102,52 @@ const EditorPlugin = () => {
 	}
 
 	const enabled = isSharingEnabled( meta );
+	const message = panelMessage( SHARE_STATUS, {
+		enabled,
+		hasRecord: !! sharedUrl,
```
kraftbj: `hasRecord: !! sharedUrl` is narrower than the thing it's guarding. `atmosphere_url` is built from `Post::META_URI` alone, while the delete path checks `has_post_records()`, which also counts `Document::META_URI` and the thread records.

With `atmosphere_should_publish_bluesky_post` filtered false the site only ever writes document records, so `hasRecord` is permanently false and the warning cannot fire on that configuration at all — while `Publisher::delete_post()` still removes the document. Same gap for a post left holding only a document record after a partial delete.

A REST field backed by `has_post_records()` would key the warning off the same fact the server uses. Keep `atmosphere_url` for the View link, which is the one thing it's actually right for.

src/editor-plugin/utils.js
```diff
@@ -19,3 +20,109 @@ import { DISABLED_META_KEY } from '../config';
...
+	}
+
+	// Nothing post-level is worth saying while the site cannot share at all:
+	// whatever went wrong or is pending is a consequence of that, and the
+	// site either explained it above or deliberately stayed quiet.
+	if ( ! shareStatus.can_share ) {
+		return null;
+	}
+
+	if ( hasPublishError && enabled ) {
```
kraftbj: `hasPublishError` returns before `pendingRemoval` is evaluated, and the two co-occur normally: `atmosphere_url` from a successful publish, then `META_LAST_PUBLISH_ERROR` from a later failed update. An author with a stale error who switches the post to private reads "Sharing to Bluesky failed. Update the post to try again." — copy that actively encourages the save that destroys the record.

The rationale above the function ("a dead connection explains both a failed share and a removal that is not happening") doesn't carry here. A past failure doesn't explain a pending irreversible removal, and it isn't superseded by one either. Removal should outrank it.

The test at `plugin.test.js:123` that names this precedence leaves `willBeUnpublished` at its default `false`, so it never constructs the conflict it's asserting about.

src/editor-plugin/utils.js
```diff
@@ -19,3 +20,109 @@ import { DISABLED_META_KEY } from '../config';
...
+	/*
+	 * Saving this change deletes the Bluesky record, and that is not
+	 * undoable: republishing later mints a new one with a new URL, so the
+	 * likes, reposts and replies on the old post are gone. Every route here
+	 * goes through the same branch in `on_status_change()`, which only fires
+	 * for a post whose *saved* status is publish, and only while the site is
+	 * connected with sharing on. "This post will not be shared" would read
+	 * like a no-op for something irreversible.
+	 */
+	if ( hasRecord && isPublished && willBeUnpublished ) {
```
kraftbj: The condition is state (record present, saved status publish, currently not shareable), not a dirty edit, so it stays true immediately after the save that caused the removal. Cron runs in a separate request, so the REST response still carries `atmosphere_url` and `status: publish`, and the panel re-renders the same future-tense "Saving this change will remove it from there, and that cannot be undone."

The reader has just saved and is being told the removal hasn't happened yet, which implies not saving again would prevent it. Worse with `DISABLE_WP_CRON`.

The notice this replaced ("still on Bluesky, will be removed the next time your site syncs") was the accurate copy for the after state. Both are needed, keyed on whether the change is saved.

includes/wp-admin/class-admin.php
```diff
@@ -552,64 +554,98 @@ public static function handle_disconnect(): void {
...
+	 * sentence is swapped instead, so a reader who cannot reconnect is
+	 * told who can rather than handed a link they cannot follow. The
+	 * notice is dismissible per page-load only, so the user is reminded
+	 * again on their next visit.
 	 *
 	 * Swaps copy when the disconnect was operator-initiated (the user
 	 * clicked Disconnect) so the message does not falsely claim "your
 	 * session has expired" for a state the user just chose.
 	 */
 	public static function maybe_render_reauth_notice(): void {
```
kraftbj, line 606: Dropping the `manage_options` guard entirely means `admin_notices` renders this for every logged-in user on every admin screen, `profile.php` included. On a WooCommerce or membership site that's every customer being told the site's Bluesky connection is broken and to go ask an administrator.

The reasoning in the docblock is about reaching the person hitting Publish, and I agree with it — but that's `edit_posts`, which still covers authors, editors and contributors and drops subscribers. Was no-gate deliberate, or did it fall out of swapping the copy instead of the gate?

includes/functions.php
```diff
@@ -624,6 +766,44 @@ function settings_url(): string {
...
+	 * the admin notice, both editor surfaces, and Site Health at its own
+	 * screen. Returning '' drops the link and leaves the prompts as plain
+	 * text, which is what happens by default when there is no screen to
+	 * link to.
+	 *
+	 * @since unreleased
+	 *
+	 * @param string $url Admin URL to reconnect at, or '' when there is none.
+	 */
+	return (string) \apply_filters( 'atmosphere_reconnect_url', $url );
```
kraftbj: Round 1 cleared the reconnect URL on the grounds that every producer went through `admin_url()`, so a `javascript:` scheme couldn't reach the React anchor. The filter removes that invariant.

The PHP sinks are fine — the admin notice runs `esc_url()` plus `wp_kses` with an `a[href]` allowlist, Site Health runs `esc_url()`. `Block_Editor::script_data()` at line 121 localizes it raw, and `reconnect-notice.js:33` renders `<a href={ RECONNECT_URL }>`. react-dom 18.3.1 only warns in dev and renders the attribute anyway.

A filter callback is trusted PHP, so this isn't a privilege boundary, but it's a footgun for a host plugin doing `add_filter( 'atmosphere_reconnect_url', fn() => get_option( 'myplugin_connect_url' ) )`. `return \sanitize_url( ... )` on this line covers every consumer at once, and the existing empty-string branches already degrade to plain text. Worth a line in the filter docblock too.

src/editor-plugin/utils.js
```diff
@@ -19,3 +20,109 @@ import { DISABLED_META_KEY } from '../config';
...
+	if ( hasPublishError && enabled ) {
+		return {
+			kind: 'publishError',
+			severity: 'error',
+			message: '',
+			action: false,
+		};
+	}
+
+	/*
```
kraftbj: "Every route here goes through the same branch in `on_status_change()`, which only fires for a post whose *saved* status is publish" — neither half is right.

`on_share_meta_changed()` hooks `added_/updated_/deleted_post_meta` for the toggle and custom-text keys and schedules `atmosphere_update_post`. That worker deletes whenever `! is_post_publishable() && has_post_records()`, with no saved-status requirement. So a draft still holding records (published while connected, connection died, reverted to draft before cleanup could run) loses them on a plain toggle save, unwarned.

src/editor-plugin/plugin.js
```diff
@@ -83,30 +102,52 @@ const EditorPlugin = () => {
...
 	   shown only to users who can open the settings page; everyone
-	   else is told who can. */
+	   else is told who can.
+
+	   These strings are kept self-contained (not built from the shared
+	   ReconnectAction lead + link) to preserve their existing
+	   translations. There is no variant that drops the call to action:
+	   this notice only renders when the site can share, so it can never
+	   sit under a banner that already carries one. */
 	const needsReconnect = publishError?.needs_reconnect;
```
kraftbj, line 210: These three branches look unreachable. `panelMessage()` only reaches `publishError` past `if ( ! shareStatus.can_share ) return null;`, and `share_status()` sets `can_share` false on every `! is_connected()` path — so reaching this code means the site is connected. The REST field computes `needs_reconnect = $reconnect_class && ! is_connected()`, which is then always false, so `needsReconnect` never selects `reconnectMessage`.

Only a stale-snapshot race gets here. Either delete them or say that's what they're for; the comment above currently claims the opposite.

includes/rest/admin/class-pre-publish-controller.php
```diff
@@ -287,29 +298,31 @@ public function get_preview( WP_REST_Request $request ) {
...
-			// external forces it off despite the user's saved preference being
-			// on — connection-only mode OR the `atmosphere_should_auto_publish`
-			// filter. Only blame settings when the stored option is itself off,
-			// so the editor never tells the author "turned off in settings" while
-			// their checkbox is checked.
-			$stored_on = '1' === (string) \get_option( 'atmosphere_auto_publish', '1' );
+		/*
+		 * Checked first, ahead of the per-post toggle and the connection
+		 * state: when nothing is being cross-posted automatically, the
+		 * document panel (home of the per-post toggle) is not even
```
kraftbj: Stale after the enqueue gate came out — the panel is enqueued now, the controls just don't render. The conclusion still holds, but the stated mechanism doesn't.

includes/functions.php
```diff
@@ -602,14 +603,155 @@ function get_reauth_reason(): string {
...
+ * same fact twice, in two severities, or with its explanation suppressed.
+ *
+ * Precedence is the whole point:
+ *
+ *  - Sharing off outranks the connection. When ATmosphere is not the thing
+ *    publishing, the connection has no bearing on the post being edited.
+ *  - Sharing forced off from outside says nothing at all: a host plugin owns
+ *    that experience and the reader cannot act on the arrangement.
+ *
+ * @since unreleased
```
kraftbj: Two description paragraphs sit below `@since unreleased`, so parsers attach them to the tag. Prose above the tag block.

Same file around line 590: `reauth_reason_lead()`'s docblock still names "the admin reconnect notice and the Site Health test" as its readers. The notice goes through `reauth_lead_for_current_user()` now, and the two editor surfaces are new indirect consumers. `Health_Check::reauth_description()` repeats the claim.

src/config.js
```diff
@@ -35,12 +35,13 @@ export const PREVIEW_PATH =
...
- * URL of the ATmosphere settings page, for reconnect prompts.
+ * Where the reconnect prompts should link: the settings page, the
+ * Connectors screen when the settings page is hidden (connection-only
+ * mode), or empty when neither exists.
  *
  * @type {string}
  */
-export const SETTINGS_URL =
-	options.settingsUrl || 'options-general.php?page=atmosphere';
+export const RECONNECT_URL = options.reconnectUrl || '';
```
kraftbj, line 44: `SHARE_STATUS` is documented as `{ state, message, severity, action, can_share }` but omits `sharing_enabled`, which is the key deciding whether the toggle, the custom-text field and the whole pre-publish panel render. `CAN_MANAGE` still says the destination is the settings page; it's `reconnect_url()` now.

kraftbj (changes requested):
The reauth work is in good shape now. Round 1's blockers are fixed, `share_status()` is a better shape than the loose flags it replaced, and splitting site policy from connection health is what makes the two surfaces stop arguing with each other. I verified all four: the connection check moved below post type, password and private; `needs_reconnect` follows the emitted cause; and a non-admin on a deliberately disconnected site now gets consistent copy on both panels.

One is only partly fixed, and the removal warning that arrived during review isn't ready.

**The toggle is still gone when auto-publish is off.** The enqueue gate went away and the link, the error notice and the removal warning came back with it, but the toggle and the custom-text field are still behind `SHARE_STATUS.sharing_enabled` at `plugin.js:224`. That was the backfill argument from round 1, and it's the reason the thread got resolved, so I want to flag that the fix doesn't quite reach it. Either the toggle renders with help text that explains what it's for, or `backfill` stops reading the meta when auto-publish is off. Right now the meta still decides whether recovery touches the post and there's still no way to see or change it.

**The removal warning has more holes than coverage.** It's a destructive-action confirmation for something irreversible on a remote service, so the bar is higher than for the rest of this. As written it misses:

- **Move to Trash**, the most common way to unpublish. Trash isn't a save, so the edited status never becomes `trash` and no render happens in the warning state. Server-side `publish` to `trash` is the canonical cleanup transition.
- **Rescheduling.** Core rewrites `publish` to `future` when the date is a minute or more ahead. `willBeUnpublished` has no date input at all.
- **Document-only sites.** `hasRecord` is `!! atmosphere_url`, which is built from `Post::META_URI` only. The delete path uses `has_post_records()`, which also counts `Document::META_URI`. With `atmosphere_should_publish_bluesky_post` false there is never a bsky URI, so the warning can't fire on that whole configuration.
- **Meta-only saves on a non-published post.** `on_share_meta_changed()` schedules `atmosphere_update_post` with no status requirement, and that worker deletes when the post isn't publishable but has records. The comment at `utils.js:106` says `on_status_change()` is the only route and requires a saved status of publish; neither half holds.

And where it does fire, it fires twice. The condition is state, not a dirty edit, so straight after the save that triggered the removal all three inputs are still true and the panel repeats "Saving this change will remove it from there, and that cannot be undone" about something already in flight. A reader would reasonably conclude the save didn't take.

A warning that misses trash, rescheduling and an entire site mode, then cries wolf on the two paths it does cover, is worse than the honest silence we had. I'd split it out and land the reauth half, which stands on its own and is genuinely good.

**Two smaller things I'd want either way.** The admin notice lost its capability gate entirely, so it now renders on every wp-admin screen for every logged-in user — subscribers on a membership site included. The reasoning in the docblock ("the person hitting Publish is often not the one who can reconnect") argues for `edit_posts`, not for no gate. And the new `atmosphere_reconnect_url` filter invalidates the escaping argument I made in round 1: PHP sinks still call `esc_url()`, but `Block_Editor::script_data()` localizes the value raw and React renders `<a href={ RECONNECT_URL }>` verbatim. A filter is trusted PHP so this isn't a remote attack, but it's a footgun for a host plugin piping in an option value. One line.

Checked: PHPCS clean (40/40), 38 JS tests pass, `lint:js` clean, and `build/` is in sync (rebuilt from `src/` into a scratch directory, byte-identical including both asset hashes). CI green. Thanks for the screenshots — the state-by-state walkthrough answers the verification gap from last time.

### Notes that didn't warrant their own thread

- `share_status()` has no direct PHPUnit test, and `severity`/`action`/`state` are never asserted anywhere in `tests/` — those three fields decide notice level and whether the reconnect link renders.
- No test pins that the sidebar and the pre-publish panel agree for a given state, which is the class of bug round 1 was about.
- `sharing_off_external` fires for any `atmosphere_should_auto_publish` filter returning false, not just connection-only mode, and the pre-publish panel then removes itself with no explanation. The comments justify the silence on connection-only grounds; the condition is broader.
- An admin on a never-connected site gets no action link in the pre-publish panel, though `reconnect_url()` is already resolved and every other unhealthy state offers one.
- Quick Edit, bulk edit, REST, XML-RPC and CLI all hit the same cleanup branch with no warning. Inherent to a client-side guard — worth being a conscious decision, and worth keeping the changelog line narrower than "warns you before a change removes a post from Bluesky."
- `BaseControl` as a bare margin wrapper is still deprecated in components ^38.

src/editor-plugin/plugin.js
pfefferle: Toggle and custom text now render in every state, and `shareHelpText()` grew a variant that says what the control is for when automatic sharing is off: the preference is kept for `wp atmosphere backfill` and for if sharing is turned back on. Went that way rather than changing backfill, because making it ignore the meta would silently start backfilling posts an author had switched off.

While in there I also narrowed the whole-panel silence from `! sharing_enabled` to `sharing_off_external`, which covers your other note. An owner who turned sharing off in settings keeps the panel and the toggle now; only the host-plugin case stays quiet.

src/editor-plugin/plugin.js
pfefferle: Rescheduling is folded in through `isEditedPostBeingScheduled()`, which applies core's own rule (`date - ONE_MINUTE_IN_MS` in the future), so it matches what `wp_insert_post()` does instead of me reimplementing the threshold.

Trash I could not reach from this panel. The REST delete redirects with no render in between, so no client-side guard here catches it. Scoped it honestly instead: the changelog now says "warns you in the editor before saving a change that removes a post from Bluesky", and the code comment names trash, Quick Edit, bulk edit, REST, XML-RPC and CLI as outside the guard.

src/editor-plugin/plugin.js
pfefferle: New `atmosphere_has_record` REST field, backed by the same `has_post_records()` the cleanup path calls, so the warning and the deletion key off one fact. `atmosphere_url` stays for the View link, which is the thing it is right for. Two tests: the AT-URI case, and a document-only post where the URL is empty but the flag is true.

src/editor-plugin/utils.js
pfefferle: Removal outranks the failure now. The test was the worse half of this: it asserted a precedence it never constructed. It sets both `hasPublishError` and `willBeUnpublished` now, so the conflict is real, and there is a second test for a failure with nothing being removed.

src/editor-plugin/utils.js
pfefferle: Two variants keyed on `isEditedPostDirty()`. Unsaved gets the future tense, saved gets the old "still on Bluesky, will be removed the next time your site syncs" wording you pointed at. Test walks both.

includes/wp-admin/class-admin.php
pfefferle: Not deliberate, it fell out of swapping the copy. `edit_posts` now, which is the gate the docblock reasoning actually argues for. The existing test was running as user 0, so it was passing for the wrong reason; it runs as a contributor now, plus two new ones pinning that a subscriber and a logged-out visitor get nothing.

includes/functions.php
pfefferle: `sanitize_url()` on the return, so it covers every consumer at once rather than each sink. Docblock says so, and there are two tests: a `javascript:` scheme degrades to the existing empty-string branch, and a normal admin URL with a query string survives intact.

src/editor-plugin/utils.js
pfefferle: Comment was wrong on both halves, rewritten. It names `on_share_meta_changed()` and its worker now, and the saved-status requirement is gone from the condition: holding records at all is what matters, so a draft that kept its records warns too.

src/editor-plugin/plugin.js
pfefferle: They are reachable, but only on a stale snapshot, which is the bug this panel is about: `SHARE_STATUS` is fixed at page load while `atmosphere_publish_error` refreshes on every save, so a connection dying mid-session leaves `can_share` true while the REST field starts reporting `needs_reconnect`. Kept them and rewrote the comment to say that instead of claiming the opposite.

includes/rest/admin/class-pre-publish-controller.php
pfefferle: Rewritten. It no longer leans on the enqueue gate, just on the fact that with automatic sharing off neither the toggle nor a reconnect changes the outcome.

includes/functions.php
pfefferle: Prose moved above the tag block. `reauth_reason_lead()` now names its real readers, Site Health directly and the notice plus both editor surfaces through `reauth_lead_for_current_user()`, and the `Health_Check` comment says the same thing.

src/config.js
pfefferle: `sharing_enabled` documented, and `CAN_MANAGE` no longer claims the settings page is the destination. Also spelled out which of the two flags feeds the help text and which one decides the panel removing itself.

jeherve (dismissed):
This looks good and tests well for me, no blockers. 🚢

jeherve (dismissed):

PR #244, opened 2026-08-26
Handle Bluesky rate limits when sharing and backfilling

### LINKED ISSUE #241: Feature Request: handle Bluesky rate limits in the backfill path
### What

Make the plugin aware of the PDS rate limits: read the rate limit headers, treat a 429 as its own error instead of a generic PDS failure, and let `wp atmosphere backfill` pace itself or stop cleanly instead of failing through the rest of the queue.

### Why

A partner is about to backfill around 80,000 posts, and as far as I can tell we do not handle this at all right now.

`API::request()` (`includes/class-api.php`) retries exactly two things: the DPoP nonce challenge and a 401 token refresh. It never looks at the `ratelimit-remaining` / `ratelimit-reset` headers, and a 429 falls straight through `is_success_status()` into a generic `atmosphere_pds` error.

`Backfill_Command` publishes one post at a time with nothing in between. `--batch` is only progress ticks and object cache eviction, its own docblock says so. And the loop only breaks early on `atmosphere_not_connected`, so a 429 just prints a warning, counts as an error, and moves to the next post at full speed.

The numbers, from https://bsky.network/docs/rate-limits/: 5,000 points per hour and 35,000 per day, a create costs 3 points. Every publish writes two records (the `app.bsky.feed.post` and the `site.standard.document`), so 6 points per post. That is 833 posts per hour and 5,833 per day. Unthrottled, the loop burns the hourly budget in maybe 10 to 20 minutes, and then every remaining post gets a 429.

Nothing is lost when that happens. A failed post never gets `Document::META_URI`, so it stays in the unsynced set and a re-run picks it up. But the operator has no way to see it coming and no way to slow down.

We already classify 429 as transient in `Atmosphere::is_transient_publish_error()`, with a 60/300/900 second ladder. That is only wired to the `atmosphere_publish_post` cron though. The CLI calls `Publisher::publish_post()` directly and never sees the ladder.

### How

Rough plan, happy to hear other ideas:

- In `API::request()`, keep the rate limit headers around and return a dedicated `atmosphere_rate_limited` error code with the reset timestamp in the error data. No sleeping there, that code also runs inside normal web requests.
- In the retry ladder, when the error carries a reset time, wait for that instead of the fixed delays.
- In `Backfill_Command`, stop the run on a rate limit error the way it stops on a lost connection, and add a `--throttle` flag so a long run can pace itself.

I am not sure yet whether the throttle should be reactive (wait for the 429) or proactive (count points and pause before hitting the wall). Proactive is nicer, but it means tracking spend across runs, which is more state than I would like.

pfefferle:
Fixes #241

## Proposed changes:

* `API::request()` now reads the `ratelimit-*` headers and returns a dedicated `atmosphere_rate_limited` error for a 429, with the reset timestamp in the error data. No sleeping in there, that code also runs during normal web requests.
* A rate-limited share waits for the window the PDS reported instead of the ladder's own 60/300/900 step. The header value is capped at a day, so a broken or hostile `ratelimit-reset` cannot park a queued share weeks out, but the cap only applies to the header, it never shortens the ladder.
* `wp atmosphere backfill` stops when it hits the limit, the same way it stops on a lost connection, and says roughly when the window reopens. Nothing is lost, a failed post never gets `Document::META_URI` so a re-run continues where it stopped.
* New `--throttle=<seconds>` flag to pace a long run. Skipped posts and dry runs cost no budget, so they are not paced.

One thing that was not obvious to me: a long post publishes as a thread, and there the 429 does not always arrive as the top level error code. If a reply create spends the last of the budget, the compensating rollback draws on the same spent budget and gets a 429 too, so the failure comes back as `atmosphere_thread_rollback_failed`. The abort looks one level down for that now.

What this does not solve:

* The throttle is reactive and dumb, a fixed sleep. It does not count points and it does not slow down as the budget runs low. `API::last_rate_limit()` keeps the last reading around so we could do that later, but nothing reads it yet. I went with this because proactive pacing means tracking spend across runs, which is the state I did not want to add (see the issue).
* A thread whose rollback was itself rate limited still leaves orphan records on the PDS and is still not retried automatically. A retry would mint fresh TIDs next to the live partial records and give you a duplicate post, so that case keeps needing a human.
* Only `ratelimit-*` is read, not `Retry-After`. A self hosted PDS behind Cloudflare or nginx would send the latter, and we would fall back to the plain ladder there. Probably worth adding, I just did not want to guess at the shape without a real response to look at.

### Other information:

## Testing instructions:

`npm run env-test` and `composer lint` both pass. New coverage is in `class-test-api.php`, `class-test-atmosphere.php` and `tests/phpunit/tests/cli/class-test-backfill-command-invoke.php`.

A real 429 is hard to trigger on purpose, so for a manual check drop this into a mu-plugin:

```php
add_filter( 'atmosphere_pre_apply_writes', function () {
	return new WP_Error(
		'atmosphere_rate_limited',
		'Rate limit exceeded.',
		array(
			'status' => 429,
			'reset'  => time() + 30 * MINUTE_IN_SECONDS,
		)
	);
} );
```

* With the filter active and a few unsynced posts on the site, run `wp atmosphere backfill`. It should try one post, warn that the limit resets in about 30 minutes, and stop, instead of walking the whole queue and printing the same warning for every post.
* Swap the error code in the filter for `atmosphere_thread_rollback_failed` and wrap the 429 above as `'original_error'` in the data array. Same result, the run still stops.
* Publish a post with the filter active, then check `wp cron event list | grep atmosphere_publish_post`. The retry should sit about 30 minutes out, not 60 seconds.
* Remove the filter. `wp atmosphere backfill --throttle=3 --dry-run` should finish right away, a dry run spends no budget. Without `--dry-run` it should pause about three seconds between posts.

### DIFF
--- a/.github/changelog/add-rate-limit-awareness
+++ b/.github/changelog/add-rate-limit-awareness
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Handle Bluesky rate limits when sharing: a share that hits the limit now waits for it to lift before retrying, and bulk syncs can be paced so they stay under it.
--- a/includes/class-api.php
+++ b/includes/class-api.php
@@ -136,7 +136,7 @@ class API {
 			$body = array();
 		}
 
-		self::capture_rate_limit( $response );
+		$rate_limit = self::capture_rate_limit( $response );
 
 		// Persist any nonce the server sends back.
 		$response_nonce = \wp_remote_retrieve_header( $response, 'dpop-nonce' );
@@ -242,7 +242,7 @@ class API {
 		 * Waiting is the caller's decision.
 		 */
 		if ( 429 === $status ) {
-			return self::rate_limited_error( $body );
+			return self::rate_limited_error( $body, $rate_limit );
 		}
 
 		if ( ! is_success_status( $status ) ) {
@@ -263,13 +263,17 @@ class API {
 	 * A response without any of the headers leaves the previous reading
 	 * in place rather than clearing it: an intermediary that strips them
 	 * on one call should not erase what the last real answer told us.
+	 * The return value is still null in that case, so a caller reasoning
+	 * about *this* response cannot mistake a stale reading for a fresh
+	 * one.
 	 *
 	 * @since unreleased
 	 *
 	 * @param array $response Raw `wp_remote_*` response.
-	 * @return void
+	 * @return array|null Snapshot for this response, or null when it
+	 *                    carried no rate-limit headers.
 	 */
-	private static function capture_rate_limit( array $response ): void {
+	private static function capture_rate_limit( array $response ): ?array {
 		/*
 		 * `wp_remote_retrieve_header()` returns an array when a header
 		 * repeats, so each read is guarded by a type check rather than a
@@ -288,10 +292,12 @@ class API {
 		);
 
 		if ( null === $snapshot['limit'] && null === $snapshot['remaining'] && null === $snapshot['reset'] ) {
-			return;
+			return null;
 		}
 
 		self::$last_rate_limit = $snapshot;
+
+		return $snapshot;
 	}
 
 	/**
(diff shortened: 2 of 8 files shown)


jeherve (approved):
This is looking good in my tests, I don't have anything major to suggest. 🚢

PR #246, opened 2026-08-27
Warn in the editor when the site isn't connected to Bluesky

### LINKED ISSUE #245: The editor panel should tell you when your site isn't connected to Bluesky

jeherve:
Fixes #245

## Proposed changes:

Right now the ATmosphere panel in the block editor looks the same whether or not your site is connected to Bluesky. With sharing on (the default), the "Share this post" toggle says "This post will be shared via ATmosphere when published", even when the connection has lapsed or was never set up. So you publish the post, nothing goes out to Bluesky, and you only find out later, which isn't a fun surprise :)

This adds a warning under the toggle when the site isn't connected: "Your site isn't connected to Bluesky, so this post won't be shared", with a link to reconnect on the settings page. Anyone who can't manage the connection gets the same warning without the link, ending in "Ask an administrator to connect it" instead.

Under the hood, `Block_Editor::script_data()` now exposes an `isConnected` flag from `Atmosphere\is_connected()` (which is also false for a lapsed `needs_reauth` connection). A new pure helper, `shouldShowNotConnectedNotice()`, decides when to show the notice: only when sharing is on for the post, and not when a publish error is already telling you to reconnect.

I considered just hiding the panel until you're connected, but the toggle stores a per-post preference that still matters once you reconnect, so hiding it would stop you from setting share intent on drafts while a connection is being sorted out. A notice keeps the panel useful and fills the gap. The pre-publish panel was already connection-aware, so this mostly brings the document sidebar panel in line with it.

### Other information:

- [x] Have you written new tests for your changes, if applicable? Unit tests cover the new `shouldShowNotConnectedNotice()` helper.

One note on tooling: the JS test/lint runner is blocked in my local environment, so I couldn't run `npm run lint:js` / `test:unit:js` myself. `composer lint` passes, the full PHP suite passes (1121 tests), and I verified the behavior on a live site (below). CI will run the JS lint and unit tests.

## Testing instructions:

1. Connect your site to Bluesky under Settings -> ATmosphere, then start a new post. The ATmosphere panel shows "Share this post" on, with no warning.
2. Disconnect the site. Quickest way over WP-CLI:
   - Never connected: `wp option delete atmosphere_connection`
   - Lapsed token: `wp option update atmosphere_connection '{"access_token":"x","needs_reauth":true}' --format=json`
3. Reload the editor. Under the toggle you should see an amber warning: "Your site isn't connected to Bluesky, so this post won't be shared. Connect on the settings page."
4. Reconnect (or restore the option) and reload. The warning is gone.
5. Log in as an author (not an admin) and repeat step 3. The warning shows without the settings link, ending in "Ask an administrator to connect it".

I ran this on a live Jurassic Ninja site across the connected / disconnected / lapsed-`needs_reauth` states, and it behaved as expected in each.

### Screenshots

Before (disconnected, current release): the panel still shows "Share this post" on with "will be shared when published", so there's no hint the post won't go out.

After (disconnected, this branch): the amber warning shows under the toggle.

(I'll add the before/after screenshots to the PR.)

### Changelog entry

Already committed in `.github/changelog/add-editor-not-connected-notice` (Minor / Added), so no need to check the auto-create box below.


pfefferle:
This and other states are already addressed in this PR https://github.com/Automattic/wordpress-atmosphere/pull/233

jeherve:
Ah, I had missed that PR! Closing this, I'll review the other one.

PR #263, opened 2026-09-03
Offer a reconnect link when an expired Bluesky session blocks a share

pfefferle:
## Proposed changes:

When a share failed because the Bluesky session had expired, the editor and the posts list showed the auth server's raw error text and no way to act on it, while the site-wide notice was already telling the user to reconnect. The two surfaces contradicted each other in exactly the state they exist to explain.

The cause is that `Client::refresh()` returned one error code, `atmosphere_refresh`, for two unrelated failures:

* **permanent** — `invalid_grant` / `invalid_client` / `unauthorized_client`, the branch that already calls `mark_needs_reauth()`. Only a reconnect fixes it.
* **transient** — an auth server 5xx or rate limit, where the refresh token may be untouched and the next attempt can succeed.

Both consumers of that code classify through `is_reconnect_error()`: `is_transient_publish_error()` decides whether the retry ladder runs again, and `get_publish_error()` sets the `needs_reconnect` flag that the editor panel and the posts-list column read. With one code covering both branches, either answer is wrong for the other branch — which is why simply adding `atmosphere_refresh` to `RECONNECT_ERROR_CODES` is *not* the fix: it would take a transient 5xx off the retry ladder and drop posts the auth server would have accepted a minute later.

So the permanent branch now returns `atmosphere_needs_reauth` — the same code and wording `access_token()` already returns once the flag is set — and the transient branch keeps `atmosphere_refresh`.

**Retry behaviour is unchanged on both branches.** The permanent case returned HTTP 400, which `is_transient_publish_error()` already classified non-transient; it now reaches the same conclusion by code instead of by status.

The auth server's `error_description` is dropped from the author-facing message on purpose: it reaches authors verbatim in the posts-list column, where `invalid_grant` tells them nothing they can use. It is written to the debug log instead — that message is also all `log_cron_error()` ever sees, and without it `invalid_grant` (token consumed or revoked) would be indistinguishable from `invalid_client` / `unauthorized_client` (the client registration being rejected), which is a different problem with a different fix.

### Other information:

Three tests, written before the fix — one red on the bug, two guarding the regression described above:

* `test_permanent_failure_returns_reconnect_class_error` — the failing case
* `test_transient_failure_is_not_reconnect_class` — pins the transient branch as retryable and unflagged
* `test_refresh_failure_classes_split_across_the_retry_ladder` — pins both codes to opposite sides of the ladder
* `test_permanent_refresh_failure_surfaces_reconnect_on_the_post` — end-to-end: a stale `expires_at` sends a real publish through `refresh()`, and the post ends up carrying `needs_reconnect` rather than the auth server's text. Verified red without the `class-client.php` change.

## Testing instructions:

**Automated**

```bash
npm run env-test -- --filter='test_permanent_failure_returns_reconnect_class_error|test_transient_failure_is_not_reconnect_class|test_refresh_failure_classes_split_across_the_retry_ladder'
```

All three pass on this branch. Reverting the `class-client.php` change alone turns the first one red.

Full suite and lint both clean: 1335 tests, PHPCS 42/42.

**Manual**

1. Connect a Bluesky account under Settings → ATmosphere.
2. Revoke the plugin's access from the Bluesky side, so the stored refresh token is genuinely dead.
3. Force the access token to look expired, which sends the next publish through `refresh()`:
   ```bash
   wp option patch update atmosphere_connection expires_at 1
   ```
4. Publish a post and let the share worker run.
5. Open that post in the editor, and look at its row in Posts → All Posts.

*Before:* the notice reads "Sharing to Bluesky failed. Update the post to try again." with the auth server's `invalid_grant` text beneath it, and the posts-list column shows that raw string. Updating the post can never help.

*After:* the notice reads "Sharing to Bluesky failed because your site is no longer connected to Bluesky." with a link to reconnect, and the posts-list column explains the expired session. Both now agree with the site-wide reconnect notice.

To confirm the transient branch still retries, replay step 4 with the auth server returning a 503 instead — the ladder should schedule another attempt and the notice should say the site will retry automatically.

### Changelog entry

Committed on the branch as `.github/changelog/fix-expired-session-reconnect-prompt`.


jeherve (approved):
This should be good to merge. 🚢

PR #272, opened 2026-09-15
Stop the Bluesky login from expiring every two weeks

pfefferle:
Fixes the two-week disconnects behind #233, #234 and #263.

## Proposed changes:

The plugin was a public OAuth client. Bluesky caps a public client's session at two weeks, no matter how often the token is refreshed (see [client.ts](https://github.com/bluesky-social/atproto/blob/main/packages/oauth/oauth-provider/src/client/client.ts) in the reference server). That is why every site had to reconnect every two weeks, and why #233, #234 and #263 could only make the expiry visible.

Confidential clients get two years, with a refresh at least every three months. So the plugin now:

* generates a per-site signing key (encrypted like the tokens) and publishes its public half in a new client metadata document at `atmosphere/v2/client-metadata`. That URL is the new `client_id`.
* sends a `private_key_jwt` assertion on PAR, code exchange and refresh.
* keeps serving the old public-client document at `atmosphere/v1/client-metadata`. Sessions created before this update have no stored `client_id` and keep refreshing there without an assertion, so nothing breaks until the user reconnects. The old endpoint has to stay until those sessions are gone.
* no longer forces a reconnect on `invalid_client` / `unauthorized_client`. Reconnecting the same client cannot fix a rejected client registration, so the session is kept and Site Health says what is wrong instead.
* adds two Site Health checks: a rejected client configuration (critical) and no renewal for more than a day (recommended, usually a dead WP-Cron).

I fixed three things in the first version that would have broken every new connection: the v2 route fataled when called through the REST server, the assertion `aud` was the token endpoint instead of the issuer, and `token_endpoint_auth_signing_alg` was missing from the metadata. All three are pinned by tests now.

After the review, the plugin also revokes the refresh token as the client the session belongs to when you disconnect, and replaces a signing key that no longer decrypts (rotated salts) as soon as no live session depends on it, so reconnecting works without touching the database.

A rejected client no longer retries the refresh on every request: refreshing pauses for five minutes after a client-configuration or signing-key failure, and a publish that lands in that window is retried by the normal ladder.

### Other information:

## Testing instructions:

* Update a connected site to this branch. Nothing should change: `wp-env run cli wp option get atmosphere_connection` has no `client_id`, and publishing still works.
* Open `/wp-json/atmosphere/v2/client-metadata`. It should show `token_endpoint_auth_method: private_key_jwt`, `token_endpoint_auth_signing_alg: ES256` and a `jwks` with one key that has no `d` member.
* Open `/wp-json/atmosphere/v1/client-metadata`. It still says `none` and has no `jwks`.
* Disconnect and connect again. The connection now stores a `client_id` ending in `/v2/client-metadata`, and `wp-env run cli wp cron event run atmosphere_refresh_token` renews it. Site Health → Info → ATmosphere shows the renewal time.
* Site Health → Status: run `wp-env run cli wp option patch update atmosphere_refresh_status last_success 1` and reload, the "not renewed" recommendation appears. Then set `last_error` to `invalid_client` and `last_failure` to the current timestamp, the critical entry appears and no reconnect notice is shown.

### Changelog entry

Two entries are already on the branch.


includes/oauth/class-client.php
```diff
@@ -593,6 +616,11 @@ public static function handle_callback( string $code, string $state ): true|\WP_
 			'code_verifier' => $verifier,
 		);
 
+		$token_body = Client_Authentication::sign_request( $token_body, $issuer );
+		if ( \is_wp_error( $token_body ) ) {
+			return $token_body;
+		}
```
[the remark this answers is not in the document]
pfefferle: Fixed. The pending flow now remembers which client started it. A callback from a transient without a client ID is exchanged as the public client and the session is stored as legacy, so an authorization in flight during the update still completes. Covered by a test.

kraftbj (commented):
Two blockers, then smaller stuff. I checked the assertion against the atproto OAuth spec and it lines up: inline `jwks` is allowed, `aud` is the issuer, `iat`/`jti` are there, `private_key_jwt` is declared. The legacy split is careful too, including a flow started under the old version mid-upgrade.

**Disconnect stops revoking the refresh token.** `revoke_refresh_token()` wasn't updated. `includes/oauth/class-client.php:2181` still sends `self::client_id()`, which is now the v2 ID, and never attaches an assertion. For a legacy session that's a mismatched client for a token minted under v1, so the server rejects it. For a confidential session the ID is right but RFC 7009 defers client auth to RFC 6749 section 2.3, which requires it. Either way the revocation fails silently (best-effort cron, `debug_log` only): the user hits Disconnect, we wipe the local row, and the token stays live on the PDS. That's worse than before this PR, since the whole point here is that those tokens now live a long time. The worker only receives ciphertexts, so the session's `client_id` needs to ride along in the cron args, with a default for events already queued.

**A key that stops decrypting bricks the plugin.** `Client_Authentication::key()` returns a `WP_Error` and never regenerates, and nothing clears `KEY_OPTION` outside uninstall. `Encryption` falls back to `AUTH_KEY . AUTH_SALT` by default, and its own docblock already calls out salt rotation as something that happens to real sites. After a rotation: tokens are undecryptable, the connection flags `needs_reauth`, and the documented fix is "reconnect" — but reconnect goes through `authorize_via_par()` to `sign_request()` and hits the same error. The v2 metadata endpoint 500s too. The only way out is deleting the option by hand.

I get why you don't want to silently replace the key (the spec binds a live session to its `kid`/`jkt`), but in that state the sessions are already dead, so there's nothing left to protect. Regenerating when there's no usable connection, or a Site Health action that resets it, would both work.

Smaller:

- `renewal_is_stale()` only looks at `last_success`, so any failure that isn't `invalid_grant` or a client-config error (network, 5xx, DPoP) ages it out and we tell the admin to "Configure a real server cron" on a site whose cron is fine. Worth checking `last_failure > last_success` and saying something else in that case.
- Nothing strips `jwks_uri` from the filtered metadata. `pinned_fields()` re-merges `jwks`, but the spec says a confidential client supplies one or the other and not both, so a host plugin's filter can quietly break client auth. One `unset()` next to the re-merge.
- `test_sign_request_attaches_an_assertion` decodes the claims but never verifies the signature, and nothing asserts the assertion's `kid` matches the published JWKS. If those ever drift, every connect and refresh fails and the suite stays green. An `openssl_verify()` round-trip against the public key would pin it.
- `is_transient_publish_error()` doesn't know about `atmosphere_client_authentication` / `atmosphere_client_authentication_key`, so an unreadable key burns three publish retries per post.
- The legacy-reconnect nudge lands on the settings page and Site Health, but not the connectors card or the editor surfaces. In connection-only mode the settings page is hidden, so Site Health is the only place left to see it.
- `use function Atmosphere\is_legacy_connection` is out of alphabetical order in `class-health-check.php:33`.
- The race comment in `key()` promises a bit more than WordPress gives: `add_option()` uses `INSERT ... ON DUPLICATE KEY UPDATE`, so a genuine tie clobbers rather than the first writer winning.

`composer lint` is clean. I haven't run the PHP suite.

pfefferle:
Thanks @kraftbj, both blockers were real.

- Revocation: the session's `client_id` now rides along in the cron args (empty for a legacy row, and for events queued before this change), and the worker signs the revocation for a confidential session, on the nonce retry too. ab90500
- Key rotation: an unreadable or malformed signing key is discarded and regenerated as soon as no live session matches the current key material, so reconnecting works after a salt rotation without touching the database. While a matching session still exists the key stays put and the error is reported. 3f3935e
- The smaller ones are in 34a42b9: a failed renewal gets its own Site Health state instead of the cron advice, `jwks_uri` is stripped next to the re-merge, the assertion is verified with `openssl_verify` against the published JWKS including the `kid`, the two key error codes are permanent for the publish ladder, the connectors card shows the reconnect hint, and the import order and the race comment are fixed.

I left the editor surfaces out of the nudge on purpose: they explain a post's share state, and in connection-only mode cross-posting is off anyway, so a reconnect hint there would be noise.

includes/oauth/class-client-authentication.php
```diff
@@ -0,0 +1,214 @@
...
+	 * @return bool
+	 */
+	private static function valid_key( $key ): bool {
+		return \is_array( $key )
+			&& 'EC' === ( $key['kty'] ?? '' )
+			&& 'P-256' === ( $key['crv'] ?? '' )
+			&& \is_string( $key['x'] ?? null )
+			&& \is_string( $key['y'] ?? null )
+			&& \is_string( $key['d'] ?? null );
+	}
```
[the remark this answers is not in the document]
pfefferle: Fixed in 5a18376. Every member is now strictly base64url-decoded and must be exactly 32 bytes, so a decryptable but corrupt row goes through the same replace-or-report path as an unreadable one instead of being published. Covered by a data-provider test for empty, non-base64url, too-short and too-long members.

includes/oauth/class-client.php
```diff
@@ -2059,12 +2187,27 @@ public static function revoke_refresh_token(
...
+
 		$body = array(
 			'token'           => $refresh_token,
 			'token_type_hint' => 'refresh_token',
-			'client_id'       => self::client_id(),
+			'client_id'       => $confidential ? $client_id : self::legacy_client_id(),
 		);
 
+		if ( $confidential ) {
+			$body = Client_Authentication::sign_request( $body, $auth_server_issuer );
```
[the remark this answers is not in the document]
pfefferle: Updated the description: revocation and key regeneration are part of the PR now, only the refresh backoff stays on the follow-up list.

includes/oauth/class-client-authentication.php
```diff
@@ -0,0 +1,227 @@
...
+		}
+
+		foreach ( array( 'x', 'y', 'd' ) as $member ) {
+			$value = $key[ $member ] ?? null;
+			if ( ! \is_string( $value ) || '' === $value ) {
+				return false;
+			}
+
+			$bytes = \base64_decode( \strtr( $value, '-_', '+/' ), true ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.obfuscation_base64_decode
+			if ( false === $bytes || 32 !== \strlen( $bytes ) ) {
```
kraftbj, line 251: This rejects about 1 in 70 of the keys that `DPoP::generate_key()` itself produces. `openssl_pkey_get_details()` strips leading zero bytes from the EC components, so `x`, `y` or `d` comes back as 31 bytes about 1/256 of the time. I ran generate → base64url → this check 2,000 times on PHP 8.5, and 29 keys (1.45%) failed.

On a fresh site, the first connect then fails with "The OAuth client-authentication key could not be saved." The key is stored, re-read, rejected here, and deleted because nothing is connected yet. A retry works, but this is the flow the PR exists for.

It's worse for the legacy sessions the reconnect nudge targets. Those sessions are still connected, so `session_bound_to_key()` returns true, the bad key stays in place, and every reconnect attempt fails until they disconnect (see the comment on that function below).

`jwk_to_pem()` already left-pads to 32 bytes and only rejects members that are too long, which is why signing never tripped over this. I'd pad in `generate_key()`:

```php
'x' => self::base64url( \str_pad( $ec['x'], 32, "\0", STR_PAD_LEFT ) ),
```

Do the same for `y` and `d`, and keep the strict check here. That also changes the DPoP JWK for newly generated keys, so it's worth a real connect and publish against a PDS.

This also makes `test_corrupt_key_members_are_rejected` and `test_unreadable_key_is_regenerated_without_a_session` flaky at the same rate, because both generate a fresh key inside `jwks()`. A test that feeds in a 31-byte component would pin this down reliably.

includes/oauth/class-client-authentication.php
```diff
@@ -0,0 +1,227 @@
...
+	 * @return bool
+	 */
+	private static function session_bound_to_key(): bool {
+		if ( ! is_connected() ) {
+			return false;
+		}
+
+		$fingerprint = (string) ( get_connection()['key_fingerprint'] ?? '' );
+
+		return '' === $fingerprint || \hash_equals( Encryption::key_fingerprint(), $fingerprint );
```
kraftbj, line 226: I think Copilot's point on this one still stands. A legacy session never signs with this key, but it counts as bound here because its `key_fingerprint` is either empty or matches. If the key row goes bad while a legacy session is connected, the key is never replaced, the v2 metadata endpoint returns a 500, and the reconnect the new nudge asks for fails with "Disconnect and connect again."

Scoping this to confidential sessions fixes it:

```php
$conn = get_connection();
if ( ! is_connected() || empty( $conn['client_id'] ) ) {
	return false;
}
```

Heads up: `seed_connection()` in the tests never sets `client_id`, so `test_unreadable_key_is_an_error_while_a_session_uses_it` is actually asserting the legacy case right now. That helper needs a `client_id`, plus a new test where a legacy row gets its key replaced.

includes/oauth/class-client.php
```diff
@@ -1134,6 +1204,26 @@ private static function refresh_locked( array $conn ): true|\WP_Error {
...
 			}
 
+			/*
+			 * A client-registration failure is not repaired by authorizing
+			 * the same client again. Keeping the session intact avoids a
+			 * misleading "disconnected" state and leaves a still-valid access
+			 * token usable while the administrator fixes stale client metadata,
+			 * a blocked client-metadata endpoint, or a provider-side policy.
+			 */
+			if ( self::is_client_configuration_error( $error ) ) {
```
kraftbj, line 1285: What happens here when the site's own `client_id` URL changes? A domain move, a switch to or from `index.php` permalinks, or a `rest_url` filter all change `Client::client_id()`, but refresh keeps sending the stored one. The auth server can't fetch that document and answers `invalid_client`. With this change we keep the session instead of asking for a reconnect, so publishing quietly stops once the access token expires. Site Health then tells the admin to check the client configuration, when reconnecting is the actual fix.

Could we compare the stored `client_id` against `self::client_id()` first and fall through to `mark_needs_reauth()` when they differ? Keeping the session still makes sense when they match.

includes/oauth/class-client.php
```diff
@@ -1134,6 +1204,26 @@ private static function refresh_locked( array $conn ): true|\WP_Error {
...
+				debug_log( \sprintf( 'refresh rejected client configuration (%s): %s', $error, $msg ) );
+
+				return new \WP_Error(
+					'atmosphere_client_configuration',
```
kraftbj, line 1309: I know backoff is on the follow-up list, but the impact is a bit bigger than "retries on every request." Nothing here moves `expires_at` or sets `needs_reauth`, so every `access_token()` call past the five-minute window makes a synchronous token request (15s timeout). Publish, backfill and reaction sync all go through that path. A concurrent caller that hits the lock can also sit in `wait_for_token_refresh()` for up to 45 seconds first. The five-minute dedupe only throttles the status write, not the request itself.

Before this PR, `invalid_client` ended that loop by marking the session for reauth. So I'd lean toward a minimal guard in this PR: a short transient set next to `record_refresh_failure()` that `refresh()` checks before taking the lock. The real backoff can still come later.

includes/class-atmosphere.php
```diff
@@ -2812,6 +2814,9 @@ private static function is_transient_publish_error( \WP_Error $error ): bool {
 		}
 
 		$permanent_codes = array(
+			'atmosphere_client_authentication',
+			'atmosphere_client_authentication_key',
+			'atmosphere_client_configuration',
```
kraftbj, line 2819: Small one: `atmosphere_dpop_keygen_failed` is still missing here (Copilot flagged it last round). `Client_Authentication::key()` returns it straight from `DPoP::generate_key()`, and it's as deterministic as the two codes above. Worth adding it to the test that covers those two as well.

includes/rest/class-client-metadata-controller.php
```diff
@@ -81,24 +110,30 @@ public function register_routes(): void {
...
-			'client_id'                  => Client::client_id(),
-			'client_name'                => sanitize_text( \get_bloginfo( 'name' ) ) . ' (ATmosphere)',
-			'client_uri'                 => \home_url( '/' ),
-			'redirect_uris'              => array( Client::redirect_uri() ),
-			'grant_types'                => array( 'authorization_code', 'refresh_token' ),
-			'response_types'             => array( 'code' ),
-			'token_endpoint_auth_method' => 'none',
+		$pinned = $this->pinned_fields();
+		if ( \is_wp_error( $pinned ) ) {
+			return new WP_REST_Response( array( 'code' => $pinned->get_error_code() ), 500 );
```
kraftbj: Every other controller in the plugin returns a `WP_Error` on failure, so the REST server builds the usual `code` / `message` / `data.status` body. This one sends only `{"code": ...}`. Returning `new \WP_Error( $pinned->get_error_code(), ..., array( 'status' => 500 ) )` would match, with a generic message since the route is public. The return type would need to become `WP_REST_Response|\WP_Error`.

includes/wp-admin/class-health-check.php
```diff
@@ -470,6 +512,64 @@ private static function connection_state(): string {
...
+			&& (int) $status['last_success'] < \time() - self::RENEWAL_STALE_AFTER;
+	}
+
+	/**
+	 * Whether the latest renewal failure is a client-configuration problem.
+	 *
+	 * @param array $status Refresh status as read from `Client::refresh_status()`.
+	 * @return bool
+	 */
+	private static function client_configuration_failed( array $status ): bool {
```
kraftbj, line 591: `atmosphere_client_authentication_key` never matches here, so a session stuck on an unreadable signing key gets the same generic "could not renew" recommendation as a 503 that will clear on its own. This failure doesn't fix itself while the session is bound to the key, and the fix is specific (disconnect, then connect). Could it get its own critical state with that advice?

kraftbj (commented):
Both blockers and the smaller items from my last pass are fixed. I checked each one against the code, not just the commit messages.

Inline comments below. The key-padding one is what I'd fix before merging: it came in with the stricter key validation, it breaks roughly 1 in 70 first connects, and it makes two of the new tests flaky.

docs/php-class-structure.md
```diff
@@ -273,9 +273,11 @@ via `register_routes()`; they are all instantiated together in
 
 Route namespaces are versioned deliberately:
```
kraftbj, line 276: The directory tree at the top of this file still only lists the old `oauth/` and `rest/` files. Could you add `class-client-authentication.php` and `class-legacy-client-metadata-controller.php` there too?

pfefferle:
Thanks @kraftbj, all eight are in (b6a8751).

Padding was the real find, I could reproduce your rate (22 of 2000 keys). `generate_key()` pads now, which is also what RFC 7518 asks for, and a test checks 200 generated keys. The bound-session check is scoped to confidential sessions, with the fixture fixed and a legacy-replacement test. A moved client_id marks reauth with its own reason and copy. The hold is a 5 minute transient checked before the lock; it returns its own code so the publish ladder retries instead of dropping the post. Keygen code added to the permanent list, the metadata route returns a `WP_Error`, the unreadable key has its own critical Site Health state, docs tree updated.

One thing I found while re-checking: the reference server answers a missing or mismatched metadata document with `invalid_client_metadata`, not `invalid_client`, so I added that code, otherwise neither the moved-domain branch nor the hold would have fired against bsky.social.

Known limit: on sites where a multilingual plugin varies `home_url` per request, the recomputed client_id differs between cron and web context, so a client rejection there ends in a reconnect prompt rather than the configuration state. I think that is acceptable, the rejection means refresh had already failed.

I have not done a live connect and publish against a PDS with the padded DPoP keys yet.

includes/oauth/class-client-authentication.php
```diff
@@ -0,0 +1,268 @@
...
+		 * An authorization in flight has already published this key's
+		 * `kid` through PAR; rotating it before the callback would fail
+		 * the exchange.
+		 */
+		$pending = \get_transient( 'atmosphere_oauth_resolved' );
+		if ( \is_array( $pending ) && ! empty( $pending['client_id'] ) ) {
+			return true;
+		}
+
+		$conn = get_connection();
```
kraftbj, line 218: This is the one I'd fix before merge. `authorize()` writes `atmosphere_oauth_resolved` (with `client_id`) at `class-client.php:385`, before `authorize_via_par()` signs, so this branch sees the connect attempt itself as a flow in progress and keeps the unreadable key. "Disconnect and connect again" loops, and every attempt resets the transient's hour.

I don't think the pending check protects anything. `stored_key()` only asks here once the key is already unusable, and an unusable key can't sign the callback's token exchange either, so a flow that went through PAR with it fails whether we replace the key or not. Dropping the block fixes the loop. (Resolving the key in `authorize()` ahead of the transients also works, but a leftover marker from an earlier attempt would still pin the key for up to an hour.)

`test_unreadable_key_is_kept_while_an_authorization_is_pending` would flip. A test that runs `authorize()` with PAR mocked against an unreadable key would pin the real path; the current one seeds the transient by hand, which is how this got past it.

```suggestion
	 * Whether a live confidential session is bound to the current signing key.
	 *
	 * Only a confidential session signs with this key; a legacy session
	 * never does, so it must not hold a broken key in place. A connection
	 * whose tokens were encrypted under a different key is already lost,
	 * so nothing is left for the signing key to protect either. A pending
	 * authorization is no reason to keep it: an unreadable key cannot sign
	 * the callback's token exchange, so that flow fails either way.
	 *
	 * @return bool
	 */
	private static function session_bound_to_key(): bool {
		$conn = get_connection();
```

kraftbj (commented):
Nice round. One thing I'd fix before merging, inline. Everything else is non-blocking.

Worth doing, not blocking:

- Legacy sessions after a domain or permalink move (Copilot's point, `class-client.php:1292`). The v1 `client_id` is recomputed from `rest_url()`, so a moved site sends a different ID, gets `invalid_client` or `invalid_client_metadata`, and lands in the configuration state plus the hold. Before this PR that was a reconnect. For a legacy session reconnecting is always the fix (it moves to v2), so I'd send every client-configuration error on a legacy session to reauth.

Non-blocking:

- The hold isn't bound to a session. The row check and `set_transient()` aren't atomic, so a reconnect landing between them inherits a 5-minute hold. Storing a hash of the `refresh_token` ciphertext in the hold and ignoring mismatches closes it. Low impact, since a fresh token rarely refreshes inside 5 minutes.
- `test_generate_key_pads_every_member_to_32_bytes` still passes about 10% of the time with the padding reverted (roughly 3 in 256 keys come back short, over 200 draws). A fixed-vector test with a 31-byte component would pin it.
- The new `is_safe_https_url( $client_id )` guard in `revoke_refresh_token()` has no test.
- `Client_Metadata_Controller::ROUTE_NAMESPACE` shipped as a public constant and is gone now. Nothing in-tree uses it, but an alias is one line and matches the other controllers.
- The PR description still says there's no backoff.

`composer lint` is clean and CI is green. I haven't run the suite locally.

includes/oauth/class-client-authentication.php
pfefferle: You were right, the pending check saw the connect attempt itself. Dropped it in 182d732, with your docblock. The hand-seeded test is replaced by one that runs `authorize()` with PAR mocked against an unreadable key; I checked that it fails with the old check in place.

includes/oauth/class-client.php
```diff
@@ -346,6 +390,7 @@ public static function authorize( string $handle, string $origin = 'settings' ):
 				'auth_server'  => $auth_meta,
 				'handle'       => $handle,
 				'origin'       => 'connectors' === $origin ? 'connectors' : 'settings',
+				'client_id'    => self::client_id(),
 			),
```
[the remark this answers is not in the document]
pfefferle: The pending-authorization check is gone (182d732), so the order of the marker and the first assertion no longer matters.

includes/rest/class-client-metadata-controller.php
```diff
@@ -25,15 +28,19 @@
...
+	 * ever served one must keep serving it unchanged while sessions minted
+	 * under it exist. The public-client document for legacy sessions is
+	 * served by {@see Legacy_Client_Metadata_Controller}.
 	 *
 	 * @var string
+	 *
+	 * @since unreleased
 	 */
-	public const ROUTE_NAMESPACE = 'atmosphere/v1';
+	public const VERSION = 'v2';
```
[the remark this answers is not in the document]
pfefferle: Added `ROUTE_NAMESPACE` back on both controllers, derived from `VERSION` (182d732).

includes/oauth/class-client-authentication.php
```diff
@@ -0,0 +1,268 @@
...
+	 * such session exists, the row is discarded so the caller mints a fresh
+	 * key and the site can connect again.
+	 *
+	 * @return array|\WP_Error|null The key, an error for an unreadable row, or null when none is stored.
+	 */
+	private static function stored_key(): array|\WP_Error|null {
+		$stored = \get_option( self::KEY_OPTION, '' );
+		if ( ! \is_string( $stored ) || '' === $stored ) {
+			return null;
+		}
```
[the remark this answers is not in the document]
pfefferle: Leaving this one. If the row is deleted under a live confidential session, the regenerated key makes the next refresh fail with `invalid_grant` ("session was initiated with a different key", that is the reference server's own message), which lands in the normal reauth path with a working reconnect. Routing it through the signing-key state instead would force a disconnect first, a worse path for something that only happens through a manual database edit.

pfefferle:
Third round is in (182d732). The blocker was real, thanks for the trace: the pending check saw the connect attempt itself and looped. It is gone, and a test now runs `authorize()` against an unreadable key.

Also done: a legacy session reconnects on any client-configuration error, the hold is bound to the session whose refresh failed, the padding test uses a fixed 31-byte vector, the revocation client_id guard has a test, `ROUTE_NAMESPACE` is back, and the description no longer claims there is no backoff.

One thing I noticed in the reference server while re-reading the refresh path: it lets a session that started as a public client upgrade to confidential on refresh under the same client_id. If that holds against bsky.social, a follow-up could make the v1 document confidential too and the remaining legacy sessions would upgrade on their next refresh without anyone reconnecting. I have not tested it live, so nothing in this PR depends on it.

kraftbj (dismissed):

jeherve (commented):
Hey @pfefferle! I went through the branch again at 182d732. The new commits fix the key padding and the unreadable-key recovery, and revocation now goes out as the client the session belongs to.

There are a few things left that were surfaced by Claude, and I think we should address the first one before we merge.

**A rejected client configuration only shows up in Site Health.** For a confidential session, `invalid_client` / `unauthorized_client` now keep the session and put the refresh on hold. That makes sense to me, since reconnecting the same client wouldn't fix anything. That said, `maybe_render_reauth_notice()` never fires in that case, and `atmosphere_client_configuration` is a permanent publish error. So posts stop going to Bluesky, and the only signs in the admin are a Site Health entry and a debug log line. I'm wondering if we could show an admin notice for that state too, pointing people to Site Health.

Smaller things, feel free to take or leave:

- An anonymous `GET /wp-json/atmosphere/v2/client-metadata` creates the signing key and writes the option on a site that has never connected. It only happens once, so I don't think it's a big deal; I'd just rather the first write came from an admin connecting.
- The "more than 24 hours" in the stale-renewal Site Health message is hardcoded right next to `RENEWAL_STALE_AFTER`, so the two could drift apart if the constant ever changes.
- The `atmosphere_client_metadata` docblock still describes `client_id` as "advertised as the OAuth client identifier", but it's now reset after the filter runs, and filters can't change `token_endpoint_auth_method` anymore either. Mentioning that in the hook doc would help anyone who already has a filter in place.

Let me know what you think!

pfefferle:
Thanks @jeherve. The notice is in (153b10c): both states that keep the session but block renewal, a rejected client registration and an unreadable signing key, now get a site-wide notice that points to Site Health, with the same audience rules as the reconnect notice. The Site Health test and the notice read the same predicate so they cannot drift. The 24 hours now come from the constant, and the filter docblock says which fields are reset after it runs.

The anonymous GET I left as is. The auth server fetches the document during connect, after `authorize()` has already created the key, so on a connecting site the first write does come from an admin. On a never-connected site, refusing to create it would mean serving an invalid document or a 500, and the Site Health reachability test probes exactly that URL. The public half is not a secret and creation is first-writer-wins.

kraftbj (approved):

jeherve (approved):
This is looking good, should be good to merge now! 🚢

jeffpaul:
@pfefferle THANK YOU FOR THIS ONE!

